PR #87, opened 2017-10-24
Changed speed of game clock and adjust max character slots

### LINKED ISSUE #69: Day Night Cycle is too fast
The current day/cycle is too fast, transitioning through a full Cycle every 10min or so. It's so quick you can see the moon moving across the sky in jerky phases.

According to ParagonWiki: 
One real time hour is equivalent to two in-game days, with midnights occurring on the hour and on the half hour.

They have a handy chart as well: https://paragonwiki.com/wiki/Time

And includes information regarding timed events, which means eventually we'll want to expose the game "clock" to the Lua scripting engine for missions and spawns.

broxen:
Closes #69. Though it may be reopened if there's a case for syncing the game clock precisely to the half-hour server-time. The only scenario I can think of is perhaps spawn events that should be triggered at midnight or noon, etc.

4.8 instead of 48 because I couldn't see where the math multiplied the delta by 10. Maybe I missed something?

- Modified World::update so that every 1 sec of real time equals 48 sec of game time. 48:1
- Experimented with branches on my fork, so sorry for filling the commit log up with junk
- Not sure if I need to merge upstream before pulling, but I did anyway.


broxen:
Added the adjustment to character slots per account to this after discovering the unused macro in Character.h

PR #101, opened 2017-10-27
Implemented getAccountMaxSlots() and moved max_account_slots from

### LINKED ISSUE #79: Increase number of character slots per account and allow configuration in settings.cfg
Currently the server allows for only two character slots per account.

I couldn't find where this is regulated in the source, but we should expand this to 12.

ldilley:
Closes #79.

- Implemented getMaxAccountSlots()
- Moved max_account_slots from AdminServer/CharacterDatabase to GameServer
- Made settings.cfg a bit prettier
- Moved MAX_CHARACTER_SLOTS constant from Character.h to GameServer.h


Projects/CoX/Servers/GameServer/GameServer.h
```diff
@@ -27,6 +27,8 @@
 
 #include <string>
 
+#define MAX_ACCOUNT_SLOTS 8
```
nemerle: No need to pollute the header files with defines, just move it into the GameServer.cpp

Issue #66, opened 2017-10-20
Local chat outputs to Broadcast channel instead, meanwhile Broadcast chat appears to do nothing


broxen:
Using Local chat (either through UI or with slash commands) outputs message to the Broadcast channel, meanwhile the broadcast channel appears to do nothing.

Test:
`/l test local!` output: `[Broadcast] test local!`
`/b test broadcast!` output: _none_

I haven't been able to test Team, Supergroup, or Private Messages yet.

nemerle:
yup this is a known issue, since we didn't have a mapserver scene graph while it was implemented so there was no way to calculate the distance one player could "hear" the local chat's from another.

nemerle:
A few steps towards proper implementation

1.  update ChatMessage::localMessage to use CHAT_Local, see what it does.
1. make MapServer instance accessible in ChatMessage, maybe from Entity ? ( i.e. implement Entity::getCurrentMap() )
1. implement neighborhood query in MapInstance ( MapInstance::getEntitiesInRadius(float R, int type) )
1. send the local chat to entities return in previous step.

broxen:
This issue probably relates to #41

broxen:
CHAT_local works. I'll have to test if there is any range to it.

I tried adding a function for broadcast, but couldn't figure out where to add the slash command `/b` or `/broadcast`

nemerle:
MapInstance::on_console_command handles all console input
broadcast messages are encode as ```b Message text```

broxen:
Latest commits break broadcast chat. Trying to fix with pull #117

nemerle:
As of a849fc2 chat looks like:
![chat test](https://user-images.githubusercontent.com/96597/32393712-ebf8c4ae-c0da-11e7-8157-21c4b11f0a7b.png)

broxen:
This looks good, but we need an additional space between the channel and sender_name.

nemerle:
Should be easy to do:
replace 
```cpp
QString prepared_chat_message = QString("[Local]%1: %2").arg(sender_char_name,msg_content.toString());
```
with
```cpp
QString prepared_chat_message = QString("[Local] %1: %2").arg(sender_char_name,msg_content.toString());
```

nemerle:
This seems to be resolved, closing.

Issue #94, opened 2017-10-25
[Movement] Rubberbanding movement and Interpolation


broxen:
**This issue began with work on orientation, which was not being updated correctly. The scope has since expanded to include all movement**

![screenshot 48](https://user-images.githubusercontent.com/10159505/31977986-1e9b1716-b8f4-11e7-8b23-a37e0e8bafb9.png)

### This will need
- [ ] Correct Position Update math (math in `physicsStep(...)`)
- [ ] (interp-v2, sending junk vals) Correctly send interpolated movement values in EntityUpdateCodec (`storeUnknownBinTree(...)`)
- [ ] Magic all of these values together
- [x] Correctly store ControlBits states (keypresses and releases) with keypresses time.
- [x] Do we need MotionStates (surface and physics params) for this to work? Probably.
- [x] Do we need TimeStates (record of when keys were pressed relative to server time) for this to work? Probably.
- [x] (interp-v2) Store InputStates to vector InputStateStorage
- [x] (interp-v2, I think 😜) Correct Velocity calculations -- Currently the velocity the client believes you to have, does not match with the server's expectation
- [x] Verify stops to server/client -- it appears as though the client doesn't *always* recognize that you've stopped moving (may relate to ControlBits)
- [x] Store Position updates to PosUpdates for interpolation
- [x] Calculate Interpolation values from PosUpdates (return BinTree)
- [x] Does EntityUpdateCodec send m_update_pos_and_cam every time? Or only for special updates?
- [X] Correct Entity Rotation upon spawn (Fixed by #104)
- [x] Correct Entity Rotation during movement in EntityUpdateCodec (addressed by #166)
- [x] Verify Speed calculations
- [x] Verify Drag calculations
- [x] Verify Gravity calculations
- [x] Client does not recognize ABS_TIME, though it is sent by the server

### Bonus
- Send Animations
- Send NetFX
- Correctly calculate jumping (this is going to be a whole other set of algos) 

### Useful testing commands (require Access Level 9)
- `/controldebug 1` - outputs to client console and shows markers in the world with position deltas
- `/entdebugclient 1` - displays GUI window that shows active data, and displays reticle on entity
- `/moveto {x} {y} {z}` - moves entity to position
- `/postest` - Outputs server values for position. Only available on my broxen/segs/interp-v2 branch
- `/interp` - Runs an interpolation test. Only available on my broxen/segs/interp-v2 branch

### SEGS version
[...]

PR #103, opened 2017-10-27
Added broadcast and modified local chat functions. Corrected orientation of other entities on spawn

broxen:
Addresses #66 but does not resolve either.

- Add broadcastMessage
- Reset localMessage to local channel
- TODO: restrict local channel to range

### DIFF
--- a/Projects/CoX/Servers/MapServer/EntityStorage.cpp
+++ b/Projects/CoX/Servers/MapServer/EntityStorage.cpp
@@ -87,7 +87,7 @@ void EntityManager::InsertPlayer(Entity *ent)
     m_map_entities[m_last_ent++] = ent;
     ent->m_idx = m_last_ent-1;
     ent->pos = glm::vec3(128.0,16,-198); //-60.5;
-    ent->qrot= glm::angleAxis(glm::radians(-15.0f), glm::vec3(0.0f, 1.0f, 0.0f));
+    ent->qrot= glm::quat(0.0f,0.0f,0.0f,1.0f);
     m_entlist.push_back(ent);
 }
 Entity * EntityManager::CreatePlayer()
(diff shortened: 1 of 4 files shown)


Projects/CoX/Servers/MapServer/EntityStorage.cpp
```diff
@@ -87,7 +87,7 @@ void EntityManager::InsertPlayer(Entity *ent)
     m_map_entities[m_last_ent++] = ent;
     ent->m_idx = m_last_ent-1;
     ent->pos = glm::vec3(128.0,16,-198); //-60.5;
-    ent->qrot= glm::quat(0.0f,0.0f,0.0f,1.0f);
+    ent->qrot= glm::angleAxis(glm::radians(-15.0f), glm::vec3(0.0f, 1.0f, 0.0f));
```
nemerle: Let's fix this by translating qrot when it is sent, to euler angles (pitch yaw roll ) and if that works ok, leave this code as it was before ?

broxen: It seems like we still need to modify this default value, as entities are currently spawning sideways before they even move? Maybe I'm misunderstanding? I'll continue to look at modifying the AngleQuantized formula as in #94, as that seems to affect this issue as well.

I agree with you that I'm confused by the necessary -15deg adjustment on angle, and further confused that the code that exists here isn't working as intended based upon the math I'm seeing.

I'm still figuring out git, but I'll try to resubmit this pull, without this change, to incorporate the ChatMessage changes, unless there's a way to reject this change on your end?

nemerle: Problem exists at the point we're sending qrot, here the value is still ok.
As for pull request, if You've created a branch for it ( as it seems You did :smile: ) just create a new commit with that specific change, and push it to your branch, and it should automatically update the pull request

Issue #65, opened 2017-10-20
Compiling\Installation Guide Needs Updating (and install script?)


broxen:
The installation guide on the forums and github wiki need updating and/or additional detail.

Specific objectives should be:
- How to Compile in major platforms and the tools or libraries required. See #36
- What the server needs in order to operate properly (data files, piggs, etc), and where to get those files
- How to configure local_1.cfg (might need a name change? proposed: settings.cfg
- How to setup and configure the postgres database
- Include a SQL script to setup the database automagically
- How to acquire the client and what version of the client work
- How to configure the client to access your server
- Common Issues/FAQ

These guides might be best broken up into three separate how-tos: howto-compile howto-setup-server and howto-setup-client.

I'm happy to tackle this task, but as of now, I'm not 100% sure I'm doing it correctly myself. Once I've gotten everything running on both Linux and Windows, I'll feel more confident taking this on.

In the future, an install script that walks a user through the setup/config would be excellent.

nemerle:
Note: SEGS supports sqlite now, so that might be a better choice for default setup guide target.

broxen:
Just leaving myself a reminder: bin.pigg must be extracted to the server data folder.

`./piggtool -x PATH_TO_COH_INSTALL/piggs/bin.pigg`

broxen:
This has overlap with #36

broxen:
Initial DRAFT attached. As always, feedback is appreciated.

Not sure if you actually want to offer binary downloads, but this file supposes that they exist.

Remaining tasks below.
- [ ] Creation of https://segs.nemerle.eu/download with links to latest stable release binary
- [ ] Building from source should provide a /db/ folder with empty segs and segs_game databases. Could this be done in a way that eliminates a lot of the database setup?
- [ ] Rename local_1.cfg to settings.cfg
- [ ] How to create Admin Account for server?

```Markdown
INSTALLATION GUIDE & FAQ
======

CONTENTS OF THIS FILE
------
- Requirements and notes
- Installation
- Setup Database
- Configuration
- More information
- FAQs

REQUIREMENTS AND NOTES
------

SEGS requires:

- A copy of the CoH client from Issue 0 (release) version 0.22. This can be found through 
  various sources online, or via magnet link: 
  - magnet:?xt=urn:btih:c2e7e0aa72004594337d73ae6a305242e23d815f
- One of the following databases:
  - PostgreSQL 8.3 (or greater) (http://www.postgresql.org/).
[...]

PR #105, opened 2017-10-28
Updated README and added default files to output

broxen:
Addresses #65 and adds several "starter" files to the executable output folder at compile, including settings.cfg, and the sqlite starter databases.

- Add new README.md to /out
- Add default settings.cfg to /out
- Add default segs db to /out
- Add default segs_game db to /out
- TODO: README needs proofing, and some additional information about setting up an Admin User account during configuration.


Projects/CoX/CMakeLists.txt
```diff
@@ -6,3 +6,8 @@ ADD_SUBDIRECTORY(Utilities)
 ADD_SUBDIRECTORY(Servers)
 ADD_SUBDIRECTORY(Clients)
 
+# Add README, settings.cfg, and Starter Databases
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/docs/README.md ${EXECUTABLE_OUTPUT_PATH}/README.md COPYONLY)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Data/settings.cfg ${EXECUTABLE_OUTPUT_PATH}/settings.cfg COPYONLY)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Data/sqlite/segs ${EXECUTABLE_OUTPUT_PATH}/segs COPYONLY)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Data/sqlite/segs ${EXECUTABLE_OUTPUT_PATH}/segs_game COPYONLY)
```
nemerle: Typo here, should be configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Data/sqlite/segs_game ${EXECUTABLE_OUTPUT_PATH}/segs_game COPYONLY)

broxen: Yes, and also, I realized that this will overwrite any existing db or settings.

Perhaps it's better to move into a subfolder like `/data/` or perhaps `/defaults/` instead?

nemerle:
default_databases ?
After that we can add an Issue to make the servers copy databases from default_database if they don't find their own

Projects/CoX/CMakeLists.txt
nemerle: default_databases ?

Projects/CoX/CMakeLists.txt
```diff
@@ -6,3 +6,31 @@ ADD_SUBDIRECTORY(Utilities)
...
 ADD_SUBDIRECTORY(Clients)
 
+# Add README, settings.cfg, and Starter Databases
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/docs/README.md ${EXECUTABLE_OUTPUT_PATH}/default_dbs/README.md COPYONLY)
+if(EXISTS ${EXECUTABLE_OUTPUT_PATH}/README.md)
+  # nothing else needs to be done
+else
+  configure_file(${CMAKE_CURRENT_SOURCE_DIR}/docs/README.md ${EXECUTABLE_OUTPUT_PATH}/README.md COPYONLY)
+endif
+
```
nemerle: Maybe consider using ```if(NOT EXISTS ${EXECUTABLE_OUTPUT_PATH}/README.md)```

Unless we want to inform the user about the state of affairs ?

broxen: Ahh, I wondered if there was such a thing. Much cleaner!

nemerle:
Ok, the cmake code looks clean. and good enough for now. Later on this functionality should be the part of the server setup procedure:
1. server starts up, finds that it's missing required config files/databases,
1. asks user to run ServerMegaSetupUtility
1. ServerMegaSetupUtility creates cfg files, depending on selected database format also creates databases.
1. ...
1. profit !

PR #117, opened 2017-10-30
Remove duplicate README.md and fix typo for Broadcast Chat

broxen:
Removes the duplicate copy of the new README from the default_dbs folder


broxen:
Actually broadcast still isn't working for me. Let me explore further.

PR #126, opened 2017-11-02
Fixed issue with broadcast channel not working

broxen:
Removed superfluous space from check for chat commands. Fixes error with `/b` not working

Addresses #66

### DIFF
--- a/Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
+++ b/Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
@@ -143,7 +143,7 @@ bool CharacterDatabase::fill( Character *c)
     CharacterCostume *main_costume = new CharacterCostume;
     // appearance related.
     main_costume->m_body_type = m_prepared_char_select.value("bodytype").toUInt(); // 0
-    c->setMapName("Some map name"); // "V_City_00_01.txt" STR_OR_EMPTY(r.getColString("current_map"))
+    c->setMapName(STR_OR_EMPTY(m_prepared_char_select.value("current_map").toString()));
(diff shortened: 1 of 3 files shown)


nemerle:
Does this mean that incoming chat messages are formatted as "/bMESSAGE_TEXT" ?

nemerle:
You could always set a break-point at ChatMessage::eChatTypes getKindOfChatMessag and see what's coming from the client

broxen:
Shouldn't have any impact on the output, as I'm only modifying the string used for lookup in the check for chat type?

I'll confirm tonight, and will add some debug tests.

nemerle:
Well, the change You're proposing would make getKindOfChatMessage return true for every command starting with letter "l"

broxen:
Ah, yes. I completely missed that.

broxen:
I've resolved any errors with Malformed chat commands, added range to local chat (100.0f for now?), fixed the broadcast chat issues, and made local_chat the fallback for any unknown scripts or commands (as in original servers). I did leave a bit of commented out debugging code for testing purposes. Please test and let me know what you think, but overall I think I tested for everything.

broxen:
Oh, I also added some code for pulling current map from the database on the char-select screen. This is probably what I'll try to tackle next.

nemerle:
Ok, writing a simple "hello" on the local chat window results in the client sending
"l hello\n"
and using "/local hi"
"local hi"
Both methods seem to work correctly as-is, could you give me an example that fails ?

nemerle:
But You were right, getKindOfChatMessage should only ever chack first character since it's extracted from the start of message until first ' ' character, sorry my bad

PR #150, opened 2017-11-09
Add FloatingDamage and InfoMessageCmd commands

nemerle:
Also allow accessing owner client of the Entity ( if it has one. )


mobbyg:
You can see it in action in this video... 
[https://youtu.be/zQ0Tduks85Q](url)

Issue #133, opened 2017-11-06
Versioning database tables


nemerle:
If we want to support database upgrades ( and we really should ), we need to store an additional table in the database, that will contain a version number of each table, so we can lookup proper 'upgrade' scripts to run. Something like this:

| ID | Table_Name | Version number |
| ------------- | ------------- | ------------- |
| 1  | "accounts"  | 1 |

Then in database management tool we can query current database for the table list and lookup their versions, and run apropriate update scripts.

broxen:
Perhaps we can utilize SQLite's `PRAGMA user_version` to store our own schema version information. There is also `PRAGMA schema_version` but it seems that SQLite auto increments that for every `CREATE` or `DROP` command, but it's not recommended for manual use as it may be changed by SQLite itself.

http://www.sqlite.org/pragma.html#pragma_user_version

Then, as you said, we could check this version info upon authserver startup to see if it matches the version in our default files under the default_dbs folder. If not, `qFault()` and recommended running dbtool to update the databases.

We can either store the update scripts in a default_dbs subfolder and call them in order until we reach version parity, or hard code those queries within the dbtool. This is similar to how Drupal runs dbase updates in PHP/MySQL.

Or worst case, the dbtool can be used to overwrite the current database completely with a new (updated) one.

I'll start to work on this issue. If you have any specific template or roadmap for the database schema in mind please let me know.

broxen:
So definitely no `schema_version`

> SQLite automatically increments the schema-version whenever the schema changes. As each SQL statement runs, the schema version is checked to ensure that the schema has not changed since the SQL statement was prepared. Subverting this mechanism by using "PRAGMA schema_version" may cause SQL statement to run using an obsolete schema, which can lead to incorrect answers and/or database corruption.

http://www.sqlite.org/pragma.html#pragma_schema_version

nemerle:
My thoughts:
* Add a table_versions table to database
* initialize it with all table names that are already in our database
* set all versions to 0.

Using sqlite-only pragma's is not the way to go, since it will prevent using any other DB engine ( and since we use Qt, we get a few of those 'for free' )

nemerle:
As for storing updates, I'd prefer to store them in tool, since some updates could need to perform some logical operations, for example if we had a table with firstname_lastname column, and the update wanted to split that column into two firstname and lastname, then writing a standalone sql script that would do such a thing could become complicated. We could even go one step further towards the standard way of doing such things, and call them `migrations` :P

[...]

PR #154, opened 2017-11-11
Updated database schemas and cleaned up data folder

broxen:
Updated database schemas based upon #133, added database schema exports to the Data folder

Addresses #133

- Updated `segs` and `segs_game` database schemas per #133
- Added schema exports to the Data folder as backup
- Proposed: delete other SQL scripts in the `./Projects/CoX/Data/` folder, as they are no longer relevant.


nemerle:
The old sql scripts were targeting pgsql, but I agree that only storing sqlite is good enough for now.
On day, when we add a capability of creating/initializing databases from the db tool, then we can start worrying about the specific DB server specific scripts.

broxen:
Ah I see. I can try to import the schema into dbdesigner.net as pgsql and export scripts for that. My only recommendation then, would be to add a subfolder to data for pgsql, so it's clear to users what is what.

nemerle:
That would be great, thank you !

broxen:
Leaving myself a note to test this tool out: https://github.com/dimitri/pgloader

broxen:
I've been on Chromebook all weekend, so I haven't been able to test this, but assuming it imports correctly, this should be ready to go.

I can test late tonight and give the OK, but would love any feedback. Thank you in advance!

broxen:
After some small tweaks, I'm showing successful tests with both sqlite and pgsql.

Unless you find something different, this is OK to commit.

Issue #159, opened 2017-11-14
Warning : Failed to locate origin index for [Origin Type]


broxen:
### Issue Summary
Upon creating a character the server window outputs the following error message: `Warning : Failed to locate origin index for "Technology"` where `Technology` is whatever Origin you've chosen.

### System Info
Latest from Git 02ec61b
Qt 5.9.2 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 7.2.1 20170915 (Red Hat 7.2.1-2)) on "xcb" 
OS: Fedora 26 (Workstation Edition) [linux version 4.13.8-200.fc26.x86_64]

nemerle:
Question is, is this a failure when searching for the "Technology" string ?, or is it a problem with loading origin/class data files ?
Does the server report a correct load of:
* origins.bin
* villain_origins.bin
* classes.bin
* villain_classes.bin

broxen:
Yes, it appears to be loading the bins correctly. 

Also, it only shows this error when creating a _new_ character, not when loading an existing character from the dbase.

My console at runtime below.
```
Super Entity Game Server
http://segs.sf.net/
Copyright (c) 2006-2017 Super Entity Game Server Team (see Authors.txt)
This software is licensed! (See License.txt for details)

segs v0.4.0 alpha (The Real Beginning)
main
Debug   : Config file is missing 'location_addr' entry, will try to use default
Debug   : Config file is missing 'auth_pass' entry, will try to use default
Warning : Reading map data from ./data/bin/ folder
Debug   : Reading ./data/bin/costume.bin ... OK
Debug   : Reading ./data/bin/supergroupColors.bin ... OK
Debug   : Loading origins:
Debug   : Reading ./data/bin/origins.bin ... OK
Debug   : Reading ./data/bin/villain_origins.bin ... OK
Debug   : Loading classes:
Debug   : Reading ./data/bin/classes.bin ... OK
Debug   : Reading ./data/bin/villain_classes.bin ... OK
Warning :  All map data read
Warning :  Postprocessing runtime data ..  Hashes filled
```

broxen:
Fixed by latest commit #161

Issue #160, opened 2017-11-14
Zoom-in camera to First Person View results in `Not all bits were consumed`


broxen:
### Summary
Zooming in camera to First Person View results in `Not all bits were consumed` debug message.

### SEGS version
segs v0.4.0 alpha (The Real Beginning)

### System Info
Qt 5.9.2 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 7.2.1 20170915 (Red Hat 7.2.1-2)) on "xcb" 
OS: Fedora 26 (Workstation Edition) [linux version 4.13.8-200.fc26.x86_64]

nemerle:
Should be resolved by a8b0b0d

broxen:
Handled by #161. Can probably close this until we know what to do with `EVENT_DECL(evSwitchViewPoint        ,264)`

PR #161, opened 2017-11-14
Fixes the class name retrieval bug

nemerle:
+ fix ContactBonusLength variable type.

Should help with #159, @broxen, it seems it works here, but just to be sure could You test this please ?


broxen:
This has resolved #159, which can be closed.

#160 now shows `Warning : Unhandled switch viewpoint to true` which is as currently designed.

Thanks Nemerle! Great work!

Issue #155, opened 2017-11-11
Add markup testing code


nemerle:
Write a chat command that will cause the server to read a simple markup language X file (smlx for short :wink: ) and send it as a standard dialog command to the client. It's main role is to help with writing markup language files for various game messages.
Steps to implement:
1. Modify ```void MapInstance::on_console_command(ConsoleCommand * ev)``` to handle 'smilex ' console command.
2. While processing 'smilex' request, check if server has a 'test.smlx' file on disk
3. If it's there, send it's contents to the client wrapped in StandardDialogCmd instance.
4. If it's not there, send appropriately mournful message to the client.

broxen:
Hi @nemerle!

I'm looking this over and I have several questions. Forgive me in advance, as I'm sure I'm just overlooking something obvious, but here goes...
- Aren't we somewhat reinventing the wheel with this additional scripting method? We already have XML parsing through QT, and LUA with SOL2, either of which could handle this task.
- Is there some benefit to parsing the text outside of the script file itself? Maybe localization? Speed? I'm probably just overlooking an obvious benefit.
- How is reusability affected? Can I reference the same string multiple times?
- In `misc.pigg` the text strings already contain psuedo HTML, how would we rewrite this? For example in `Contact2_Task_taskset.ms`
```
Things are getting out of control, and I can tell you need to learn all about combat, {HeroName}. First, I want you to <A HREF={ACCEPT}>learn about how tough things can be</A> by talking to Professor Hoffman in front of the deactivated robots to my left, then talk to Lt. MacReady for your combat lesson.
```

Just trying to save you work on another scripting engine, if we can leverage existing tools to complete this same task.

broxen:
This example from the same file is probably better
```
[...]

PR #164, opened 2017-11-18
Added Smilex slash command which loads smlx folder from ./scripts/

broxen:
First run at #155

- add handler for /smilex {filename} command
- load contents of {filename} into dlg
- error msg if none found

ToDo:
- Parse smlx file?? Not sure the end goal with /smilex, is it just for testing only?

### DIFF
--- a/Projects/CoX/Servers/MapServer/MapInstance.cpp
+++ b/Projects/CoX/Servers/MapServer/MapInstance.cpp
@@ -573,17 +573,17 @@ void MapInstance::on_console_command(ConsoleCommand * ev)
         src->addCommandToSendNextUpdate(std::unique_ptr<InfoMessageCmd>(info));
     }
     else if(ev->contents.startsWith("smilex ")) {
-        int first_space = ev->contents.indexOf(' ');
-        int second_space = ev->contents.indexOf(' ',first_space+1);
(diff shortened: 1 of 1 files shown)


broxen:
Currently loading the entire file and displaying within the dlg window.

screenshot from 2017-11-18 00-01-56

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -572,6 +572,22 @@ void MapInstance::on_console_command(ConsoleCommand * ev)
         }
         src->addCommandToSendNextUpdate(std::unique_ptr<InfoMessageCmd>(info));
     }
+    else if(ev->contents.startsWith("smilex ")) {
+        int space = ev->contents.indexOf(' ');
+        QString fileName("scripts/" + ev->contents.mid(space+1));
+        QFile file(fileName);
```
nemerle: Consider making file extension optional, that way we can just perform
```
/smilex testing_dlg
```
instead of /smilex testing_dlg.
just checking ev->contents.trimmed().endsWith(".smlx") should be enough

broxen:
Makes perfect sense. Changed now.

Issue #131, opened 2017-11-06
Local Client Configuration Options are not loaded automatically


broxen:
### Summary

Client configuration options are not loaded automatically. This can be replicated by logging into any account/character and attempting to use the `Q` and `E` keys to rotate your character left and right. These keybinds will not work initially. However if you go into Options and (without changing anything) save the options, the keybinds will work as configured.

Perhaps the client checks with the server for some kind of information before loading these local config settings? Saving locally simply forces the client to recognize the local settings regardless of what the server says? I'm just speculating.

### SEGS version
segs v0.4.0 alpha (The Real Beginning) - the latest from git

### Operating system
Fedora 26 or Windows 10

### C++ compiler and version
g++ 5.3.0

nemerle:
Yup client options are sent from the server when the client logs in, as we don't have them stored anywhere, we send a minimal set of those, and that means many things are not defaulted, but zeroed out :)

broxen:
After #154 is committed, I'll need an output of the default options and window settings from the client. I'll then add those to the database schema as defaults, and we can proceed to loading those defaults for new characters from the database.

Finally, we'll need to write any changes to options/keybinds/windows to the database. 

Or maybe we should hardcode the defaults and only save options that override that per character? This is how ParagonChat handles keybinds, for instance.

Maybe I'll look at a config save from a later (I23+) client to see how the client stores that information, as I assume it's sent to the server in a similar format.

broxen:
Saving here for reference. Default client configs from an I23+ client.

[keybinds.txt](https://github.com/Segs/Segs/files/1469320/keybinds.txt)
[options.txt](https://github.com/Segs/Segs/files/1469321/options.txt)
[wdw.txt](https://github.com/Segs/Segs/files/1469322/wdw.txt)

broxen:
Nemerle, when you're back from your travels, can you help me with this? I'm sure these are just newbie questions, but I'm not finding the answers I'm looking for online.

I'm trying to set defaults in sendKeybinds method of Character. But it doesn't seem like it's having the effect I expected? In the below code, the `Q` and `E` keys don't work =(

```
    /* Default Config
     *  ' "quickchat"
     *  - "prev_tray"
     ...
     *   UP "+forward"
     *   UPARROW "+forward"
     *   V "+ctm_invert"
     *   W "+forward"
     *   X "+down"
     *   Z "powexec_abort"
     */
    for(int i=0; i<256; i++)
    {
        switch(i)
        {
[...]

PR #165, opened 2017-11-19
Added default keybinds

broxen:
Step 1 of #131  .

- Added default keybinds to Character.cpp
- Modified degrees_for_turns in sendOptionsFull()
- Modified the map instance data load message to make it more clear
- Tested: it works! For `q` and `e` at least. I haven't tested every button.


nemerle:
The code looks good, although the mod-keys are not the ones you've guessed, they look like this:
```cpp
enum ModKeys
{
    NO_MOD = 0,
    CTRL_MOD = 0x1,
    SHIFT_MOD = 0x2,
    ALT_MOD = 0x3,
};
```

broxen:
Oh. Cool. Thanks Nem!

I'm not familiar with best practices, but should I move the enums to the header file?

broxen:
Also, I wasn't sure if bind[114] was correct. Is there any other definition for mouse chords?

nemerle:
1. Enums should be in a header file, if any other piece of code want's to use them, otherwise putting them in an anonymous namespace inside the cpp file should be preferred. 
2. I do not believe that I0 client handles those mouse chords as a part of keybindings.

Also default keybinds can be read from bin/kb.bin, so we could store keybinds in a MapServerData and read them on startup ?
I'm going to add serializer for the kb.bin just in case :)

broxen:
> Also default keybinds can be read from bin/kb.bin, so we could store keybinds in a MapServerData and read them on startup ?

This is preferable. From there we could override any user set keybinds from the dbase. This would reduce the clutter in the dbase and here in Character.cpp.

In light of that, I'll push these changes and open a new issue to follow up?

nemerle:
Agreed, I should be able to push the serialization code today, depends on how the cold, I've somehow acquired, allows :sneezing_face:

broxen:
Feel better!

Conferences will do that to you. Too many new germs around.

broxen:
This is ready for commit. Thanks Nem. Feel better brotha.

PR #166, opened 2017-11-27
Corrected update of orientation for other entities

broxen:
Addresses the issue #94 but still needs to resolve the issue of the `camera_pyr` affecting `inp_state.direction`, which really should be pulled from the client separately and should reflect the model's orientation.

- Modifies `SetOrientation()` to pull from `inp_state.direction`
- Modifies `AngleQuantize()` which was eliminating half of the rotation
- TODO: figure out how to update without relying on `camera_pyr`, which rotates the model when using the camera to look.
- TODO: remove remnant debug code left in DEBUG_INPUT ifdefs.

This can be tested by logging into a character and typing the command `/turnleft 1` into the chat window. This will set your character to rotate left indefinitely. Then log in with another character and confirm correct rotation.


broxen:
As far as I can tell, the client should be sending an update of the model's orientation along with the position and camera_pyr. If we can capture that value directly, we can update direction correctly, and avoid some of the issue in this pull request. Thoughts @nemerle?

Projects/CoX/Servers/MapServer/EntityUpdateCodec.cpp
```diff
@@ -138,8 +141,8 @@ uint32_t AngleQuantize(float value,int numb_bits)
 {
     int max_val = 1<<numb_bits;
 
-    float v = std::abs(value)>1.0f ? std::copysign(1.0f,value) : value ;
-    v  = (std::asin(v)+F_PI)/(2*F_PI); // maps -1..1 to 0..1
+    float v = value;
+    v = (v+F_PI)/(2*F_PI);
```
nemerle: I have no idea why my code is trying to use arcsin there :frowning_face:

Projects/CoX/Servers/MapServer/EntityUpdateCodec.cpp
```diff
@@ -138,8 +141,8 @@ uint32_t AngleQuantize(float value,int numb_bits)
 {
     int max_val = 1<<numb_bits;
```
nemerle: we need to either check that incoming angle is in a sane range, or normalize any value to match the range:
```cpp
float normalizeRadAngle(float ang)
{
    float res = ang;
    if ( ang > F_PI )
        res -= 2*F_PI;
    if ( res <= -F_PI )
        res += 2*F_PI;
    return res;
}
```

Projects/CoX/Servers/MapServer/EntityUpdateCodec.cpp
```diff
@@ -151,17 +154,42 @@ void storeOrientation(const Entity &src,BitStream &bs)
...
+#ifdef DEBUG_INPUT
+            fprintf(stderr,"v: %d\n", v);
+#endif
+            bs.StoreBits(9,v);   // normalized quat, 4th param is recoverable from the first 3
         }
+#ifdef DEBUG_INPUT
+        else
+            fprintf(stderr,"what happened? update_rot failed.\n");
+#endif
     }
```
nemerle: ```update_rot``` is meant as a query function, basically it should check what the server believes the client thinks about the orientation, and if the current value is different then the client's belief, it should return true, otherwise is should rturn false. This way CoH is trying to reduce the amount of bits transmitted.
So when update_rot returns false, it means that no update is needed :)

broxen: Ah, that makes sense.

Projects/CoX/Servers/MapServer/EntityUpdateCodec.cpp
broxen: Added this to CoHMath.h as that seems to be the appropriate place?

broxen: Also removed `toEulerAngle()` as it's replicating the function `toCoH_YPR()`

nemerle: Yes, CoHMath seems to be the place for it.

broxen:
Meant to rebase at 50e9b45 instead of merge. Hope that doesn't mess anything up.

Projects/CoX/Servers/MapServer/EntityUpdateCodec.cpp
```diff
@@ -138,29 +126,48 @@ uint32_t AngleQuantize(float value,int numb_bits)
...
-    //
-    //RestoreFourthQuatComponent(pEnt->qrot);
+#ifdef DEBUG_INPUT
+    fprintf(stderr,"\nupdates: %i\n",updates);
+#endif
     float pyr_angles[3];
-    toEulerAngle(src.qrot,pyr_angles[0],pyr_angles[1],pyr_angles[2]);
+    //glm::vec3 vec = toCoH_YPR(src.inp_state.direction);
+    pyr_angles[0] = 0.0f;
+    pyr_angles[1] = src.inp_state.camera_pyr.y; // set only yaw value
```
broxen: here we can do:
```
pyr_angles[1] = glm::yaw(src.inp_state.direction);
```
to achieve the same effect using the direction quat. `direction` seems better, as it's more forward-thinking in regards to the camera rotation issue.

broxen:
Despite what I expected, pulling this from `camera_pyr` or subsequently `direction` doesn't rotate the entity when the camera rotates. Why? I have no idea. haha. 

At first I thought `camera_pyr` was mislabeled, but it definitely changes value when the camera moves, however it also changes when the entity rotates. So I'm not really clear what that is.

You can test with the following slash commands in the chat window (NOTE: `/camrotate` simply flags whether or not the camera rotates or the entity rotates:
```
/turnleft 1
/camrotate 1
/camrotate 0
/turnleft 0
```

In a way, this makes this easy on us. In the future we simply check if the entity IS_FLYING and if so, then update vec.x, vec.y, and vec.z instead of setting x and z to 0 value.

Projects/CoX/Common/NetStructures/Entity.h
```diff
@@ -86,7 +86,11 @@ struct AngleRadians
     float v;
     int toIntegerForm() const
     {
-        return int((v + float(M_PI)) * 2048.0f / (2* float(M_PI)) + 0.5f);
```
broxen: I dont think this is being utilized anywhere, but the math shouldn't include `+ 0.5f`

Projects/CoX/Common/NetStructures/Entity.h
```diff
@@ -112,22 +116,21 @@ class InputStateStorage
...
-    int m_B_ang11_probably=0;
-    uint8_t m_received_server_update_id=0;
-    bool m_no_coll = false;
-    bool has_input_commit_guess=0;
-    bool pos_delta_valid[3]= {};
-    bool pyr_valid[3]= {};
+    int m_time_diff1                    = 0;
+    int m_time_diff2                    = 0;
+    uint8_t input_vel_scale             = 0; // TODO: Should be float?
+    float m_orientation_pyr[3]          = {}; // Stored in Radians
```
broxen: Replacing `m_A_ang11_probably` & `m_B_ang11_probably` with this float array. `A_ang11` is the angle of Pitch in radians, and `B_ang11` the angle of Yaw in radians.

These values are useful, because they stop updating when `/camrotate 0` is set. Meaning that they refer to the angle of your character's direction, not the camera_pyr. This solves our issue with camera rotation vs entity rotation.

Projects/CoX/Common/NetStructures/Entity.h
```diff
@@ -112,22 +116,21 @@ class InputStateStorage
...
+    uint8_t m_received_server_update_id = 0;
+    bool m_no_coll                      = false;
+    bool has_input_commit_guess         = 0;
+    bool pos_delta_valid[3]             = {};
+    bool pyr_valid[3]                   = {};
     glm::vec3 pos_delta;
     glm::quat direction;
```
broxen: `glm::quat direction` can probably go away entirely, as we don't really need it now that I've defined `m_orientation_angles`

I'll do more research and push any proposed changes.

Projects/CoX/Servers/MapServer/Events/InputState.cpp
```diff
@@ -45,6 +45,9 @@ InputStateStorage &InputStateStorage::operator =(const InputStateStorage &other)
     {
         if(other.pos_delta_valid[i])
             pos_delta[i] = other.pos_delta[i];
+            
+        if(other.m_orientation_pyr[i])
```
nemerle: I think that we should explicitly check `0==other.m_orientation_pyr[i]` since m_orientation_pyr is not a bool thingy
and code will be a bit clearer.

PR #168, opened 2017-11-28
[WIP] Calculate Hitpoints Base and Cap from Archetype

broxen:
Based upon the data at https://paragonwiki.com/wiki/Hit_Points

This PR aims to configure the default values for HP and HP cap based upon archetype. As far as I can tell this information isn't stored client-side, so we need to define it server-side.

**Marking this [WIP] because while it compiles cleanly, I can't figure out how to feed the character entity to my `setHPBase()` class in Character.cpp:**
https://github.com/broxen/Segs/blob/16a70841f482a168ac4b473f0da3f79898f3c4c1/Projects/CoX/Common/NetStructures/Character.cpp#L32

- [X] Add hitpoints_definitions.h
- [X] Shared class, level and origin with Entity (because comments mentioned it)
- [X] Add Hitpoint Mod and Cap Values
- [X] Added debug output for `m_unkn1 m_unkn2 m_unkn3` because I suspect these have to do with HP/Endurance
- [ ] Reference these values when first creating a character
- [ ] Eventually pull this data from a lua table in hitpoints.lua

I know that my feeble attempts at coding are probably more annoyance than actual help, but I'm trying to do what I can. Please know that I'm completely open to learning a better way to do what I'm trying to achieve here, or focusing on specific tasks that you feel need attention.

On that note, I know you're super busy Nem, but whenever you have a little bit of spare time, if you can point me in the right direction with the final piece of this, I'd appreciate it. I'll keep researching in the meantime and will update here if I discover my own solution.

### DIFF
--- a/Projects/CoX/Common/NetStructures/Character.cpp
+++ b/Projects/CoX/Common/NetStructures/Character.cpp
@@ -10,6 +10,7 @@
 #include "BitStream.h"
 #include "Costume.h"
 #include "GameData/keybind_definitions.h"
+#include "GameData/hitpoints_definitions.h"
 #include <QtCore/QString>
 #include <QtCore/QDebug>
 
@@ -67,6 +68,9 @@ void Character::serializefrom( BitStream &src)
     src.GetString(m_origin_name);
     m_unkn1 =src.GetFloat();
     m_unkn2 =src.GetFloat();
+    qDebug() << m_unkn1;
+    qDebug() << m_unkn2;
+    
     src.GetString(m_mapName);
     /*uint32_t unkn3 =*/ src.GetPackedBits(1);
     //uint32_t unkn4 = src.GetBits(32);
@@ -85,6 +89,7 @@ void Character::serializeto( BitStream &tgt) const
     //tgt.StorePackedBits(1,m_villain);
     tgt.StoreString(m_mapName);
     tgt.StorePackedBits(1,m_unkn3);
+    qDebug() << m_unkn3;
     //tgt.StorePackedBits(32,m_unkn4); // if != 0 UpdateCharacter is called
 }
 void Character::sendWindow(BitStream &bs) const
--- a/Projects/CoX/Common/NetStructures/Character.cpp
+++ b/Projects/CoX/Common/NetStructures/Character.cpp
@@ -29,7 +29,7 @@ Character::Character()
     m_supergroup_costume=false;
     m_sg_costume=nullptr;
     m_using_sg_costume=false;
-    m_current_attribs.m_HitPoints = 25;
+    m_current_attribs.m_HitPoints = 45; // setHPBase(self)
     m_max_attribs.m_HitPoints = 50;
     m_current_attribs.m_Endurance = 33;
     m_max_attribs.m_Endurance = 43;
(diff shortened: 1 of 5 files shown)


nemerle:
I'm pretty busy recently ( deploying a large system at a new location ) + working on shader support.

I'll take a look at this as soon as I can :)

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -67,6 +68,9 @@ void Character::serializefrom( BitStream &src)
     src.GetString(m_origin_name);
     m_unkn1 =src.GetFloat();
     m_unkn2 =src.GetFloat();
+    qDebug() << m_unkn1;
```
nemerle: I don't believe that this function is even called anywhere, feel free to remove it :smile:

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -85,6 +89,7 @@ void Character::serializeto( BitStream &tgt) const
     //tgt.StorePackedBits(1,m_villain);
     tgt.StoreString(m_mapName);
     tgt.StorePackedBits(1,m_unkn3);
+    qDebug() << m_unkn3;
     //tgt.StorePackedBits(32,m_unkn4); // if != 0 UpdateCharacter is called
```
nemerle: Same story with this function, it's also unused.

nemerle:
As for origins and archetypes, try running binConverter tool on classes.bin, there You can probably find many interesting things regarding Hp and End

Projects/CoX/Common/NetStructures/Character.cpp
broxen: What does the client expect here?
```
tgt.StorePackedBits(1,m_unkn3); // What are we storing and why?
```

Projects/CoX/Common/NetStructures/Character.cpp
broxen: Doesn't `GetFloat()` store some kind of value there? Or are you saying that we've ran out of values to store by this point?

Or am I misunderstanding the way the bitstream works =/

broxen:
ah binConverter is amazing! 

Classes.bin already calculates HP per level under `/EntityClasses/AttribMaxTable` and HP Cap at `/EntityClasses/AttribMaxMaxTable`, so no need for the majority of this commit. 

It looks like they changed the modifiers for several classes HP between I0 and I24+ (which is where I assume ParagonWiki culled it's data)

Since you've already serialized this data. How can I pull archetype specific data from that serialization? And more importantly, do I set the value in `m_max_attribs.m_HitPoints = 50;` under Character::Character() as it's set now? Seems like it should be set elsewhere, so I can reference the specific entity.

nemerle:
Well each archetype has it's own block in the classes bin.

I can add a few helper functions to retrieve class data per class, class/level etc.

Projects/CoX/Common/NetStructures/Character.cpp
nemerle: void Character::serializeto( BitStream &tgt) const is not used anywhere, I don't think it was ever used.

broxen: ahh. I understand. Yes `serializeto()` and `serializefrom()` aren't used anywhere. I suppose I thought they were sort of there for future use.

broxen:
Closing this pull request as it needs a full rewrite that leverages existing serialization

Issue #174, opened 2017-12-04
Targeting another entity produces `Unhandled command event type 32`


broxen:
### Summary
Targeting another entity produces `Unhandled command event type 32`

### Steps to Replicate
Log in with two entities. Click on another entity to target.

### SEGS version
Commit 8bc8fd3

broxen:
Attempted to resolve with #175, however it's still giving me the error message.

broxen:
#175 addresses this, but outputs `-1` for what I'm calling `entity_idx` when targeting other players and the following when hitting ESC to clear targets:
```
Debug   : Not all bits were consumed
Warning : Unhandled set target request: 1919
Console command received ü
Debug   : Unhandled game command: "Ã¼\u001B"
```

So maybe there is some kind of boolean isTargetingSelf, and then the idx of the target in there? I'm not sure how to parse the bitstream in `class SetTarget`

nemerle:
Will look into it tomorrow ( around 12h from now)

broxen:
Based upon our IRC chat, #175 now resolves this.

broxen:
Closing this as it's now a duplicate of #192

PR #175, opened 2017-12-04
MOAR Slash Commands, Titles, AFKs, ChatMessages, String Replacements, and other unlabeled-bits

broxen:
Many small change in this one. Just trying to move the ball forward.

- Attempting to track down unlabeled bits and their purpose. 
- Added slash commands for fly, dazed, jumppack, jumpheight, and backupspd, AFK, plus more.
- JumpHeight might work in tandem with u3 to enable "superjump" powers.
- Added ChatMessage for Request, Friends, Group, Supergroup, and associated Slash commands
- Pulled description and battle_cry out of entity. Fixed them so they update in your ID card properly. yay!
- Added handler for selecting different chat channels from the chat window input box. code 37
- TODO: still need to save things to the database, Titles, Desc, Battle_Cry, MapName, etc.
- TODO: Jumppack and Dazed still need their associated costume parts and special FX (stars above head) respectively.
- TODO: tried to add handler for client message 32, but it doesn't work, maybe I'm missing something?

The following slash commands now work in the client:
```
/f
/req
/sg
/g
/fly
/dazed
/jumppack
/setJumpHeight [float]
/setBackupspd [float]
/afk [msg]
/setTitles [the_flag] [generic] [origin] [special]
/setu1 [int]
/setu2 [int]
/setu3 [int]
/setu4 [int]
/setu5 [int]
/setu6 [int]
```

There are still many unknown bits here, but I have suspicions about these two in SendServerControlState()
```
        bs.StoreBits(1,u2);     // if 1/true entity anims stop, can still move, but camera stays. Slipping on ice?
        bs.StoreBits(1,u3);     // leaping? seems like the anim changes slightly?
```

EDIT: Typos.


broxen:
Oh neato. It carries my own comments over.

Well the crashing issues are fixed, but I always encourage testing. Also, please feel free to let me know if there are better ways of achieving my goals, and if you know what any of these unknown bits do.

nemerle:
This looks great.

Tomorrow should be the last day of that big at-work thing I'm currently busy with, so the day after I'l look through my resources for all those u1/u2 etc. bits and pieces and merge this

mobbyg:
Hey guys, thank you for the work you’ve done on this project. It’s really great reading the emails for the commits. I try and update server as soon as you post changes so I can offer any info I can, as I don’t know C++, but I have a For Dummies book now. 😜

broxen:
Great. Thank you both!

When you get a chance to look this over, can you tell me how I can call `toggleAFK()` when a player moves? Seems like it should go in `sendServerControlState()` but I can't access Entity there because it's `Const` (and rightly so)

Intended behavior is that movement (WASD, or mouse) will turn off AFK flag.

broxen:
Hope you don't mind me pushing another update to this PR.

- Added more slash commands for testing.
- Also, I realized that I need to set m_afk to false on movement, not Toggle. So if you can point me in the right direction for that, I'd appreciate it.
- Identified and began framework for opcodes 2, 8, 18, 29
- Also fixed opcode 16, which was somehow removed from MapEventFactory.cpp

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -114,10 +88,27 @@ void Character::setName(const QString &val )
     else
         m_name = "EMPTY";
 }
+
+void Character::setTitles(bool prefix, QString generic, QString origin, QString special)
+{
+    if(!prefix && generic.isEmpty() && origin.isEmpty() && special.isEmpty())
```
nemerle: consider writing this like ?
```cpp
m_has_titles = prefix || !generic.isEmpty() || !origin.isEmpty() || special.isEmpty();
if(!m_has_titles)
  return;
```

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -688,3 +679,27 @@ void ClientOptions::init()
     };
 }
 #undef ADD_OPT
+
+void Character::toggleAFK(const QString &msg)
+{
+    m_afk = !m_afk;
+    m_afk_msg = msg;
```
nemerle: Do we need to clean up m_afk_msg if client is no longer afk ?

Projects/CoX/Common/NetStructures/Character.h
```diff
@@ -175,13 +177,36 @@ const   QString &       getClass() const { return m_class_name; }
...
+        uint32_t            m_experience_debt   = 0;
+        uint32_t            m_influence         = 1;
+        bool                m_has_the_prefix    = false;
+        bool                m_has_titles        = false;
+        QString             m_titles[3];        // Generic, Origin, Special
+        QString             m_battle_cry;
+        QString             m_character_description;
+        bool                m_afk               = false;
+        QString             m_afk_msg;
+        // Experience Requirements Chart. Same for all classes.
```
nemerle: We should load it from the data files anyway :)
If everything goes right today, I'll be pushing code to access those kinds of things.

Projects/CoX/Servers/MapServer/MapEvents.h
```diff
@@ -376,6 +378,42 @@ class SetDestination final : public MapLinkEvent
...
+    {}
+    void serializeto(BitStream &bs) const override
+    {
+        bs.StorePackedBits(1,29);
+        bs.StorePackedBits(12,entity_idx);
+    }
+    void serializefrom(BitStream &bs) override
+    {
+        entity_idx  = bs.GetPackedBits(12);
+        boost_unk1  = bs.GetAvailSize();  // How many bits!?
```
nemerle: please mark things like this with
```cpp
//TODO:
```
Qtcreator has a very nice plugin that can list allo TODO: / BUG: etc. markers in the project :)

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -469,27 +481,58 @@ void MapInstance::on_window_state(WindowState * ev){
...
-            msg.startsWith("b ") || msg.startsWith("broadcast ");
+    return msg.startsWith("l ")     || msg.startsWith("local ")      ||
+            msg.startsWith("b ")    || msg.startsWith("broadcast ")  || msg.startsWith("y ")       || msg.startsWith("yell ")     ||
+            msg.startsWith("g ")    || msg.startsWith("group ")      || msg.startsWith("team ")    ||
+            msg.startsWith("sg ")   || msg.startsWith("supergroup ") ||
+            msg.startsWith("req ")  || msg.startsWith("request ")    || msg.startsWith("auction ") || msg.startsWith("sell ")     ||
+            msg.startsWith("f ")    || msg.startsWith("friends ")    ||
+            msg.startsWith("t ")    || msg.startsWith("tell ")       || msg.startsWith("w ")       || msg.startsWith("whisper ")  ||
+            msg.startsWith("p ")    || msg.startsWith("private ");
 }
```
nemerle: Maybe we could simplify this by using QStringList::contains here ?
```cpp
static const QStringList chat_prefixes = { "l","local","b","broadcast",/* the rest would follow */ };
QString prefix(msg.mid(0,indexOf(' '));
return chat_prefixes.contains(prefix);
```

Projects/CoX/Common/NetStructures/Character.cpp
broxen: Testing this: no, it doesn't seem like it matters. But my (admittedly terrible) memory tells me that the official server behavior was actually to save this msg between uses (during the same connection)

I think that because I used to go afk with msg, move (removing afk flag) and then set afk flag again without msg and it would have my old msg left in the system (probably client side memory).

I guess I could achieve this with a simple `if` statement, but is there any reason (memory) to clear this value instead?

Projects/CoX/Common/NetStructures/Character.cpp
broxen: Ooo. Much cleaner!

Projects/CoX/Servers/MapServer/MapEvents.h
broxen: Oh neato! I'll be more consistent with this then.

Projects/CoX/Common/NetStructures/Character.h
nemerle: Huh, You're right, I'd swear I've seen the exp table somewhere

Projects/CoX/Servers/MapServer/MapInstance.cpp
broxen: Thank you! I was struggling to figure out a better way to do this, and my brain just wasn't going anywhere.

Projects/CoX/Common/NetStructures/Character.h
nemerle: Yup, it's in experience.bin added issues #176

broxen: Seems like it would be, there's no real harm in exposing this info client side. And, you've looked through more of the `bin` files than me, so I'm sure you're right, it's there somewhere.

broxen:
Updated based upon feedback, and added additional TODO comments and cleaned up the comments overall.

Also, I cleaned up the code a tiny bit.

Projects/CoX/Servers/MapServer/Events/MapEventTypes.h
```diff
@@ -5,43 +5,48 @@ class MapEventTypes : public CRUD_EventTypes
...
+    EVENT_DECL(evClientResumedRendering     ,204)
+    EVENT_DECL(evCookieRequest              ,206)
+    EVENT_DECL(evSwitchTray                 ,208) // Switch Tray using left-right arrows next to power tray
+    EVENT_DECL(evEnterDoor                  ,209)
+    EVENT_DECL(evSetDestination             ,211)
+    EVENT_DECL(evWindowState                ,214)
+    EVENT_DECL(evInspirationDockMode        ,217)
+    EVENT_DECL(evPowersDockMode             ,218)
+    EVENT_DECL(evUseInspiration             ,229) // Hit F1, F2, F3, F4, F5 or Click on Inspiration in Tray
+    EVENT_DECL(evSetTarget                  ,232) // Click on another entity
```
nemerle: I'm almost 100% positive that command 32 is related to unqueing a power or inspiration execution:
Take a look at client code:
```cpp
void sendPacket32(Entity *a1)
{
    my_pakSendPackedBits(g_input_pak, 1, 32);
}
void  AbortQueuedPower()
{
    Entity *e = getControlledPlayerEntity();
    sendPacket32(e);
    inspirationUnqueue();
    g_cursor.target_world = 0;
    s_ptsActivated = nullptr;
}
```

Projects/CoX/Servers/MapServer/MapEvents.h
```diff
@@ -376,6 +378,43 @@ class SetDestination final : public MapLinkEvent
...
+    void serializeto(BitStream &bs) const override
+    {
+        bs.StorePackedBits(1,29);
+        bs.StorePackedBits(12,entity_idx);
+    }
+    void serializefrom(BitStream &bs) override
+    {
+        entity_idx  = bs.GetPackedBits(12);
+        boost_unk1  = bs.GetAvailSize();  // TODO: Not all bits were consumed
+    }
```
nemerle: Packet 29 - activate inspiration slot  ( it does not contain entity_idx)
packedbits(1,29);
packedbits(3,slot_idx);
packedbits(3,row_idx);

Projects/CoX/Servers/MapServer/MapEvents.h
```diff
@@ -454,7 +508,52 @@ class ChatReconfigure final : public MapLinkEvent
         m_chat_bottom_flags = bs.GetPackedBits(1);
     }
 };
+class PowersDockMode final : public MapLinkEvent
```
nemerle: Packet 18 is used when a secondary tray is toggled:
* packedBits(1,18)
* bits(1,toggle_secondary_tray)

Projects/CoX/Servers/MapServer/MapEvents.h
```diff
@@ -454,7 +508,52 @@ class ChatReconfigure final : public MapLinkEvent
...
+    }
+    void serializefrom(BitStream &bs)
+    {
+        dock_mode = bs.GetBits(32); // TODO: output -1
+        dock_unk1 = bs.GetBits(32); // TODO: output -1
+        dock_unk2 = bs.GetBits(32); // TODO: output -1
+        // TODO: Not all bits were consumed
+    }
+};
+class SwitchTray final : public MapLinkEvent
```
nemerle: Packet 8 contains serialized PowerTrayGroup

Projects/CoX/Servers/MapServer/Events/MapEventTypes.h
broxen: I actually read this a different way, informed by the client behavior during testing.

I think command 32 is still related to targeting, as we already have `EVENT_DECL(evAbortQueuedPower           ,233)` which fires when you hit 'Z' (bound to unqueue powers by default)

But I think `AbortQueuedPower` is a function called when you hit `ESC` and is not related to command 32, but simply calls it's send function.

This would make sense because hitting `ESC` should do all of the following:
1. set target to self 
```cpp
    Entity *e = getControlledPlayerEntity();
    sendPacket32(e);
```
2. unqueue any powers or inspirations you've activated `inspirationUnqueue();`
3. clear any active waypoints/markers on the map (presumably) `s_ptsActivated = nullptr;`

broxen: latest push appears to correct this behavior here, and supports my supposition in that hitting `ESC` triggers both `SetTarget` and `AbortQueuedPower` resolving #174

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -443,16 +432,16 @@ void Character::sendChatSettings(BitStream &bs) const
...
 {
-    bs.StoreBits(1,1);
-    bs.StoreString("Tz1");
-    bs.StoreString("Tz2");
-    bs.StoreString("Tz3");
+    bs.StoreBits(1,m_has_the_prefix);           // likely an index to a title prefix ( 0 - None; 1 - The )
+    storeStringConditional(bs, m_titles[0]);    // Title 1 - generic title (first)
+    storeStringConditional(bs, m_titles[1]);    // Title 2 - origin title (second)
+    storeStringConditional(bs, m_titles[2]);    // Title 3 - yellow title (special)
 }
```
nemerle: Sadly this is incorrect vs. the client code:
```cpp
void  receiveTitles(DbPacket *pak)
{
    Entity *e = my_getControlledPlayerEntity();
    e->ent_player2->titleThe = my_pakGetBits(pak, 1);
    strncpy(e->ent_player2->Title1, my_pakGetString(pak), 32);
    strncpy(e->ent_player2->Title2, my_pakGetString(pak), 32);
    strncpy(e->ent_player2->Title3, my_pakGetString(pak), 128);
}
```
So if You had to convert it to conditional strings, then we have a problem somewhere before the sendTitles

Projects/CoX/Servers/MapServer/MapEvents.h
broxen: This comment doesn't show as cleared, but is resolved in the latest push.

Projects/CoX/Servers/MapServer/Events/EntitiesResponse.cpp
```diff
@@ -278,33 +290,35 @@ void sendServerControlState(const EntitiesResponse &src,BitStream &bs)
...
-        bs.StoreBits(1,0); // key push bits ??
-        bs.StoreBits(1,0); // key push bits ??
-        bs.StoreBits(1,0); // key push bits ??
+
+        bs.StoreFloat(m_jump_height); // How high entity goes before gravity bring them back down. Set by leaping default = 0.1f
+        bs.StoreBits(1,m_is_flying);    // is_flying flag
+        bs.StoreBits(1,m_is_stunned);     // is_stunned flag (lacks overhead 'dizzy' FX)
+        bs.StoreBits(1,m_has_jumppack);  // jumpack flag (lacks costume parts)
+
+        bs.StoreBits(1,u2); // if 1/true entity anims stop, can still move, but camera stays. Slipping on ice?
```
nemerle: this is likely "controls disabled" bit

Projects/CoX/Common/NetStructures/Character.cpp
broxen: Ah, it just clicked for me how `storeStringConditonal` actually works. 

Not sure where I got the idea to use this, but I can confirm that `/setTitles bFlag generic origin special` actually works.

Maybe the client expects a has_titles flag before this group is received?

broxen: Can also confirm that prior to using Conditional, the client would crash sporadically and this was my best fix.

Projects/CoX/Servers/MapServer/Events/EntitiesResponse.cpp
broxen: That makes sense. Though you can still move?

broxen:
I'll try to rebase this and resolve those conflicts in EntitiesResponse.cpp late tonight.

broxen:
Note that this `update_part_2` section is completely wrong. Setting update_part_2 flag to true results in wild behavior, including flashing of location and the entity being relocated to `/loc 3.0 0.0 3.0`

Maybe this section has to do with teleporting?
- m_if_teleporting
- set client pos id (maybe a reference to current location?)
- speed? maybe actually pos, since currently it sends you to 3.0 0.0 3.0
- zeroes? maybe flags for FX? maybe a point in space to direct you to for teleport?
- set orientation upon arrival
- no idea on u6?

```
    // Used to force the client to a position/speed/pitch/rotation by server
    bs.StoreBits(1,update_part_2);
    if(update_part_2)
    {
        bs.StorePackedBits(1,u5);            // sets g_client_pos_id_rel default = 0
        storeVector(bs,spd);
        storeVectorConditional(bs,zeroes);      // vector3 -> speed ? likely; was zeroes

        storeFloatConditional(bs,0);         // Pitch not used ?
        storeFloatConditional(bs,ent->inp_state.camera_pyr.y); // Pitch
        storeFloatConditional(bs,0);         // Roll
        bs.StorePackedBits(1,u6);            // sets the lowest bit in CscCommon::flags default = 0
    }
```

Projects/CoX/Common/NetStructures/Character.cpp
broxen: I pulled this from sendTitles in Entity (which is duplicitous and I have now removed)

I've tried to send bs.StoreString() instead, but it crashes the client =/

Any ideas?

broxen:
Lots of updates here, but I couldn't figure out the sendTitles() storeStringCondtional() issue.

Projects/CoX/Common/NetStructures/Character.cpp
broxen: Also, why `e->ent_player2` instead of player1

broxen:
@nemerle any ideas on the sendTitles() issue? I can't make that work without storeStringConditional()

Also, is there anything else you need me to do to prepare this for commit? I think it's ready.

broxen:
Pulled many functions out of classes and added some utility functions to get Entity object by IDX or Name. Needs testing though, as I'm on a chromebook until the New Year.

Projects/CoX/Common/GameData/attrib_serializers.cpp
```diff
@@ -159,10 +163,14 @@ void serialize(Archive & archive, Parse_CharAttribMax & target)
...
-    archive(cereal::make_nvp("Stunned",target.m_Stunned));
+    archive(cereal::make_nvp("Stunned",target.m_is_stunned));
     archive(cereal::make_nvp("Sleep",target.m_Sleep));
     archive(cereal::make_nvp("Fly",target.m_is_flying));
+<<<<<<< HEAD
+    archive(cereal::make_nvp("Jumppack",target.m_has_jumppack));
+=======
     archive(cereal::make_nvp("Jumppack",target.m_Jumppack));
+>>>>>>> upstream/master
     archive(cereal::make_nvp("Teleport",target.m_Teleport));
```
nemerle: This looks like You've commited a file with a merge conflict :)

broxen: Not sure how that happened because I don't see it on my local branch? Also my local fork shows the variable as `m_has_jumppack` following our naming guidelines. Weird.

Anyway, looks like it's resolved :man_shrugging:

Projects/CoX/Servers/GameServer/GameServer.cpp
```diff
@@ -229,12 +229,36 @@ GameServerData &GameServer::runtimeData()
...
+    return pEnt;
+}
+
+Entity * GameServer::getEntityByIdx(const int32_t &idx)
+{
+    Entity *pEnt = nullptr;
+    EntityManager ref_em;
+
+    // Iterate through all active entities and return entity by idx
+    for (auto *em : ref_em.m_live_entlist)
```
broxen: Not sure how else to poll for entities in GameServer.

Projects/CoX/Servers/GameServer/GameServer.cpp
```diff
@@ -229,12 +229,36 @@ GameServerData &GameServer::runtimeData()
...
-        qDebug() << "Entity" << name << "does not exist.";
-}
-*/
+Entity * GameServer::getEntityByName(const QString &name)
+{
+    Entity *pEnt = nullptr;
+    EntityManager ref_em;
+
+    // Iterate through all active entities and return entity by name
+    for (auto *em : ref_em.m_live_entlist)
```
broxen: Not sure how else to poll for entities in GameServer.

Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
```diff
@@ -298,31 +301,30 @@ bool CharacterDatabase::update( Character *c )
         "specialtitle=:specialtitle, supergroupid=:supergroupid "
         "WHERE id=:id "
     */
-    m_prepared_char_update.bindValue(":id", getDbId(e));
```
broxen: Here I need to access the Entity that Character is associated with, because I need to access e->m_db_id;

How else could I do that?

Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
```diff
@@ -298,31 +301,30 @@ bool CharacterDatabase::update( Character *c )
...
-    m_prepared_char_update.bindValue(":xp", getXP(c));
+    m_prepared_char_update.bindValue(":hitpoints", c->m_current_attribs.m_HitPoints);
+    m_prepared_char_update.bindValue(":endurance", c->m_current_attribs.m_Endurance);
+    m_prepared_char_update.bindValue(":inf", getInf(*c));
+    m_prepared_char_update.bindValue(":xp", getXP(*c));
     m_prepared_char_update.bindValue(":xpdebt", c->m_experience_debt);
     m_prepared_char_update.bindValue(":xppatrol", c->m_experience_patrol);
-    m_prepared_char_update.bindValue(":alignment", e->m_alignment);
+    m_prepared_char_update.bindValue(":alignment", c->m_alignment);
     m_prepared_char_update.bindValue(":posx", e->pos.x);
```
broxen: Here we need to access entity again, to save pos values.

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -751,23 +755,23 @@ void MapInstance::on_console_command(ConsoleCommand * ev)
         }
     }
     else if(lowerContents == "fly") {
-        toggleFly(ent);
+        toggleFly(*ent);
```
broxen: Since pulling these functions out of the entity/character classes I have to pass a pointer to the entity/character, but I'm not understanding why. The functions all take references.

Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
```diff
@@ -155,8 +158,8 @@ bool CharacterDatabase::fill( Character *c)
 //    else if (results.num_rows()>1)
 //        ACE_ERROR_RETURN((LM_ERROR, ACE_TEXT ("(%P|%t) CharacterDatabase::fill query returned wrong number of results. %s failed.\n"), query.str().c_str()),false);
 
-    c->setDbId(m_prepared_char_select.value("id").toUInt());
-    c->setLevel((uint8_t)m_prepared_char_select.value("char_level").toUInt());
+    //setDbId(c,m_prepared_char_select.value("id").toUInt());
```
broxen: Here I need Entity again to ensure that the m_db_id value is set from the database.

Projects/CoX/Common/NetStructures/Character.h
```diff
@@ -216,19 +216,19 @@ const   QString &       getMapName() const { return m_mapName; }
 };
 
 void                serializeStats(const Character &src, BitStream &bs, bool sendAbsolute);
-uint32_t            getLevel(Character &src) const { return src.m_level; }
+uint32_t            getLevel(Character &src) { return src.m_level; }
```
broxen: Moved all of these out of the classes. But I've created some kind of linking error and I'm getting over 600 error messages that these functions are duplicated.

Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
```diff
@@ -60,13 +64,42 @@ void CharacterDatabase::on_connected(QSqlDatabase *db)
...
+                "last_online=:last_online, "
+                "hitpoints=:hitpoints, endurance=:endurance, "
+                "inf=:inf, xp=:xp, xpdebt=:xpdebt, "
+                "xppatrol=:xppatrol, alignment=:alignment, "
+                "posx=:posx, posy=:posy, posz=:posz, "
+                "orientp=:orientp, orienty=:orienty, orientr=:orientr, "
+                "title=:title, badgetitle=:badgetitle, "
+                "specialtitle=:specialtitle, supergroup_id=:supergroup_id, "
+                "options=:options, gui=:gui "
+                "WHERE id=:id ");
```
nemerle: Many of those fields should be serialized into CharacterData `blob`

Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
```diff
@@ -204,36 +246,137 @@ bool CharacterDatabase::create( uint64_t gid,uint8_t slot,Character *c )
...
-    m_prepared_char_insert.bindValue(7, c->m_mapName);
+
+    /*
+    ":char_level, :slot_index, :account_id, :char_name, :archetype, :origin, :description, "
+    ":battlecry, :current_map, :bodytype, :last_costume_id, :last_online, :hitpoints, :endurance, "
+    ":inf, :xp, :xpdebt, :xppatrol, :alignment, :posx, :posy, :posz, :orientp, :orienty, :orientr, "
+    ":title, :badgetitle, :specialtitle, :supergroup_id, :options, :gui"
+    */
+
+    m_prepared_char_insert.bindValue(":char_level", c->m_level);
```
nemerle:
```cpp
#include <iostream>
#include <cereal/archives/json.hpp>
#include <cereal/types/vector.hpp>
//......
  std::stringstream ss;
  {
    cereal::JSONOutputArchive archive( ss );
    CharacterData charData;
    archive( charData );
  }
  m_prepared_char_insert.bindValue(":char_data", QByteArray::fromStdString(charData.str()));
```

broxen: I'm trying to look at costume parts for inspiration here, but I'm failing to understand this. Can you point me towards an example?

I know that I need:
- new CharacterData class containing hp, end, m_current_attribs and m_max_attribs etc.
- new `serializefromdb()` and `serializetodb()` methods for serializing these values. I see an example in `Costume.cpp` for parts.

Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
```diff
@@ -204,36 +246,137 @@ bool CharacterDatabase::create( uint64_t gid,uint8_t slot,Character *c )
...
+    m_prepared_char_insert.bindValue(":orientr", 0);
+    m_prepared_char_insert.bindValue(":title", c->m_titles[0]);
+    m_prepared_char_insert.bindValue(":badgetitle", c->m_titles[1]);
+    m_prepared_char_insert.bindValue(":specialtitle", c->m_titles[2]);
+    m_prepared_char_insert.bindValue(":supergroup_id", c->m_supergroup_id);
+
     if(!doIt(m_prepared_char_insert))
         return false;
     int64_t char_id = m_prepared_char_insert.lastInsertId().toLongLong();
+    c->m_db_id = char_id;
```
broxen: Here I'm trying to pass the Character Table row uid to c->m_db_id. Later I try setting c->m_db_id to e->m_db_id.

Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
```diff
@@ -204,36 +246,137 @@ bool CharacterDatabase::create( uint64_t gid,uint8_t slot,Character *c )
...
+    m_prepared_costume_insert.bindValue(":id",quint64(char_id));
+    m_prepared_costume_insert.bindValue(":costume_index",uint32_t(0));
+    m_prepared_costume_insert.bindValue(":skin_color",uint32_t(cst->skin_color));
+    m_prepared_costume_insert.bindValue(":parts",costume_parts);
 
     if(!doIt(m_prepared_costume_insert))
         return false;
     return true;
 }
+bool CharacterDatabase::update( Character *c )
```
broxen: This still doesn't run and I believe it's because c->m_db_id (:id) is never defined. It's always 0.

Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
nemerle: An example from anim_serializers.cpp
```cpp
template<class Archive>
void serialize(Archive & archive, TextureAnim_Data & m)
{
    archive(cereal::make_nvp("Speed",m.speed));
    archive(cereal::make_nvp("Scale",m.stScale));
    archive(cereal::make_nvp("Type",m.scrollType));
    archive(cereal::make_nvp("flags",m.flags));
}

template
void serialize<cereal::JSONOutputArchive>(cereal::JSONOutputArchive & archive, TextureAnim_Data & m);
template
void serialize<cereal::JSONInputArchive>(cereal::JSONInputArchive & archive, TextureAnim_Data & m);
template
void serialize<cereal::BinaryInputArchive>(cereal::BinaryInputArchive & archive, TextureAnim_Data & m);
template
void serialize<cereal::BinaryOutputArchive>(cereal::BinaryOutputArchive & archive, TextureAnim_Data & m);
```

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -76,4 +77,10 @@ void World::updateEntity(Entity *e, const ACE_Time_Value &dT) {
         if(e->m_time_till_logout<0)
             e->m_time_till_logout=0;
     }
+    
+    if(e->m_type==Entity::ENT_PLAYER)
+    {
+        CharacterDatabase *char_db = AdminServer::instance()->character_db();
+            char_db->update(&e->m_char);
```
broxen: Is this the best place to put this?

nemerle: The best ? no. What we should do is a 'service thread' that periodically runs over all chars, and stores their state into db, what w should do in here, is mark the character state as 'dirty'.

But this will work for now ( add a ticket to fix/design a saner db-update code ) :)

Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
broxen: this works now, as expected

Projects/CoX/Servers/AdminServer/CharacterDatabase.cpp
broxen: this now works as intended

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
broxen: Actually this will not work for now. When more than one client is connected it crashes the server. My guess is that it tries to write to the dbase simultaneously?

broxen:
Current status of this PR:
- `/updatechar` works and updates character in database, but I cannot call `CharacterDatabase::update()` in WorldSimulation.cpp or the server will freeze when multiple clients are attached. :man_shrugging: 
- started fleshing out a CharacterData class based upon anim_serialization. I'm not confident that I'm heading in the right direction though. Maybe it's better to just build off of the existing Character class instead?
- `getEntity()` works but will not function from different map instances and still relies on the EntityManager object created by MapInstance (or MapClient). For now this should work fine, as we only have one instance, but in the future we'll need a better way!

TODO:
- [ ]  If I can get some help figuring out how to call `getEntity()` from the `CharacterDatabase::update()` method, I can save position information to the database. First steps towards persistence!
- [ ]  May I please have some additional guidance on the character stats serialization? Am I moving in the right direction?

Once those two items are resolved, this should be ready for commit.

broxen:
Identified a regression disabling entity rotation caused by converting m_orientation_pyr from float to vec3. But rotation is still choppy. Not sure how to speed up those updates.

broxen:
This compiles cleanly and everything seems to work. I think it's ready to commit with the following notices:
- I'm stuck on serializing to the database, I'm looking to Costume update, but missing something. So I think I've serialized everything in Character and Entity classes, but we're not currently doing anything with that JSON data, and frankly, I'm not sure how to call it elsewhere.
- I was unable to pull `m_target_idx` out of `inp_state` due to scope. InputStorage is unaware of the current entity. 
- I am unable to run `CharacterDatabase::update()` during `World::updateEntity()` because it freezes the clients with more than one character logged in.

There's still plenty to do, but I think this PR is "complete" and achieved it's original goals (and then some)

PR #204, opened 2017-12-24
Update README with more accurate and detailed information

### LINKED ISSUE #202: Documentation doesn't specify client command line arguments
Client must be run with ` -project SEGS` command line argument, otherwise it will output error message about patching using coh_updater.exe.

Docs should be updated to reflect this necessary step.

### LINKED ISSUE #203: Tasks for streamlining the initial installation process for newcomers
Many things in the readmes concerning the installation and configuration of the SEGS server and CoX client are either outdated or poorly explained. We should fix these problems so that newcomers can easily join the project.

### Segs/Projects/CoX/docs/README.md
- [x] Under Installation, steps 1 and 2 should be consolidated into 1a and 1b, to clarify that the user should do either one or the other.
- [x] OUTDATED: When the user builds the project, the `/data/` directory is not created by default, so the user will need to do it manually.
- [x] The dll's required for the utilities (like piggtool and authserver) to run are not created or moved to the out folder by default. References Issue #201. EDIT: This now only applies to `libACE.dll`.
- [x] Add more detailed instructions about how to operate piggtool. The user will need to open Windows Command Prompt, navigate to the out folder, and input this command (which is displayed incorrectly in the readme): `piggtool -x ./data/bin.pigg`. The user will then need to move the created bin folder into the `/data/` directory.
- [x] Include instructions on how to configure the CoX client to run without CohUpdater.exe. The user will need to create a shortcut to CoX.exe, and access its properties. In the Target field, the user will need to add `-project SEGS` to bypass the updater. Optional commands for windowed mode: `-fullscreen 0 -screen 1024 768`

erpmario:
Closes #202.
Closes #203.


Projects/CoX/docs/README.md
```diff
@@ -44,57 +44,87 @@ Note that all directories mentioned in this document are always relative to the
...
+3. Copy the PIGG files to SEGS.
 
-4. Copy the PIGG files to SEGS
+   The CoX client uses files ending in `.pigg` to store important game information. SEGS uses these same files to run the game world.
+   
+   Once installed, navigate to the SEGS folder. If you compiled from source, this will be in the `/out/` folder wherever you saved your Qt project. If you downloaded a release version, this will be wherever you extracted it to. Create a new folder and name it `data`.
+   
+   Navigate to the CoX folder and copy the contents of the `/piggs/` subfolder to the `/data/` directory of SEGS.
+
+4. Copy over required DLL files to SEGS.
```
broxen: These DLLs are only required on windows.

Perhaps we should skip this step in the readme, and automatically copy these DLLs over during the build process for Windows platforms only.

Projects/CoX/docs/README.md
```diff
@@ -44,57 +44,87 @@ Note that all directories mentioned in this document are always relative to the
...
+   - Qt5Gui.dll
+   - Qt5Network.dll
+   - Qt5Sql.dll
+   - Qt5Widgets.dll
 
-5. Using the piggtool located in the main SEGS folder, extract BIN files from bin.pigg
+7. Using the piggtool located in the main SEGS folder, extract the BIN files from bin.pigg.
 
-   SEGS requires that we extract bin.pigg to the `/data/bin/` folder. From the console:
+   SEGS requires that we extract bin.pigg to the `/data/bin/` folder. Open a command prompt console. Using the `cd` command, navigate to the SEGS folder, which should contain `piggtool.exe`. To confirm you are in the correct directory, simply enter `piggtool` in the console. If you receive a help menu about its usage, you are in the correct folder. Type the following into the console and press Enter:
```
broxen: Let's remove the .exe extension and make this more platform agnostic. The command should work in either platform without the extension.

Projects/CoX/docs/README.md
```diff
@@ -44,57 +44,87 @@ Note that all directories mentioned in this document are always relative to the
...
-   ./piggtool -x PATH_TO_COH_INSTALL/piggs/bin.pigg
+   piggtool -x ./data/bin.pigg
    ```
+   
+   You should then see a folder called `bin` be created in your SEGS directory. Simply move it into the `data` folder.
+   
 6. Point the CoX client towards your server.
 
-   Run `CoX-Auth.exe` and set the IP address to that of your server. If running your server locally this CANNOT be 127.0.0.1 (localhost) and must point to your actual IP address on the network.
+   Run `CoX-Auth.exe` from the CoX directory and set the IP address to that of your server. If running your server locally this CANNOT be 127.0.0.1 (localhost) and must point to your actual IP address on the network.
```
broxen: This will remain an exe even in Linux (users will run it with Wine)

Projects/CoX/docs/README.md
```diff
@@ -114,12 +144,16 @@ CONFIGURATION
 
 Configuration of your SEGS server is done using the `settings.cfg` file saved in the SEGS folder, and must be completed before running your server for the first time. See below for an exhaustive breakdown of server configuration.
 
-1. Setup ./settings.cfg
+1. Acquire required DLL file.
```
broxen: This will not be necessary on Linux. Maybe we should make a note here about that.

Projects/CoX/docs/README.md
```diff
@@ -130,14 +164,14 @@ Configuration of your SEGS server is done using the `settings.cfg` file saved in
    ##### [AuthServer]
    This section contains settings specific to the authentication server `authserver`. 
    - `listen_addr` -- This is the IP of the server that clients will connect to, followed by `:2106` which is the default listening port. If running your server locally this CANNOT be 127.0.0.1 (localhost) and must point to your actual IP address on the network. The port never changes.
-   - `max_character_slots` -- This is the default number of character slots allowed for each account.
```
broxen: I think this needs to be under Authserver, as it's set per account and checked by the Authserver during log in

Projects/CoX/docs/README.md
broxen: You can see how I've copied some files here: https://github.com/Segs/Segs/blob/master/Projects/CoX/CMakeLists.txt

We can add an if statement to check for a windows build and copy these in that case.

We can also add the ./data folder this way

erpmario: That would be great. Which make lists would these fall under? I'll try and look into starting this.

Projects/CoX/docs/README.md
erpmario: Ah, then maybe it's the config file that's incorrect. By default, it appears under GameServer as max_account_slots.

erpmario:
I made notes about the DLL files only being required in Windows environments.

Projects/CoX/docs/README.md
erpmario: Taken care of.

Projects/CoX/docs/README.md
broxen: Oh perhaps that's true.

Let me look to see if we're even really pulling from the cfg at all.

Projects/CoX/docs/README.md
erpmario: Sounds good, I'll make the change. I'm quite inexperienced with Linux, so it's nice to have another set of eyes.

Projects/CoX/docs/README.md
broxen: You're correct. This should be pulled from GameServer. My mistake.

The relevant code is on line 141 of GameServer.cpp

broxen:
This looks ready for commit to me.

We can open an issue for automating the DLLs and data folder creation, then submit as another PR.

Great work Eric! Pleasure having you here.

Projects/CoX/docs/README.md
```diff
@@ -44,57 +44,89 @@ Note that all directories mentioned in this document are always relative to the
...
+   
+   Once installed, navigate to the SEGS folder. If you compiled from source, this will be in the `/out/` folder wherever you saved your Qt project. If you downloaded a release version, this will be wherever you extracted it to. Create a new folder and name it `data`.
+   
+   Navigate to the CoX folder and copy the contents of the `/piggs/` subfolder to the `/data/` directory of SEGS.
 
-4. Copy the PIGG files to SEGS
+4. Copy over required DLL files to SEGS.
 
-   The CoX client uses files ending `.pigg` to store important game information. SEGS uses these same files to run the game world. Once installed, navigate to the CoX folder and copy the contents the `/piggs/` subfolder to the `/data/` directory of SEGS.
+   NOTE: This step is only required on Windows.
```
nemerle: Those dlls are related to mingw build of segs, under visual studio required dlls will be a bit different, I'm not sure we should mention this here ?

Projects/CoX/docs/README.md
```diff
@@ -44,57 +44,89 @@ Note that all directories mentioned in this document are always relative to the
...
-5. Using the piggtool located in the main SEGS folder, extract BIN files from bin.pigg
+5. Using the piggtool located in the main SEGS folder, extract the BIN files from bin.pigg.
 
-   SEGS requires that we extract bin.pigg to the `/data/bin/` folder. From the console:
+   SEGS requires that we extract bin.pigg to the `/data/bin/` folder. Open a command prompt console. Using the `cd` command, navigate to the SEGS folder, which should contain `piggtool`. To confirm you are in the correct directory, simply enter `piggtool` in the console. If you receive a help menu about its usage, you are in the correct folder. Type the following into the console and press Enter:
 
    ```
-   ./piggtool -x PATH_TO_COH_INSTALL/piggs/bin.pigg
+   piggtool -x ./data/bin.pigg
    ```
```
nemerle: by default the pigg files are in the client's pigg directory so ```data/bin.pigg``` does not look right?

Projects/CoX/docs/README.md
nemerle: Until we have a cross-platform client 😛

Projects/CoX/docs/README.md
```diff
@@ -44,57 +44,89 @@ Note that all directories mentioned in this document are always relative to the
...
-   Run `CoX-Auth.exe` and set the IP address to that of your server. If running your server locally this CANNOT be 127.0.0.1 (localhost) and must point to your actual IP address on the network.
+   Run `CoX-Auth.exe` from the CoX directory and set the IP address to that of your server. If running your server locally this CANNOT be 127.0.0.1 (localhost) and must point to your actual IP address on the network.
+
+7. Bypass the CohUpdater requirement.
 
-7. Continue Server Setup
+   The Issue 0 client requires that you launch the game with a program called `CohUpdater.exe`. To bypass this, create a shortcut to CoX.exe and place it wherever you'd like. Access its properties, and in the `Target:` field, add the following: `-project SEGS`.
+   
+   If you would like to run the client in windowed mode for debugging purposes, add these optional instructions: `-fullscreen 0 -screen 1024 768`.
+
```
nemerle: We can also run the client in windowed mode by setting a ```fullscreen``` registry key to 0

Projects/CoX/docs/README.md
```diff
@@ -114,12 +146,18 @@ CONFIGURATION
...
 Configuration of your SEGS server is done using the `settings.cfg` file saved in the SEGS folder, and must be completed before running your server for the first time. See below for an exhaustive breakdown of server configuration.
 
-1. Setup ./settings.cfg
+1. Acquire required DLL file.
+
+   NOTE: This step is only required on Windows.
+
+   Much like the utilities, the authserver requires a `.dll` file to run. Navigate to the folder containing the build of your project. It should be named something like this: `build-Segs-Desktop_Qt_YOUR_VERSION_NUMBER_MinGW_32bit-Default`. From this folder, navigate to: `./3rd_party/built/lib/`. Copy the file `libACE.dll` to the SEGS directory.
+
+2. Setup ./settings.cfg.
```
nemerle: A good step towards a well contained build would be copying the 3rd_party built dlls to out directory.

nemerle: Added #205

Projects/CoX/docs/README.md
broxen: This is at my direction. I was under the impression that the server needed a copy of all the pigg files in the `./data/` directory. Is this not the case?

Projects/CoX/docs/README.md
erpmario: Right, that makes sense. Although, I was under the impression that we were assuming the use of Qt for this readme.

Projects/CoX/docs/README.md
erpmario: I can clarify this. The readme instructs the user in a previous step to copy over the files in the client's `piggs` directory to the SEGS `data` folder. This includes a copy of `bin.pigg`. So, it doesn't matter which copy of the file the user extracts from, since they're equivalent. It just seemed easier to use this much more concise command rather than have the user type a much longer directory.

nemerle: Ok, I agree that this makes sense for now.  We could use some kind of GUI tool to locate/extract client resources though :)

Projects/CoX/docs/README.md
nemerle: Ahh, but one can use Visual Studio version of Qt :)

Projects/CoX/docs/README.md
erpmario: Oh yeah, absolutely. Streamlining this process with automated tools is a great goal for us. Until then, this is a slightly more involved method.

Projects/CoX/docs/README.md
erpmario: Oh! I didn't know that. Yeah, I agree then. We should make it a priority to copy the required dll's over automatically.

erpmario:
Hold off on reviewing this for now. I need to work out some bugs with ACE. I'm glad I tried to run this off of a fresh project.

erpmario:
More likely than not, I'll close this PR once I have a working algorithm for DLL transfer. I have a different local branch which is more up to date and refined. I'll leave this open for now though.

erpmario:
Alright, I'm closing this PR to open a new one for a new branch. This new branch is cleaner and fixes all of the problems we ran into.

PR #232, opened 2018-01-07
DbTool edits and database updates

### LINKED ISSUE #183: Move all user settings to a single blob column.
We do not need nor actually want to have each and every setting accessible in the database.
So we should just serialize user settings using cereal and store them in per-user ```account``` table, or per-character ```characters``` in an 'options' column.

erpmario:
Closes #183.

- Make `dbtool` create the `segs` and `segs_game` databases from source instead of copying existing databases.
- In CMake, copy the `pgsql` and `sqlite` directories to the `default_dbs` directory, instead of the two premade databases. This make the premade databases unnecessary, so delete them.
- Add a `db_version` field to each database's `table_versions` to catalogue the version of the entire database.
- Add a `salt` BLOB to the `segs` `accounts` table to store the store the salt string used in hashing the password.
- Add an `options` and `gui` BLOB to the `segs_game` `accounts` table, and remove the `options` and `gui` tables themselves.


Projects/CoX/Utilities/dbtool/main.cpp
```diff
@@ -62,13 +62,73 @@ void errorHandler(QtMsgType type, const QMessageLogContext &context, const QStri
...
+        QFile db_file(path);
+        QSqlDatabase segs_db(QSqlDatabase::addDatabase("QSQLITE"));
+        QSqlQuery query(segs_db);
+        if(db_file.exists())
+            db_file.remove();
+        segs_db.setDatabaseName(path);
+        segs_db.open(); // /*INSERT INTO accounts VALUES(1,'segsadmin',1,'2017-11-11 17:41:19',X'7365677331323300000000000000');*/ <- ignore this
+        if(scriptFile.open(QIODevice::ReadOnly))                                                // Execute each command in the source file.
+        {
+            // The SQLite driver executes only a single (the first) query in the QSqlQuery.
```
nemerle: Here we should start transaction:
```cpp
bool transaction_started = segs_db.transaction();
```

Projects/CoX/Utilities/dbtool/main.cpp
```diff
@@ -62,13 +62,73 @@ void errorHandler(QtMsgType type, const QMessageLogContext &context, const QStri
...
+            {
+                if(queryTxt.trimmed().isEmpty())
+                {
+                    continue;
+                }
+                if(!query.exec(queryTxt))
+                {
+                    qFatal(QString("One of the query failed to execute."
+                                " Error detail: " + query.lastError().text()).toLocal8Bit());
+                }
```
nemerle: If we failed at one query, the whole database is 'bad'
we should segs_db.rollback(); maybe even remove the created file ?

Projects/CoX/Utilities/dbtool/main.cpp
```diff
@@ -62,13 +62,73 @@ void errorHandler(QtMsgType type, const QMessageLogContext &context, const QStri
...
+                query.finish();
+            }
```
nemerle: At this point things went well, we should segs_db.commit();

Projects/CoX/Utilities/dbtool/main.cpp
```diff
@@ -62,13 +62,73 @@ void errorHandler(QtMsgType type, const QMessageLogContext &context, const QStri
     }
 }
 
+void createDatabases()
+{
+    QDir db_dir(QDir::currentPath());
+    qInfo() << "Creating database files...";
+    QStringList db_files;
+    db_files << db_dir.currentPath() + "/default_dbs/sqlite/segs_sqlite_create.sql" << db_dir.currentPath() + "./default_dbs/sqlite/segs_game_sqlite_create.sql";
+//    if(nofile)
```
nemerle: maybe we could use something like:
```cpp
QMap<QString,QString> files_and_targets = {
{ db_dir.currentPath() + "/default_dbs/sqlite/segs_sqlite_create.sql", db_dir.currentPath() + "/segs"};
{db_dir.currentPath() + "./default_dbs/sqlite/segs_game_sqlite_create.sql"}, db_dir.currentPath() + "/segs_game" };
```
and then iterate over `files_and_targets` 
should remove the need for `i` variable;

Projects/CoX/Utilities/dbtool/main.cpp
```diff
@@ -62,13 +62,73 @@ void errorHandler(QtMsgType type, const QMessageLogContext &context, const QStri
...
+//        db_files << db_dir.currentPath() + "/default_dbs/sqlite/segs_sqlite_create.sql" << db_dir.currentPath() + "./default_dbs/sqlite/segs_game_sqlite_create.sql";
+//    else
+//        db_files = args;
+    int i = 0;
+    for(const QString &db_file_string : db_files)
+    {
```
nemerle:
```cpp
for(const std::pair<QString,QString> &source_and_target : files_and_targets ) {
  const QString &source_file(source_and_target.first);
  const QString &target_file(source_and_target.second);
```

Projects/CoX/Utilities/dbtool/main.cpp
erpmario: This is already handled by the `.sql` source files, but you knew that already. Gotta be clear on record-keeping though!

Issue #208, opened 2017-12-27
Check user access level for slash commands


broxen:
User access level is currently set to `9` (the highest) by default and is not checked when using slash commands.

We should check against when slash commands are used. We may need to check per command, as some commands can be run by regular users.

nemerle:
Possible steps:
1. Define a 
```cpp
struct SlashCommand
{
    QStringList m_valid_prefixes;
    std::function<void(const QString &,Entity *)> m_handler;
    int m_required_access_level;
};
```
2. Convert all current slash commands to `SlashCommand`
3. Add a generic slash command handler, that will find the correct command, and check if given player has access to it.

broxen:
I'll tackle this next, along with moving the growing list of slash commands out of MapInstance.cpp into a SlashCommands.cpp file.

I'll add #216 as it's intimately related.

Let me know if there are any other improvements that can be made here.

broxen:
This is closed as of #264

Issue #41, opened 2017-04-29
Chatting features


ldilley:
This needs filling out:

ChatServerFeatures

broxen:
Most of these are implemented now, as of #175

broxen:
We might be able to close this?

nemerle:
Yup, closing this. When we are implementing new chat features, we'll just add feature-specific issues

PR #264, opened 2018-02-07
Moved slash commands to new class and improved ChatMessage/InfoMessageCmd

### LINKED ISSUE #216: Combine ChatMessage and InfoMessage methods
Combine ChatMessage and InfoMessage methods into one Send message() function.

Psuedocode
```
SendMessage(type,source,target,msg)
```

broxen:
Closes #216, #208, and relates to #41.

- Moved all slash commands to SlashCommand class
- Improved/streamlined ChatMessage (#216)
- Improved/streamlined InfoMessage (#216)
- Added `/help` to view all available slash commands
- Required appropriate access level to use slash commands (#208)
- Added Server MOTD dialog box on character spawn.
- Added motd.smlx for storing server MOTD
- Added`/setSuperGroupID` and `/setTeamID` to test supergroup/team chats
- Added a test smlx for tutorial mission. Test with `/smlx tutorial` or `/smlx test`
- Made `m_clients` and `m_scripting_interface` public so we can utilize them in `/whoall` and `/script`

TODO: nothing that I can think of. I worked out any kinks, and this seems to be a huge improvement to the way thing things work.


Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -26,6 +31,54 @@ void    setSpeed(Entity &e, float v1, float v2, float v3) { e.m_spd = {v1,v2,v3}
 void    setBackupSpd(Entity &e, float val) { e.m_backup_spd = val; }
 void    setJumpHeight(Entity &e, float val) { e.m_jump_height = val; }
 void    setUpdateID(Entity &e, uint8_t val) { e.m_update_id = val;}
+
+void    setTeamID(Entity &e, uint8_t val)
+{
```
nemerle: maybe rename `val` argument to team_id ?

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -95,6 +148,25 @@ int getEntityClassIndex(bool is_player, const QString &class_name)
 
 }
 
+void sendServerMOTD(Entity *e)
+{
+    MapClient *src = e->m_client;
+
+    QString fileName("scripts/motd.smlx");
+    QFile file(fileName);
+    if(QFileInfo::exists(fileName) && file.open(QIODevice::ReadOnly | QIODevice::Text))
```
nemerle: here QFileInfo::exists could be replaced with file.exists()

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -355,6 +356,8 @@ void MapInstance::on_create_map_entity(NewEntity *ev)
     setMapName(cl->char_entity()->m_char,name());
     lnk->set_client_data(cl);
     lnk->putq(new MapInstanceConnected(this,1,""));
+
+    sendServerMOTD(cl->char_entity());
```
broxen: This currently crashes the server on NEW character creation only, as m_client is undefined. I have a workaround on my local clone that checks for m_client to avoid this, but I can't figure out how to fix the underlying issue. I've tried adding `cl-char_data(e);` in the new char block just above this, but it doesn't seem to work.

Is there simply a better place for this? Shouldn't m_client be defined for new entities as well?

broxen: Added `cl->char_entity(e);` as it seems like we need it anyway. It doesn't resolve the issue. It seems like `src->addCommandToSendNextUpdate(std::unique_ptr<StandardDialogCmd>(dlg));` doesn't fire for new Characters, but only existing chars from the database. So what's different about new Characters that it won't fire?

Immediately using the `/motd` command will work, even with new chars.

broxen:
@nemerle, I know you're traveling, so no rush at all, but I wanted to let you know that this is ready for Final Review and Merge whenever you're back next week.

Thanks in advance!

PR #266, opened 2018-03-03
Resolve OSX Travis CI build issues:

MaxRis:
- Workaround to symlink qt5 installed with brew. Resolves mkspecs not found issue.
- ACE external project sources re-packed with CMakeList.txt updated to bundle config files with latest OSX versions.
- Lib names suffix corrections for OS X (dylib shared libs suffix is expected)
- mutex variable renamed to avoid naming collision on OS X


nemerle:
Thank You for this :)

PR #301, opened 2018-04-08
Error out when passing the wrong version into versioned serializers

### LINKED ISSUE #298: Report errors/warning in all serialization functions that are using versioned serialization
As an example:
```cpp
CEREAL_CLASS_VERSION(EntityData, 2); // register EntityData class version

template<class Archive>
void serialize(Archive & archive, EntityData &ed, uint32_t const version)
{
    archive(cereal::make_nvp("AccessLevel",ed.m_access_level));
    archive(cereal::make_nvp("OriginIdx",ed.m_origin_idx));
    archive(cereal::make_nvp("ClassIdx",ed.m_class_idx));
    archive(cereal::make_nvp("Position",ed.pos));
    archive(cereal::make_nvp("Orientation",ed.m_orientation_pyr));
}```
`version` argument is not used in this function, and it really should be. At least to verify things:
```cpp
    if(version==2)
    {
        archive(cereal::make_nvp("AccessLevel",ed.m_access_level));
        archive(cereal::make_nvp("OriginIdx",ed.m_origin_idx));
        archive(cereal::make_nvp("ClassIdx",ed.m_class_idx));
        archive(cereal::make_nvp("Position",ed.pos));
        archive(cereal::make_nvp("Orientation",ed.m_orientation_pyr));
    }
    else
    {
        qCritical() << "Failed to serialize EntityData, incompatible serialization format version"<<version;
    }
```
The same thing needs to be done to all other such functions

teronis84:
Closes #298.

My C++ is a little bit rusty but this should fit the bill. These are the only 2 versioned serialization functions I could find.

- Display an error if the serialization version argument does not match the version of the object to be serialized
- Add macro / constant for current object version

### DIFF
--- a/Projects/CoX/Common/GameData/chardata_serializers.cpp
+++ b/Projects/CoX/Common/GameData/chardata_serializers.cpp
@@ -4,34 +4,42 @@
 #include "DataStorage.h"
 #include "serialization_common.h"
 
-CEREAL_CLASS_VERSION(CharacterData, 2); // register CharacterData class version
+# define CHARACTER_DATA_CLASS_VERSION 2
+CEREAL_CLASS_VERSION(CharacterData, CHARACTER_DATA_CLASS_VERSION); // register CharacterData class version
 
 template<class Archive>
 void serialize(Archive & archive, CharacterData &cd, uint32_t const version)
 {
-    archive(cereal::make_nvp("Level",cd.m_level));
-    archive(cereal::make_nvp("CombatLevel",cd.m_combat_level));
-    archive(cereal::make_nvp("XP",cd.m_experience_points));
-    archive(cereal::make_nvp("Debt",cd.m_experience_debt));
-    archive(cereal::make_nvp("PatrolXP",cd.m_experience_patrol));
-    archive(cereal::make_nvp("Influence",cd.m_influence));
-    archive(cereal::make_nvp("HasTitles",cd.m_has_titles));
-    archive(cereal::make_nvp("ThePrefix",cd.m_has_the_prefix));
-    archive(cereal::make_nvp("Titles",cd.m_titles));
-    archive(cereal::make_nvp("BattleCry",cd.m_battle_cry));
-    archive(cereal::make_nvp("Description",cd.m_character_description));
-    archive(cereal::make_nvp("AFK",cd.m_afk));
-    archive(cereal::make_nvp("AfkMsg",cd.m_afk_msg));
-    archive(cereal::make_nvp("LFG",cd.m_lfg));
-    archive(cereal::make_nvp("Alignment",cd.m_alignment));
-    archive(cereal::make_nvp("LastCostumeID",cd.m_last_costume_id));
-    archive(cereal::make_nvp("LastOnline",cd.m_last_online));
-    archive(cereal::make_nvp("Class",cd.m_class_name));
-    archive(cereal::make_nvp("Origin",cd.m_origin_name));
-    archive(cereal::make_nvp("MapName",cd.m_mapName));
-    archive(cereal::make_nvp("SuperGroupCostume",cd.m_supergroup_costume));
-    archive(cereal::make_nvp("UsingSGCostume",cd.m_using_sg_costume));
-    archive(cereal::make_nvp("CurrentChatChannel",cd.m_cur_chat_channel));
+    if (version == CHARACTER_DATA_CLASS_VERSION)
+    {
+        archive(cereal::make_nvp("Level",cd.m_level));
+        archive(cereal::make_nvp("CombatLevel",cd.m_combat_level));
+        archive(cereal::make_nvp("XP",cd.m_experience_points));
+        archive(cereal::make_nvp("Debt",cd.m_experience_debt));
+        archive(cereal::make_nvp("PatrolXP",cd.m_experience_patrol));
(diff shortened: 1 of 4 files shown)


Projects/CoX/Common/GameData/chardata_serializers.cpp
```diff
@@ -4,34 +4,42 @@
 #include "DataStorage.h"
 #include "serialization_common.h"
 
-CEREAL_CLASS_VERSION(CharacterData, 2); // register CharacterData class version
+# define CHARACTER_DATA_CLASS_VERSION 2
```
nemerle: The code looks good overall , but instead of a macro here, put a version `contexpr` in struct body
```cpp
struct CharacterData 
{
    static const constexpr uint32_t class_version = 2;
    ...
}
```
and then using it as:
```cpp
CEREAL_CLASS_VERSION(CharacterData, CharacterData::class_version)
```
This will keep the typesystem on our side :smile:

Projects/CoX/Common/GameData/chardata_serializers.cpp
```diff
@@ -4,34 +4,42 @@
...
+        archive(cereal::make_nvp("Origin",cd.m_origin_name));
+        archive(cereal::make_nvp("MapName",cd.m_mapName));
+        archive(cereal::make_nvp("SuperGroupCostume",cd.m_supergroup_costume));
+        archive(cereal::make_nvp("UsingSGCostume",cd.m_using_sg_costume));
+        archive(cereal::make_nvp("CurrentChatChannel",cd.m_cur_chat_channel));
+    }
+    else
+    {
+        qCritical() << "Failed to serialize CharacterData, incompatible serialization format version " << version;
+    }
```
nemerle: And a comment from @broxen "... PR looks good. Maybe flip the logic to remove the indent? Otherwise awesome!"

so in here we'll get a fail-first-and-return-early:
```cpp
if (version != CharacterData::class_version)
{
    qCritical() << "Failed to serialize CharacterData, incompatible serialization format version " << version;
    return;
}
```

Projects/CoX/Common/GameData/chardata_serializers.cpp
```diff
@@ -4,34 +4,42 @@
...
-    archive(cereal::make_nvp("Alignment",cd.m_alignment));
-    archive(cereal::make_nvp("LastCostumeID",cd.m_last_costume_id));
-    archive(cereal::make_nvp("LastOnline",cd.m_last_online));
-    archive(cereal::make_nvp("Class",cd.m_class_name));
-    archive(cereal::make_nvp("Origin",cd.m_origin_name));
-    archive(cereal::make_nvp("MapName",cd.m_mapName));
-    archive(cereal::make_nvp("SuperGroupCostume",cd.m_supergroup_costume));
-    archive(cereal::make_nvp("UsingSGCostume",cd.m_using_sg_costume));
-    archive(cereal::make_nvp("CurrentChatChannel",cd.m_cur_chat_channel));
+    if (version == CHARACTER_DATA_CLASS_VERSION)
```
broxen: This PR is great! Great job!

Let's flip the logic here, to reduce indentation. Change to check class_version before the `archive()` functions like so:
```
if (version != CHARACTER_DATA_CLASS_VERSION)
{
    qCritical() << "Failed to serialize CharacterData, incompatible serialization format version " << version;
    return;
}

// archive()... etc
```

Then you don't need the else statement at all.

You can push those changes to the same branch and it will automatically update this PR with your revisions! After that I'll review and merge.

Great work! Great first PR! Welcome aboard! Hope we see you in discord.segs.io!

Projects/CoX/Common/GameData/entitydata_serializers.cpp
```diff
@@ -4,16 +4,24 @@
...
 
 template<class Archive>
 void serialize(Archive & archive, EntityData &ed, uint32_t const version)
 {
-    archive(cereal::make_nvp("AccessLevel",ed.m_access_level));
-    archive(cereal::make_nvp("OriginIdx",ed.m_origin_idx));
-    archive(cereal::make_nvp("ClassIdx",ed.m_class_idx));
-    archive(cereal::make_nvp("Position",ed.pos));
-    archive(cereal::make_nvp("Orientation",ed.m_orientation_pyr));
+    if (version == ENTITY_DATA_CLASS_VERSION)
```
broxen: Can use the same trick here, to reduce indentation.

teronis84:
Thank you both for the warm welcome, the feedback and for taking it easy on someone who hasn't done anything in C++ for a few years!  I think what you guys are doing is awesome so I'll definitely be following with interest and contributing when I can.

I've made the changes as requested, definitely esier to scan without the indentation. If you'd prefer a different spacing style in the header files please let me know and I'll handle it. I did notice that without the definitions for the `class_version` variables I was receiving a linker error - but I'm not sure if there was a different way I should've solved that.

broxen:
I'll leave this open for @nemerle to answer your question regarding the version vars. I think it's correct though. Great job and welcome aboard!

nemerle:
You are correct, that in this particular case `class_version` needs a storage (hence the linker complaints). This is usually the case when somewhere, somewhen the compiler needs to take an address of this variable. ( I think cereal's version macro might cause this ? ). Otherwise great work :)

PR #305, opened 2018-04-09
Since auto-account creation is going away, the db tool needs to step in

nemerle:
This adds a command argument to the dbtool utility
* `create`  creates the database, same as the dbtool did before
* `adduser -l login -p password`   adds a new account to the database


broxen:
Can we add  access_level to that? So we can create admin accounts?

nemerle:
sure thing :)

Issue #314, opened 2018-04-11
New dbtool doesn't add associated account row to segs_game database.


broxen:
New dbtool doesn't add associated account row to segs_game database for new accounts added with `adduser`

### Steps to reproduce behavior/symptoms
Build from latest master. Build completely new database files. Add new account using `adduser` command.

### Results
Client crashes on `assert(!"Unknown event encountered in dispatch.");` and `Debug   : GameEntryError: getCharsFromDb error!`

### Resolution
Manually adding associated row in `segs_game` account table resolves the issue.

broxen:
Per @nemerle, this will be resolved by the server during account login. I'll leave this open until that patch goes through.

nemerle:
Closed by #322

PR #316, opened 2018-04-11
Fix for Issue 314 by adding row to segs_game dbase when adding user

broxen:
- Iterate dbtool to v0.4
- Add row to segs_game accounts table when adding new character. Closing #314


nemerle:
This is not needed since the coming changes to the way servers communicate will cause us to create GameDB accounts on the fly, and only AuthDB accounts are created by hand

PR #319, opened 2018-04-11
Couple small fixes to m_account_level and Windows serialization

broxen:
- Fix m_access_level by pulling value from `accountInfo()`.
- Initialize window settings to 0 (hopefully this fixes random crashing when server is running debug mode)

### DIFF
--- a/Projects/CoX/Common/GameData/gui_definitions.h
+++ b/Projects/CoX/Common/GameData/gui_definitions.h
@@ -76,16 +76,16 @@ public:
     GUIWindow() { }
 
         // GUI Window Params
-        WindowIDX           m_idx;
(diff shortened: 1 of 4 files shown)


broxen:
Looks like this resolved the issue for psiLupan.

I know you're working on a bunch of other changes, so no rush here. Just letting you know it's ready for merge.

PsiLupan:
Yep, can confirm this did it.

Basically, the issue seems that the client asked for the server to provide menu locations with a new account, and the server was sending unitialized garbage back. Then the client would respond with unitialized garbage to store, which would occasionally overflow and break the pak reading to throw a fake event 2.

ldilley:
Nice work with the window parms, @broxen.

PR #322, opened 2018-04-12
Introduce unified client session store

nemerle:
Every server's handler now has a user session store that associates an
    unique id, generated after connection to AuthServer and server specific
    user session object. The unique id is the same for all sessions having their
    source in specific Auth connection. This is done by passing the session id to each of downstream
    servers during processing of ExpectClientRequest/Response logic.
    
    In addition, started to separate the DB objects vs Live objects, for the
    ones that live in DB, and are loaded from it.
    
    The main goal here is to provide 'dummy' objects as a data exchange
    layer between internal DB servers, and user-facing Auth/Game/Map ones.


ldilley:
To summarize the Discord conversation for record-keeping purposes here, the following compilation problems were encountered using this PR (#322) under CentOS 7.4 with `gcc 6.3.1`:

* Removal of `-std=c++11` from `CMAKE_CXX_FLAGS` in the top-level CMakeLists.txt caused unit tests to be performed prior to building the 3rd-party Cereal dependency despite `ENABLE_TESTS` being set to `OFF`. I would not have noticed this if I had 32-bit compat lib (multilib or glibc-devel.i686) support installed. This failure occurred during the Cereal portability unit test. Re-adding `-std=c++11` to the top-level CMakeLists.txt resolved this issue.

* `float get_interpolator_perturbation(int16_t a1,int level)` in `Segs/Projects/CoX/Common/NetStructures/StateInterpolator.cpp` was also calling `std::abs()` using an `int16_t` value which was being cast to a float when using GNU 6.3.1's instance of math.h. The fix was to wrap the call in `int()` like so: `s_coding_sequence[int(std::abs(a1))]*(1<<level);`

* `Segs/Projects/CoX/Servers/GameDatabase/GameDBSyncContext.cpp` included `<QStringLiteral>` which does not seem to exist. Including `<QString>` satisfies the literal requirement.

Making the above changes allows for a successful build.

ldilley:
1a2ff3f and ce856d3 have resolved the compilation errors detailed in my previous comment. Executing `authserver` results in a segfault most of the time. Various threads also seem to execute in no particular order which sometimes seems to trigger a SIGABRT. Some relevant output is attached:

[gdb_output.txt](https://github.com/Segs/Segs/files/1913385/gdb_output.txt)
[run_attempt_example_1.txt](https://github.com/Segs/Segs/files/1913386/run_attempt_example_1.txt)
[run_attempt_example_2.txt](https://github.com/Segs/Segs/files/1913387/run_attempt_example_2.txt)
[valgrind_verbose_full_leak_check.txt](https://github.com/Segs/Segs/files/1913388/valgrind_verbose_full_leak_check.txt)

PR #326, opened 2018-04-14
[RFC] Email System

PsiLupan:
**Additions:**
* Adds the EmailHeaders and EmailRead game commands
* Adds the EmailHeaders, EmailMessageStatus, and EmailRead events
* Support the sending of multiple headers or a single email header

Current functionality is just sending a test email header and email read with the same ID.

**TODO:**
* Add emails to the database with EmailSend (This will likely require some input sanitizing)
    Ex. /emailsend "\q[Name]\q " "Subject" "Content"
* Send EmailMessageStatus event on Success/Failure of Send, rather than using a InfoMessageCmd
* Add timestamp generation as the client does not send this information - This is the number of seconds elapsed since Y2K relative to the server's defined timezone (Currently 0, -12 GMT).
* Remove emails from database with EmailDelete (Will need to check that the entity ID is the owner, as EmailDelete only sends "emaildelete [id]")

Email Image

### DIFF
--- a/Projects/CoX/Servers/MapServer/Events/EmailHeaders.h
+++ b/Projects/CoX/Servers/MapServer/Events/EmailHeaders.h
@@ -15,7 +15,7 @@ class EmailHeaders final : public GameCommand
 {
 public:
     struct EmailHeader {
-        int valid = true;
+        bool valid; //Checked if the email exists, can probably use this to mark emails that are awaiting garbage collection
         int id;
         QString sender;
         QString subject;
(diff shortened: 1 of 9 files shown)


Projects/CoX/Servers/MapServer/Events/EmailHeaders.h
```diff
@@ -0,0 +1,77 @@
...
+                bs.StoreBits(32, m_emails.at(i).timestamp);
+            }
+        }else{
+            bs.StoreBits(1, m_valid);
+            bs.StoreBits(32, m_id);
+            bs.StoreString(m_sender);
+            bs.StoreString(m_subject);
+            bs.StoreBits(32, m_timestamp);
+        }
+    }
```
nemerle: The special code for the single-mail-header could be simplified by doing this:
```cpp
    /*Defines a single email header to send*/
    EmailHeaders(const int &id, const QString &sender,  const QString &subject, const int &timestamp) : GameCommand(MapEventTypes::evEmailHeadersCmd),
        m_fullupdate(false),
        m_count(1),
        m_timestamp(timestamp)
    {
        m_emails.emplace_back( EmailHeader { true,id,sender,subject,timestamp} );
    }
```
and this makes serializeto easier to write:
```cpp
    void    serializeto(BitStream &bs) const override {
        bs.StorePackedBits(1, type()-MapEventTypes::evFirstServerToClient);
        bs.StorePackedBits(1, m_fullupdate);
        bs.StorePackedBits(1, m_emails.size());
        for(const EmailHeader & hdr : m_email) {
                bs.StoreBits(1, hdr.valid);
                bs.StoreBits(32, hdr.id);
                bs.StoreString(hdr.sender);
                bs.StoreString(hdr.subject);
                bs.StoreBits(32, hdr.timestamp);
          }
    }
```

ldilley: Could the type of `valid`/`m_valid` be set to `bool` or will these vars be expected to accept more than two possible values (other message states that exist perhaps) in this context? We should be leveraging `bool` as a best practice when appropriate considering it is designed for exactly two values, is typically 1 byte versus 4 (int) on modern implementations, and promotes self documentation.

PsiLupan:
Addressed the reviews.

Also, removed the m_valid flag and just set it to always be true. Ideally, we just wouldn't send anything marked for deletion if we weren't updating rows on every delete. This flag doesn't serve much of a purpose otherwise, unless it was changed in later issues.

Issue #327, opened 2018-04-15
Items Required for Proper Email System


PsiLupan:
*  Add emails to the database upon EmailSend (This will likely require some input sanitizing)
Ex. /emailsend "\q[Name]\q " "Subject" "Content"

* Generate a timestamp on the sent email as the client does not send this information

 This is the number of seconds elapsed since Y2K relative to the server's defined timezone (Currently 0, -12 GMT), which can be done using 

This can be done using http://doc.qt.io/qt-5/qdatetime.html#secsTo like so:

`QDateTime y2k(QDate(2000, 1, 1), QTime(0, 0));`

`int timestamp = (int)y2k.secsTo(QDateTime::currentDateTime());`

* Use EmailMessageStatus event on Success/Failure of Send

This is used if a recipient does/doesn't exist. So, you would need to lookup the recipient name in the database.

* Remove emails from database with EmailDelete (Will need to check that the entity ID is the owner, as EmailDelete only sends "emaildelete [id]")

broxen:
Per conversation on discord April 25th, 4:30pm PST between psiLupan and nemerle, and featuring others.

### Proposed `emails` in `segs_game`
Since all that is really stored is per-user "Inbox", we should simply add Serialized email data to PlayerData struct. That serialized EmailData should be a struct containing sender_id, recipient_id, subject, message body, and any other relevant data passed by the email system.

### Tasks for Completion
- [ ] Create EmailData struct
- [ ] Add necessary fields for email system to EmailData
- [ ] Leverage new EmailData struct throughout existing EmailMessageStatus
- [ ] Add EmailData to PlayerData struct
- [ ] Add Archive functions to PlayerData to properly serialize EmailData. Look to GUISettings for an example.

@PsiLupan or @nemerle feel free to chime in if I forgot anything here.

broxen:
#367 relates to this as well

broxen:
This portion of this was completed in v6

PR #330, opened 2018-04-16
EmailMessageStatus Fix & Additions

PsiLupan:
Addresses part of #327

**Fix**: 
* Corrected EmailMessageStatus serialization

**Changes**:
* Removed debug message from EmailSend, since EmailMessageStatus will open a dialog window on failure in the future.
* Removed some instances of m_client checks, since presumably these functions can't be called until actually in-game
* Added comments showing what storeEmailInDb should do in the future partially, such as the timestamp generation and usage of EmailMessageStatus.


Projects/CoX/Servers/MapServer/Events/EmailMessageStatus.h
```diff
@@ -6,21 +6,22 @@
 class EmailMessageStatus final : public GameCommand
 {
 public:
-    EmailMessageStatus(const int status, const QString &recipient) : GameCommand(MapEventTypes::evEmailMsgStatus),
+    EmailMessageStatus(const bool status, const QString &recipient) : GameCommand(MapEventTypes::evEmailMsgStatus),
```
nemerle: IMHO const passed-by value arguments only make sense when we have a large function and want to ensure that incoming argument is not getting over-written by mistake in function's body. Here that will not be a case, since we just use the status value to initialize class member.

broxen:
Hello @PsiLupan, when you get a chance, can you please resolve the conflicts here, and address Nemerle's comment regarding the const values. Pending those changes, I don't see any reason not to merge this.

broxen:
After speaking with psiLupan, I'm going to close this PR and he'll reopen later after some changes.

PR #350, opened 2018-04-18
Fly on jump fix

### LINKED ISSUE #241: Make client side movement more bearable : set the fly bit on jump
When we jump now, the server simply increments the current Y coordinate of the player.

Since the server-side position never 'falls' with gravity, and client firmly believes that what comes up must go down, we could at least prevent 'forever falling' thing, by setting a proper bit in the entity, thus marking it as a flying one.

**Required skills**: c++
**Required knowledge**: ability to find the proper bit in Entity and set it when first 'jump' event arrives for that entity.

Thanks go to: @mobbyg for inspiration 💡

miles200:
Closes #241  .

- Added a function to set m_is_flying within Entity to true on jump event
- 
-

### DIFF
--- a/Projects/CoX/Servers/MapServer/WorldSimulation.cpp
+++ b/Projects/CoX/Servers/MapServer/WorldSimulation.cpp
@@ -6,6 +6,12 @@
 #include "Events/GameCommandList.h"
 #include <glm/gtx/vector_query.hpp>
 
+void markFlying(Entity &e ,bool is_flying) // Function to set character as flying
+{
+
+    e.m_is_flying = is_flying;
+
+}
 void World::update(const ACE_Time_Value &tick_timer)
 {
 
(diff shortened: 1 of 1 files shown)


Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -3,9 +3,16 @@
 #include "MapClient.h"
 #include "AdminServer.h"
 #include "Database.h"
+#include "Events/InputState.cpp"
```
nemerle: Ha, in c++ we include `.h` files my friend
Unless we do something funky :wink:

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -37,8 +43,17 @@ void World::physicsStep(Entity *e,uint32_t msec)
         float vel_scale = e->inp_state.input_vel_scale/255.0f;
         e->m_entity_data.m_pos += ((za*e->inp_state.pos_delta)*float(msec))/50.0f;
         e->vel = za*e->inp_state.pos_delta;
+
     }
+
+    if(e->inp_state.pos_delta[1] == float(1.0f)) // Will set 'is flying' on jump event
+
```
nemerle: Hmm.. I'm seeing double new-lines here as well as in markFlying ? Single new line is ok here:
```cpp
{
  something();
}```

miles200: Thanks fixed

Issue #351, opened 2018-04-18
Coverity defect cleanup


ldilley:
Review and handle Coverity defects however applicable. For unreachable code where vars are explicitly set to literals such as 0, false, etc., just leave a FixMe comment as a placeholder for future implementation.

PR #353, opened 2018-04-19
Batch of Coverity fixes

ldilley:
Some #351 effort. The corresponding findings have also been documented within Coverity.

- Various pointer initialization
- FixMe breadcrumbs for future implementation
- Auth session atomicity
- Null checks
- Cereal exception handling


Projects/CoX/Servers/AuthServer/AuthHandler.cpp
```diff
@@ -167,7 +167,7 @@ bool AuthHandler::isClientConnectedAnywhere(uint32_t client_id)
 void AuthHandler::on_retrieve_account_response(RetrieveAccountResponse *msg)
 {
     uint64_t sess_token = msg->session_token();
-    AuthSession *sess_ptr = nullptr;
+    std::atomic<AuthSession *> sess_ptr = {nullptr};
```
nemerle: Why atomic here ? AuthHandler, and basically all *Handler classes aim to be serviced by single-thread-of-execution-per-handler ?

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -42,15 +42,19 @@ void    setTeamID(Entity &e, uint8_t team_id)
...
     }
     else
         e.m_has_team            = true;
 
-    qDebug().noquote() << "Team Info:"
-             << "\n  Has Team:" << e.m_has_team
-             << "\n  ID:" << e.m_team->m_team_idx
-             << "\n  Size:" << e.m_team->m_team_members.size()
-             << "\n  Members:" << e.m_team->m_team_members.data();
+    if(e.m_team)
```
nemerle: Can I haz indent reduction by :
```cpp
if(nullptr==e.m_team)
    return;
```
?

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -810,8 +811,11 @@ void MapInstance::process_chat(MapClientSession *sender,QString &msg_text)
     MessageChannel kind = getKindOfChatMessage(cmd_str);
     std::vector<MapClientSession *> recipients;
 
+    if(!sender)
+      return;
+
     if(sender && sender->m_ent)
```
nemerle: since validity of sender was checked above, and function will `return` in case of invalid sender we don't need to check it again with `if(sender &&... `

Projects/CoX/Servers/AuthServer/AuthHandler.cpp
ldilley: Per Coverity ID #273590:
`Value not atomically updated
The result of the update will be determined by the interleaving of thread execution.
Using an unreliable value of sess_ptr inside the second locked section. If the data that sess_ptr depends on was changed by another thread, this use might be incorrect.`

It is detecting potential dirty writes between `m_sessions.reap_lock().unlock()` and `SessionStore::MTGuard guard(m_sessions.reap_lock());` if >1 thread will be touching session data it can point to outside a mutex. If only a single thread is guaranteed to touch this data, then it will never be a problem and we can revert to a non-atomic object.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
ldilley: Certainly.

ldilley:
The requested changes have been implemented and all conflicts created by merging #355 have been resolved. Linux and Win builds are passing. macOS is not failing due to our code, but rather a network issue with a Brew repo (see below.) The changes contained within this PR are also rather mild -- especially considering we backed out atomizing Auth session data. Given the aforementioned information, I am proceeding with the merge before more conflicts arise as a result of merging additional PRs.

`https://homebrew.bintray.com/bottles/qt-5.10.1.sierra.bottle.tar.gz
curl: (56) SSLRead() return error -9806
Error: Failed to download resource "qt"`

Issue #285, opened 2018-04-05
Create a graphical admin tool for server setup and management.


broxen:
Create a graphical admin tool for server setup and management. (building off #33)

We can leverage QT, since it's already a dependency.

Should be able to:
- [x] Create a default maps folder structure
- [x] Define path to CoX client, and copy/extract PIGG and BIN files to server directory (see piggtool)
- [x] Clean up unused PIGG files
- [x] Display message if there is a newer server **release** available on GitHub.
- [x] Run/shutdown server
- [x] Set setting.cfg options using text fields and checkboxes
- [x] Create new player accounts (new functionality)
- [x]   Add a simple settings.cfg generator, which from local IP builds the proper file.
- [x] Reset databases to default (see: `./dbtool create -f`)

Future features (very low priority) that may require additional infrastructure:
- Provide means for updating server releases
- Show server stats (version, active connections, uptime, etc)
- Update database schemas after changes (new functionality)
- Ban player accounts (new functionality)
- Reset player passwords (how to handle this?)
- Delete Accounts and associated characters
- Rename characters (TheIncredibleHulk needs a new name but is offline)
- Ban based upon IP (for the jerk hackers)
- Maybe a simple MOTD editor to edit scripts/motd.smlx

nemerle:
Maybe not a duplicate, but an extension-of :)

broxen:
Does this overlap with SLAV? Maybe I should take a look at that.

miles200:
I can start looking at the GUI part of this

miles200:
Must this be in C++ or would a Python solution suffice?

ldilley:
Good question, @miles200. This would preferably be a portable language, so Python meets that criteria. The resultant program must also possess all features outlined by broxen (and possibly others in the future.) Python can also leverage the Qt bindings provided by either [PyQt](https://riverbankcomputing.com/software/pyqt/intro) or [PySide](https://wiki.qt.io/PySide) as necessary (for GUI elements and such.)

nemerle:
We want the tool to be distributable, python + Qt is good for the prototype, later on we will have to convert the tool to c++

miles200:
We can probably use PyInstaller or PyEXE tools to distribute but never used before. I'll get on with a prototype for now, cheers all

miles200:
I notice there's a ServerMonitor in the Segs/Utilities directory, could I make use of this for the start and stop auth server?

miles200:
[...]

PR #368, opened 2018-04-22
Initial version of GUI tool for Admin functions

miles200:
Contributes to #285 

Initial version of GUI tool. Added some documentation below, not sure the best place to keep this for reference to please advise. 

## Implemented in this PR:

- Create segs and segs_game databases. Will detect existing and prompt user before overwriting.
- Add users to segs database

**Current known issues:**

- Application currently uses command / terminal for feedback/output until I can work out how to feed it back into the GUI and handle errors (like error on adding user with existing user/password)... Putting as issue but may not be, maybe okay as is
- No support for macOS/OS X... yet
- Windows exe is 18mb and Linux is around 42mb!... But may just be the external libraries used... again not sure if really an issue

## To build:

**Pre-Reqs:**

- Python 3
- PyInstaller 

> pip install pyinstaller

- PyQt5

> pip3 install pyqt5

**Build:**

Run the following from the same directory that main.py exists in

> pyinstaller -F -n SEGSAdmin main.py

An executable is created in a sub directory 'dist'. Move this into the root of the 'out' directory of your CoX build. The 'build' sub directory can be deleted after build is complete.

**Windows users:** You need libwinpthread-1.dll copied into the 'out' directory to work, this should be copied in on SEGS build but doesn't seem to be happening. This can be found within:

> Qt > 5.x.x. > mingw53_32 > bin

Windows - Run as normal .exe

Linux - Run from terminal using ./SEGSAdmin


miles200:
Hold this for now, decided on discussion to convert into C++, will close and open a new PR once done

Issue #289, opened 2018-04-07
Getting Qt5Gui.dll is missing error when running SEGS for the first time


Armingleggings:
I installed the CoX client and set the server IP to 127.0.0.1. I downloaded SEGS and extracted it. I placed the folder in Program Files (x86). When I run SEGS.exe, I get the above listed error. I can download the DLL, but I suspect something else is wrong. Should it have been included? Was there an assumption made about the user environment that isn't true in my case (a missing library or perhaps something like and SDK)? 

I'm running Windows 7 in my Virtual Machine.

Armingleggings:
My mistake. Is there a guide or walkthrough or setup steps page I missed? If not, we might want to have something to make getting started as easy as possible. 

That said, what if I set it to my local ip address? It would have the same effect as localhost without triggering some of the custom stuff ya?

Armingleggings:
Running authserver gives a series of errors about not finding tables and directories. For example: "no such table: costume" and "couldn't load costume.bin from /data/bin". What am I missing?

Armingleggings:
Following instructions. Had to run dbtool -f because the defaults didnt' seem to be working for some reason. That got rid of the SQL errors, but I'm still getting a few more like "config file is missing location_addr" and such. Going back to see if I missed anything else.

Armingleggings:
Yup, I'm stuck now. It's saying the local_addr value isn't in the config, but it is (unless settings.cfg is a decoy). Also warning me about the pigg files, but I've run the pigtool a few time and it executes just fine.

broxen:
Please ensure that the pigg files are located in the segs folder under `/data/` and likewise that the extracted bin files are located under `/data/bin/`

Armingleggings:
Odd... the pigg files are all in place, but there's no data/bin folder... let me look at those instructions for the piggtool again.

Armingleggings:
OH! It creates bin in the root SEGS folder, but it needs to be then moved to data. Got it.

Out of curiosity, couldn't the piggtool just export to data/bin directly instead of requiring the extra step?

Armingleggings:
I don't see a bin folder anywhere. It says it should be in the SEGS output folder, but I downloaded the stable release (didn't compile). I see and "out" file, but no folder and bin is nowhere in the folder structure according to Windows search.

Armingleggings:
[...]

Issue #272, opened 2018-03-12
Documentation needs to be updated to reflect changes in the latest codebase.


broxen:
The READMEs need to be updated and reviewed for conformity with the latest changes to the server, settings.cfg, and dbtool. 

Also, this is a good opportunity to interface with the community and take feedback for streamlining/improving the docs for clarity and simplified use, especially for those without substantial coding experience.

There are several READMEs that need to be updated (these are complete in #418):
- [x] [./README.md](https://github.com/Segs/Segs/blob/7450f2abffc9effaf0cb7748d0568e746d1cc214/README.md) - the root directory readme and first thing users see when viewing our GitHub. This should be geared towards a developer audience looking to compile, BUT needs to provide quick links and clear instructions on where to find our installation and setup guides (and readme)
- [x] [./Projects/CoX/docs/README.md](https://github.com/Segs/Segs/blob/7450f2abffc9effaf0cb7748d0568e746d1cc214/Projects/CoX/docs/README.md) - This readme will appear in the SEGS program (out) directory. It should be geared towards setting up and running the server. It should be clear and concise, and focus on the most common methods for install (don't need to cover every possible configuration, just the most common ones)
- [x] [./Projects/CoX/Utilities/dbtool/README.md](https://github.com/Segs/Segs/blob/0901a04dd4c05d8d92d78ef43265d4d90da45b51/Projects/CoX/Utilities/dbtool/README.md) - an argument might be made for deleting this, but I think it may be valuable to have. At bare minimum we should leave it, and include a link inside to the setup readme in the program directory.
- [x] [./Projects/CoX/Utilities/piggtool/README.md](https://github.com/Segs/Segs/blob/b68162ac97d3b41aefc8f8dc0ba44484f3798682/Projects/CoX/Utilities/piggtool/README.md) - an argument might be made for deleting this as well, but I think it may be valuable to have, if it's updated with real instructions. At bare minimum we should leave it, and include a link inside to the setup readme in the program directory.

broxen:
As part of this, we should update links to the new segs.io url.

Homepage: https://www.segs.io/ (Under construction! construction)
Forums: https://forum.segs.io/
Twitter: https://www.twitter.com/segscode
Wiki: https://github.com/Segs/Segs/wiki
Discord Server: https://discord.segs.io/
IRC: <irc://irc.segs.io/segs>

ldilley:
[...]

PR #369, opened 2018-04-23
Common scenegraph

nemerle:
Map scene graph is loaded on instance's startup

*fix bug in bitstream
**Note** this requires more unpacked pigg files:
* all geom piggs
* all maps piggs


Components/BitStream.cpp
```diff
@@ -365,8 +365,8 @@ int64_t BitStream::Get64Bits()
 
 size_t BitStream::GetAvailSize() const
 {
-    int64_t res = (int64_t)((m_size-m_write_off)-(m_write_bit_off!=0));
-    return std::max<size_t>(0,res);
+    int64_t res = int64_t(m_size)- int64_t(m_write_off)-(m_write_bit_off!=0);
```
broxen: Is `int64_t(var)` typecasting preferable to `(int64_t)var`, or do they work differently?

broxen:
We should resolve conflicts here, but also provide direction on the new piggs that need to be unpacked.

Does binConverter work on these files already? Is this something that could be added to #289? We should mention it there if so. Otherwise we'll need to add these additional instructions to the READMEs #272.

Issue #352, opened 2018-04-19
Clean up compiler warnings


ldilley:
Correct as many compiler warnings thrown by -pedantic, -Wall, etc. without breaking things.

ldilley:
Absolutely.

broxen:
I think this is pending review of #354

ldilley:
Clang and GCC builds are significantly cleaner after the merge of #421. SEGSAdmin and SLAV are still plagued with cast warnings until the Qt developers address the reported issues up to 5.9. Everything else has disappeared in gcc 6.3.1. gcc 7 and 8 will likely catch other warnings that still need to be addressed. Lastly, Appveyor and Travis build output should be reviewed and any of the warnings not related to 3rd-party software should be mitigated.

Otlichno:
Warnings I saw are below, I apologize if they are already known:
***************************************************************************
[ 25%] Building CXX object Projects/CoX/Common/NetStructures/CMakeFiles/network_structures.dir/Character.cpp.o
/home/crounce/project/Segs/Projects/CoX/Common/NetStructures/Character.cpp: In member function ‘void Character::finalizeLevel()’:
/home/crounce/project/Segs/Projects/CoX/Common/NetStructures/Character.cpp:107:40: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     if(m_char_data.m_experience_points < data.expForLevel(m_char_data.m_combat_level))
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/crounce/project/Segs/Projects/CoX/Common/NetStructures/Character.cpp: In member function ‘void Character::sendOwnedPowers(BitStream&) const’:
/home/crounce/project/Segs/Projects/CoX/Common/NetStructures/Character.cpp:253:30: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             for(int i = 0; i < power.m_total_eh_slots; ++i)
                            ~~^~~~~~~~~~~~~~~~~~~~~~~~
******************************************************************************
[ 26%] Building CXX object Projects/CoX/Common/NetStructures/CMakeFiles/network_structures.dir/Powers.cpp.o
/home/crounce/project/Segs/Projects/CoX/Common/NetStructures/Powers.cpp: In function ‘CharacterPowerSet getPowerSetData(PowerPool_Info&)’:
[...]

PR #354, opened 2018-04-19
Issue #293 (Serializing m_current_attribs), #304 (Replacing slashCommand string parameters with const), #352 (Handling Compiler Warnings)

eos428:
Issue #293

- Character.m_current_attribs removed as m_current_attribs is now part of CharacterData instead. Now, we access it by Character.m_char_data.m_current_attribs. 
- This also makes m_current_attribs serialized with CharacterData instead of Character.

Suggestion: Make m_max_attribs part of CharacterData too?

Issue #304
- Simply turned a lot of those functions taking QString& parameter to const QString& as those strings are not edited in said functions

Issue #352
- Removed a lot of compiler warnings.
- Bitfields.inl were changed because QtCreator threw plenty of useless cast warnings, later reverted as we shouldn't modify 3rd party libraries.
- Wcompare (Majority of them comparisons between signed and unsigned values) rectified, usually by making the signed variable (usually integers) unsigned to either size_t or uint32_t, since they are often compared to array sizes, etc.
- Wunused warnings are removed by making the unused parameters commented out atm (Like Entity&/*src*/)
- Removed anonymous struct (Wpedantic IIRC) warnings by giving names to structs in Colors.h    

-----------------------------------------------------------------------
First commit message (when it was just Issue #293, and since then a lot has changed even just regarding this issue):

- Added serializeToDb and serializeFromDb functions in attrib_serializers.cpp and .h
- In Struct GameAccountResponseCharacterData, added a serialized version of m_current_attribs
- In Character.cpp, m_current_attribs are now serialized to and from db

Very early on and untested, would expect a lot of edits.

PR #373, opened 2018-04-26
Compiler warning fixes to address #352

ldilley:
This PR contains fixes to quell all compiler warnings (at least in recent versions of GCC and Clang) with the exception of casts related to SLAV for primarily function pointers. `Q_UNUSED` is employed where possible (all except a single case where `<QtGlobal>` was not in include path) considering it is Qt based and self documenting/easy to search and address in the future. If we know certain vars will always be positive during comparison operations, then we can use uints in each context opposed to static casting if desired.


ldilley:
Yielding to PR #354 due to some overlap. Can re-pull once #354 is merged and address any remaining warnings at that time.

PR #375, opened 2018-04-27
Fixed spawn locations and NPC debug messages

### LINKED ISSUE #337: Adjust starting spawn location to Outbreak, since Outbreak is new starting zone.
### Issue Summary
New changes to MapInstance in latest build spawns new characters in Outbreak, but current spawn coordinates are set to Atlas Park spawn location. Should be reconfigured for Outbreak map.

As the current spawn location puts you off map, you immediately fall into the void, thus I'm tagging this a bug.

### Proposed Fix(es)
- Reset spawn coordinates to outbreak starting location
- OR make starting map and spawn location configurable in settings.cfg
- OR pull starting spawn locations from map geo data `_playerspawn` (spawners shown in this video: https://www.youtube.com/watch?v=uWyyFQDG9b4 at 0:00:07)

In my opinion, default spawn logic should be:

1. Spawn at locations set in map geo data
2. Unless overridden in ~settings.cfg~ map specific lua files: `MapInstances/City_00_01/settings.lua`

### Segs Build
3bfadac

broxen:
Closes #337

- Fixed spawn location and /stuck location based upon map
- Fixed some output for npc debug messages

### Notes
Might not be the cleanest way to achieve spawn locations. Feedback is welcome.


Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -97,6 +104,16 @@ MapInstance::MapInstance(const QString &mapdir_path, const ListenAndLocationAddr
     m_scripting_interface.reset(new ScriptingEngine);
     m_endpoint = new MapLinkEndpoint(m_addresses.m_listen_addr); //,this
     m_endpoint->set_downstream(this);
+
+    QString mapname = m_data_path.remove("MapInstances/");
```
nemerle: This `MapInstances` is a configurable part of the path though

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -97,6 +104,16 @@ MapInstance::MapInstance(const QString &mapdir_path, const ListenAndLocationAddr
...
+    qDebug() << mapname;
+    auto iter = std::find_if( g_starting_spawns.begin(), g_starting_spawns.end(),
```
nemerle: Since we're going to need a spawn-location function, why not start at this point:
```cpp
glm::vec3 getSpawnLocation(const QString &fromMap,const QString &into_map)
{
// move spawn location code from constructor here
}
```

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -568,7 +584,7 @@ void MapInstance::on_create_map_entity(NewEntity *ev)
         const MapServerData &data(g_GlobalMapServer->runtimeData());
         const Parse_AllKeyProfiles &default_profiles(data.m_keybind_profiles);
 
-        fillEntityFromNewCharData(*e, ev->m_character_data, data.getPacker(),data.m_keybind_profiles);
+        fillEntityFromNewCharData(*e, ev->m_character_data, map_session.m_current_map->m_spawn_pos, data.getPacker(),data.m_keybind_profiles);
         e->m_char->m_account_id = map_session.auth_id();
```
nemerle: And here we'd do 
```cpp
fillEntityFromNewCharData(...)
e->m_pos = getSpawnLocation("","City_00_01");
```

Projects/CoX/Servers/MapServer/MapInstance.cpp
broxen: Ahh yes, I forgot.

I'll search for the `City_` string instead? I think they all start with that (for now anyway)

Projects/CoX/Servers/MapServer/MapInstance.cpp
broxen: Makes way more sense, and I'm not sure why I didn't think of it

Projects/CoX/Servers/MapServer/MapInstance.cpp
nemerle: Sadly `City_` is not the only one prefix, we also have `Hazard_` :P

Projects/CoX/Servers/MapServer/MapInstance.cpp
broxen: Wondered if there was some reason why I shouldn't pop this out of `fillEntityFromNewCharData()`, so I left it in there.

nemerle: Well, I think in the end we'll have to use the current MapScenGraph, active Entities, and source map as the parameters to properly locate the correct spawn point.

Projects/CoX/Servers/MapServer/MapInstance.cpp
broxen: What is the thought behind `from_map` and `into_map`? Are you thinking of map transfers?

I figured the logic would be completely different for those. Or am I overlooking something?

PR #381, opened 2018-04-28
Fixed "character select list" query.

dracc:
Characters are created/stored with 'account_id' value of
chardb->accounts->account_id, whereas they are/were looked up with the
chardb->accounts->id value. Those two values are not necessarily the same.

Update: So, a lot of stuff happened.
- Progress table was removed as it fits in nicely inside the characters table as a blob.
- #393 was taken care of as the scripts are already modified anyways.
- MySQL creation scripts were fixed, as they would complain/error out if the database was already populated due to foreign key constraints.
- CharDB->accounts table lost its 'id' entry as it's not used. account_id from the same table has since been renamed 'id'.
- Supergroups now only have one id column.

### DIFF
--- a/Projects/CoX/Servers/GameDatabase/GameDBSyncContext.cpp
+++ b/Projects/CoX/Servers/GameDatabase/GameDBSyncContext.cpp
@@ -150,8 +150,8 @@ bool GameDbSyncContext::loadAndConfigure()
               "WHERE id=:id ");
 
     prepQuery(*m_prepared_fill,"SELECT * FROM costume WHERE character_id=? AND costume_index=?");
-    prepQuery(*m_prepared_account_select,"SELECT * FROM accounts WHERE account_id=?");
-    prepQuery(*m_prepared_account_insert,"INSERT INTO accounts  (account_id,max_slots) VALUES (?,?)");
+    prepQuery(*m_prepared_account_select,"SELECT * FROM accounts WHERE id=?");
+    prepQuery(*m_prepared_account_insert,"INSERT INTO accounts  (id,max_slots) VALUES (?,?)");
     prepQuery(*m_prepared_char_insert,
                 "INSERT INTO characters  ("
                 "slot_index, account_id, char_name, chardata, entitydata, "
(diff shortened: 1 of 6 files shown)


Projects/CoX/Servers/GameDatabase/GameDBSyncContext.cpp
```diff
@@ -219,7 +219,7 @@ bool GameDbSyncContext::getAccount(const GameAccountRequestData &data,GameAccoun
         if(!m_prepared_account_select->next() && !data.create_if_does_not_exist)
             return false;
     }
-    result.m_game_server_acc_id = m_prepared_account_select->value("id").toULongLong();
+    result.m_game_server_acc_id = data.m_auth_account_id;
     result.m_max_slots = m_prepared_account_select->value("max_slots").toULongLong();
```
nemerle: This does not look 'right' though, `m_game_server_acc_id` is GameDb account id, and m_auth_account_id is AuthDb account id

dracc: Are those not meant to be the same though? Because the AuthDb account_id is used against character_db account_id, and the value of account_id is what is stored in chardb->characters->account_id, which is the value used to select which characters belongs to which account.

The other possibility I see is that character creation is changed so that character->account_id uses the value of chardb->accounts->id. That would kind-of make sense, but then chardb->accounts->account_id should probably be named auth_id to better reflect what it is used for.

dracc: For future reference, as I understood, we agreed that using authdb->accounts->id and chardb->accounts->account_id interchangeably ought to be okay on discord.
And while it may not be best practice, in the given case, authdb->accounts->id is supplied through function argument and might as well be used instead of fetching the same value from a sql query result.

broxen:
@nemerle  As far as I can tell, and considering your Discord discussion with @dracc , this is ready for merge. I'll wait for your confirmation before pulling the trigger.

nemerle:
This is great work, just a few general notes:
- unique id in all tables that have it should be named `id`
- if the database table itself is bound 1-1 with another, why not store the data as serialized blob ?

dracc:
@nemerle I've now removed the progress table as it ought to fit in nicely inside characters->char_data or characters->player_data.

accounts->account_id is still named as it is as it has very tight binds to the AccDb->account->id column. Maybe should be named auth_id instead?

broxen:
As a part of this, would you be so kind as to remove the columns for Health and Endurance in Characters table per #393?

Greatly appreciated

dracc:
@broxen sure thing, no problem! :)

dracc:
CharDb -> accounts -> account_id have now been renamed `id`. I'll squash the commits and then I believe it's ready for merge?

PR #384, opened 2018-04-29
Added new logging categories, cleaned up logging output, and more

broxen:
- Renamed some variables to match the coding guidelines
- Added a method to StateInterpolator (currently does nothing)
- General Code Cleanups
- Moved some methods around to match declaration order in DataHelpers and SlashCommand


ldilley:
`Projects/CoX/Common/AuthProtocol/AuthLink.h:27:10: fatal error: QtCore/QDebug: No such file or directory` on Linux. This should be just `#include <QDebug>`.

Thank you for also passing the array for testEncVec() by reference for some efficiency. This also resolves Coverity defect ID #272685.

broxen:
Actually I need to just remove that. I added it accidentally

PR #385, opened 2018-04-30
Initial GUI tool for Admin functionality

miles200:
Contributes to  #285  .

- Initial GUI tool
- Create databases, will prompt to overwrite if they exist
- Add users 
- Start and Stop AuthServer
- Read from settings.cfg (no edit / save or ability to change file location... yet)

Code has ended up quite large and I'm sure there's much better ways to do what I've done so code review / feedback very welcome 😀


Projects/CoX/Utilities/SEGSAdmin/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    addnewuser.h
+    dialog.ui
+    dialog.cpp
+    dialog.h
+    resources.qrc
+
+)
+
+add_executable(SEGSAdmin WIN32 ${SEGSADMIN_SOURCES})
+qt5_use_modules(SEGSAdmin Core Gui Widgets Network)
```
nemerle: First overall comment: 
Please name the files after the class names that they implement, use UpperCamelCase for class names, and make names a bit more descriptive (what `dialog` is doing cannot be seen from it's name :) ) `addnewuser` should be properly camel-cased, and if it's a dialog, rename it to AddNewUserDialog

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
```diff
@@ -0,0 +1,402 @@
...
+ * @addtogroup SEGSAdmin Projects/CoX/Utilities/SEGSAdmin
+ * @{
+ */
+
+#include "SEGSAdminTool.h"
+#include "ui_SEGSAdminTool.h"
+#include "addnewuser.h"
+#include "dialog.h"
+#include <QDebug>
+#include <iostream>
```
nemerle: <iostream> should be the 'last' in the list of includes, since we follow 'distance' based order:
```
#include "this_cpps_paired_header.h"

#include "other_same_directory_headers.h"
#include "OtherPartOfProject/headers.h"
#include "ProjectBundled3rd_party_libraries/header.h"

#include <external_library.h>
#include <system_header.h>
```

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
```diff
@@ -0,0 +1,402 @@
...
+           createUser->waitForFinished();
+           ui->createUser->setText("Create User");
+           ui->createUser->setEnabled(true);
+           qApp->processEvents();
+    }
+    else
+    {
+    ui->output->appendPlainText("Failed to start DBTool Add User...");
+    qDebug() <<"Failed to start DBTool Add User...";
+    }
```
nemerle: something happened to identation above: note that qtcreator has auto-indent selection keyboard shortcut ( ctrl+i )

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
```diff
@@ -0,0 +1,402 @@
...
+
+void SEGSAdminTool::read_createuser() // Not currently working with xterm/gnome-terminal
+{
+    QByteArray out_err = createUser->readAllStandardError();
+    QByteArray out_std = createUser->readAllStandardOutput();
+    qDebug().noquote()<<QString(out_err);
+    qDebug().noquote()<<QString(out_std);
+    QString output_err = out_err;
+    QString output_std = out_std;
+    output_err.replace(QString("Press ENTER to continue..."), QString("** FINISHED **"));
```
nemerle: You should not be required to write QString("text") in this case, as a general note, when the string will be in very often used ( function called often with the string as argument ), Qt has QStringLiteral, which reduces the number of QString construction overhead.

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
```diff
@@ -0,0 +1,402 @@
...
+    output_std.replace(QString("Press ENTER to continue..."), QString("** FINISHED **"));
+    ui->output->appendPlainText(output_err);
+    ui->output->appendPlainText(output_std);
+}
+
+void SEGSAdminTool::DoesDBExist()
+{
```
nemerle: we prefer snake_case_names for our functions

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
```diff
@@ -0,0 +1,402 @@
...
+QFileInfo file1("segs");
+QFileInfo file2("segs_game");
+
+if (file1.exists() && file2.exists())
+    {
+    emit db_exists();
```
nemerle: this code will not work if somehow only 1 database exists, maybe it should be checking using `||`

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
```diff
@@ -0,0 +1,402 @@
...
+}
+
+void SEGSAdminTool::CreateDatabases(bool overwrite)
+{
+    ui->runDBTool->setEnabled(false);
+    ui->runDBTool->setText("Please Wait...");
+    ui->output->appendPlainText("Checking for existing databases...");
+    ui->output->appendPlainText("Setting arguments...");
+    qApp->processEvents();
+    qDebug() << "Checking for existing databases...";
```
nemerle: I'm not sure the code that follows actually checks for existing databases though ? 
Could be just a wrong log message :smile:

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
```diff
@@ -0,0 +1,402 @@
...
+        ui->output->appendPlainText("*** AuthServer Stopped ***");
+    }
+}
+
+void SEGSAdminTool::CheckForConfigFile() // Does this on application start
+{
+    // Load settings.cfg if exists
+    QFileInfo settings_cfg("settings.cfg");
+    if (settings_cfg.exists())
+        {
```
nemerle: indent style:
```cpp
    if(something)
    {
    }
    else
    {
    }
```

Projects/CoX/Utilities/SEGSAdmin/dialog.cpp
```diff
@@ -0,0 +1,42 @@
...
+    ui->label_2->setText("Existing databases found, overwrite?");
+    show();
+
+}
+
+void dialog::dboverwrite()
+{
+    emit overwrite(true);
+}
+
```
nemerle: This whole class could be probably replaced by http://doc.qt.io/qt-5/qmessagebox.html

miles200:
Thank you @nemerle ! I'll work through these review points later on. 👍

Projects/CoX/Utilities/SEGSAdmin/dialog.cpp
miles200: thanks @nemerle , this does look a lot neater... I guess the advantage being I can implement a messagebox from within my SEGSAdminTool class rather than having a separate dialog class?

nemerle: Take a look at `http://doc.qt.io/qt-5/qmessagebox.html#question` this is ready made function for asking the user yes/no questions

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
ldilley: The "Does" also implies this should return a bool to me.

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
miles200: Ah :P, think I had the check here before moving it to it's own function DoesDBExist. Will fix

PR #395, opened 2018-05-01
Fix Typos

broxen:
- Entities has two `i`'s in it and the internet KNOWS
- Use `persistant_name` in NpcCreator
- Add comment to `makeReadableName()` to detail what regex does

### DIFF
--- a/Projects/CoX/Servers/MapServer/MapSceneGraph.cpp
+++ b/Projects/CoX/Servers/MapServer/MapSceneGraph.cpp
@@ -119,15 +119,15 @@ struct NpcCreator
     NpcGeneratorStore *generators;
     QSet<QString> m_reported_generators;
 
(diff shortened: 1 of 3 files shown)


Projects/CoX/Servers/MapServer/MapSceneGraph.cpp
```diff
@@ -151,7 +151,7 @@ struct NpcCreator
                 glm::vec3 angles = glm::eulerAngles(valquat);
                 angles.y += glm::pi<float>();
                 valquat = glm::quat(angles);
-                e->m_char->setName(makeReadableName(npc_costume_name));
+                e->m_char->setName(makeReadableName(persistant_name));
```
ldilley: *cough* persistent *cough* --styopa

PR #398, opened 2018-05-01
Add 'null' client patch.

nemerle:
This is useful when trying to debug the patched client crashes ( patched client has many named functions exported, so stack traces in debuggers that can use them, are much more readable )


ldilley:
The I0 null stuff builds fine here, but I've no way to test the MSVC build. Appveyor passes for this however. Regarding the tint changes to MapViewer, no functionality has changed -- it's merely cosmetic to make it easier to read the parameters. Given the aforementioned information and considering there are no merge conflicts or build errors, this can be merged.

PR #404, opened 2018-05-05
SEGSAdmin Piggtool Update

miles200:
- Add further checks for data/<sub-dirs> and maps directories
- Add easy setup for end user to copy pigg files from CoX directory and extract them
- Add tidy up function to remove pigg files once extracted
- Updated application logo

Adds to #285


Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
```diff
@@ -51,16 +52,43 @@ SEGSAdminTool::SEGSAdminTool(QWidget *parent) :
...
 {
     delete ui;
 }
 
+void SEGSAdminTool::check_data_and_dir() // Checks for data sub dirs and maps dir
+{
+    QDir data_dir_bin("data/bin"); // Currently only checking for these 3 sub dirs, check for all files could be overkill
+    QDir data_dir_geobin("data/geobin");
+    QDir data_dir_objectlib("data/object_library");
+    QDir data_maps_00("maps/City_00_01"); // Only checking default 'maps' directory, if user changes this then check doesn't apply
```
broxen: For `maps`: We're already reading (and creating) the settings.cfg, why not just pull the configured value from there?

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
```diff
@@ -0,0 +1,194 @@
...
+#ifdef Q_OS_MACOS // Adding this to satisfy Travis-CI build, not tested
+        program.prepend("./");
+#endif
+        emit callPiggWorker(program);
+        QTimer timer;
+        timer.setSingleShot(true);
+        QEventLoop loop;
+        loop.connect(m_pigg_tool, SIGNAL(finished(int)), &loop, SLOT(quit())); // Couldn't get this to work on new signal/slot syntax
+        loop.connect(&timer, SIGNAL(timeout()), &loop, SLOT(quit()));
+        timer.start(30000); // 30 Second timer, will time out and quit, if for some reason the event loop doesn't recieve a finished signal from pigg_tool_worker
```
broxen: Is 30sec enough to receive a finish signal? Or is this timer waiting for any response (so it knows piggtool is working)

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
```diff
@@ -0,0 +1,194 @@
...
+    {
+        qDebug() <<"Failed to process ";
+    }
+}
+
+void SetUpData::create_default_directory() // Creates default directories, currently map directory but can be used for future requirements
+{
+    ui->label_create_directory->setEnabled(true);
+    QDir path(QDir::currentPath());
+    path.mkpath("maps/City_00_01");
```
broxen: Again, let's pull `maps` value from settings.cfg

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
miles200: To be honest I'm not sure, the extraction always took place way before 30 secs on my machine, but I suppose we need to think about slower machines? Would increasing the timer solve this?

miles200:
@broxen With regards to the version checking, is it a case of generating the MD-5 checksum of the CoX client .exe file and comparing against a hard coded checksum in SEGSAdmin ? 

Is it the same with regards to checking the piggfiles versions? 
I'll be making output available to the user so this may be okay for now.

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
nemerle: Maybe instead of quitting the loop, report the issue with pigg_tool taking to long ?
I think the most user-friendly way we can solve this, 
1. Start the timer
2. If it times out, show dialog "Pigg tool is taking a long time, [wait some more / cancel ]"
3. If user selects 'wait some more', GOTO 1 :smile: 
4. Else `loop.quit();`

miles200: @nemerle nice idea, I'll give that a go

broxen:
@miles200 yes, I think md5 checksum is the way to go. In fact I think piggtool does this already for the pigg files before extraction (or maybe the server does before reading bin files).

I agree that showing the output of piggtool eliminates the need for checking the pigg files, though I would probably still check the client version as I feel it'll be a very common troubleshooting issue.

nemerle:
You can take a look at SLAV for code that does md5 of specific files ( the code there is also downloading the list of files to check from the web, but we don't need that in admin tool )

miles200:
Piggtool output is fixed locally

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
miles200: Agreed, I've made this change locally

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
miles200: Changed made locally

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
miles200: This is fixed locally, now prompts user on timeout (yes = quit loop, no = reset timer and continue waiting)

nemerle: Awesome :+1:

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
```diff
@@ -51,16 +52,44 @@ SEGSAdminTool::SEGSAdminTool(QWidget *parent) :
...
 {
     delete ui;
 }
 
+void SEGSAdminTool::check_data_and_dir() // Checks for data sub dirs and maps dir
+{
+    QDir data_dir_bin("data/bin"); // Currently only checking for these 3 sub dirs, check for all files could be overkill
+    QDir data_dir_geobin("data/geobin");
+    QDir data_dir_objectlib("data/object_library");
+    QDir data_maps_00(ui->map_location->text()+"/City_00_01");
```
nemerle: consider doing something like:
```cpp
QStringList map_names = { "City_00_01","City_01_01","City_23_01"};
bool all_maps_exist = true;
for(const QString &map_name : map_names )
{
  all_maps_exist &= QDir::exists(ui->map_location->text()+"/"+map_name);
  if(!all_maps_exist) 
      break; // at least one map dir is missing, we can finish early
}
```
if(data_dir_bin.exists() && data_dir_geobin.exists() && all_maps_exist )
{
    ui->icon_status_data->setText("<html><head/><body><p></p></body></html>");
}
else ....
```

File: Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp [hunk 5]
```diff
@@ -0,0 +1,265 @@
...
+    ui->icon_cox_directory->hide();
+    ui->icon_extract_pigg->hide();
+    ui->icon_create_directory->hide();
+    show();
+}
+
+void SetUpData::select_piggs_dir()
+{
+    QString pigg_dir = QFileDialog::getExistingDirectory(this, tr("Select CoX Directory"),
+                                                         "/home",
```
nemerle: Maybe use [QStandardPaths](http://doc.qt.io/qt-5/qstandardpaths.html) here instead of hardcoded "/home"

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
```diff
@@ -0,0 +1,265 @@
...
+    }
+}
+
+void SetUpData::copy_piggs_files()
+{
+    ui->piggtool_output->appendPlainText("Copying Files...");
+    ui->progressBar->setEnabled(true);
+    ui->buttonBox->setEnabled(false);
+    QStringList sourceFileList;
+    sourceFileList << "bin.pigg" << "geom.pigg" << "geomBC.pigg" << "geomV1.pigg" << "geomV2.pigg" << "mapsCities1.pigg"
```
nemerle: this could be also done like
```cpp
QStringList sourceFileList = {
    "bin.pigg",
    "geom.pigg",
    .....
};
```

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
miles200: Changed locally

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
miles200: Changed locally

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
miles200: Changed locally

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
```diff
@@ -0,0 +1,265 @@
...
+                   << "mapsCities2.pigg" << "mapsHazards.pigg" << "mapsMisc.pigg" << "mapsMissions.pigg" << "mapsTrials.pigg";
+    QDir sourceDir(ui->pigg_file_url->text().append("/piggs"));
```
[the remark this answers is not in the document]
miles200: Changed locally

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
```diff
@@ -0,0 +1,265 @@
...
+    ui->piggtool_output->appendPlainText(out_err);
+    ui->piggtool_output->appendPlainText(out_std);
+
+}
+
+void SetUpData::create_default_directory(QString maps_dir) // Creates default directories
+{
+    ui->label_create_directory->setEnabled(true);
+    QDir path(QDir::currentPath());
+    path.mkpath(maps_dir+"/City_00_01");
```
nemerle: Well, we have a nice list of maps somewhere before, we might as well make that list a file-scope variable and then:
```cpp
for(const QString &map_name : s_map_names)
{
    path.mkpath(maps_dir+"/"+map_name);
}
```

miles200: @nemerle I like this, but my map lists is in another file

broxen: Stick it in its own header then shared by both files?

miles200: If I create a Globals.cpp / Globals.h then create a static QStringList to access from other classes, is this acceptable? Or would I be better implementing something using Signals / Slots to fetch the list?

Projects/CoX/Utilities/SEGSAdmin/Globals.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+
+// List of all known maps
+QStringList g_map_names = {
+    "City_00_01",
+    "City_01_01",
+    "City_23_01"
+};
+
+Globals::Globals()
```
broxen: We can clean up the empty globals class. We'll never use it in the future (As far as I can tell) and there's no need for the empty class syntactically.

Projects/CoX/Utilities/SEGSAdmin/Globals.h
```diff
@@ -0,0 +1,14 @@
+#ifndef GLOBALS_H
+#define GLOBALS_H
+#include <QStringList>
+
+extern QStringList g_map_names; // List of all known maps
+
+class Globals
```
broxen: Same here. Remove this.

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
```diff
@@ -0,0 +1,279 @@
...
+    show();
+}
+
+void SetUpData::select_piggs_dir()
+{
+    QString homedir = QStandardPaths::locate(QStandardPaths::HomeLocation, QString(), QStandardPaths::LocateDirectory);
+    QString pigg_dir = QFileDialog::getExistingDirectory(this, tr("Select CoX Directory"),
+                                                         homedir,
+                                                         QFileDialog::ShowDirsOnly
+                                                         | QFileDialog::DontResolveSymlinks);
```
broxen: Unless setting this to false would break windows compatibility, is there any reason not to allow symlinks? They are quite common on Linux and maybe used on Max (since its Unix based).

I'd shoot for not changing "expected" filesystem behavior unless we need to for compatibility reasons.

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+    float counter = 0;
+    foreach(QFileInfo fileInfo, piggInfoList)
+    {
+    //QString program = "piggtool -x data/bin.pigg";
+        QString program = "piggtool -x " + fileInfo.absoluteFilePath();
+#ifdef Q_OS_LINUX
+        program.prepend("./");
+#endif
+#ifdef Q_OS_MACOS // Adding this to satisfy Travis-CI build, not tested
```
broxen: Nothing wrong with doing it this way, but if you want to save a line you can use if-then-else here: `#elif`

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+void SetUpData::pigg_dispatcher_wait_dialog()
+{
+    QMessageBox pigg_dispatcher_wait_msgbox;
+    pigg_dispatcher_wait_msgbox.setText("Stop Process?");
+    pigg_dispatcher_wait_msgbox.setInformativeText("This process is taking longer than expected, do you want to wait?");
+    pigg_dispatcher_wait_msgbox.setStandardButtons(QMessageBox::Yes | QMessageBox::No);
+    pigg_dispatcher_wait_msgbox.setDefaultButton(QMessageBox::Yes);
+    int confirm = pigg_dispatcher_wait_msgbox.exec();
+    switch (confirm) {
```
broxen: Please put opening brackets on a new line:
```
switch (confirm) 
{
//...
```

broxen:
Getting this error when creating a new database on Linux with SQLite 
```Starting DBTool Add User...
SQL_ERROR: QSqlError("19", "Unable to fetch row", "UNIQUE constraint failed: accounts.username")```

Not sure if it's dbtool or SEGSAdmin, but the server still loads using the default username.

Also noticed that the console output seems doublespaced in some places? Not sure why. But for readability maybe we can clean up?

File: Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp [hunk 14]
```diff
@@ -0,0 +1,279 @@
...
+    foreach(QFileInfo fileInfo, piggInfoList)
+    {
+    //QString program = "piggtool -x data/bin.pigg";
+        QString program = "piggtool -x " + fileInfo.absoluteFilePath();
+#ifdef Q_OS_LINUX
+        program.prepend("./");
+#endif
+#ifdef Q_OS_MACOS // Adding this to satisfy Travis-CI build, not tested
+        program.prepend("./");
+#endif
```
nemerle: both linux and osx need prepeend so you can also do:
```cpp
#if defined(Q_OS_LINUX) || defined(Q_OS_MACOS)
    program.prepend("./");
#endif
```

mobbyg:
I couldn’t find this pull. Hmm.. 😜

Projects/CoX/Utilities/SEGSAdmin/Globals.cpp
miles200: Fixed locally

Projects/CoX/Utilities/SEGSAdmin/Globals.h
miles200: Fixed locally

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
miles200: It was part of an example I used to construct this, I'll remove it 😄 

Fixed locally

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
miles200: Fixed locally

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
miles200: Fixed locally, have changed as nem suggested to use Q_OS_LINUX || Q_OS_MACOS which tidies up this specific code issue

Projects/CoX/Utilities/SEGSAdmin/SetUpData.cpp
miles200: Fixed locally

Issue #388, opened 2018-04-30
Catch cereal exceptions


nemerle:
Cereal will throw exceptions during serialization of invalid data
We should wrap all cereal functions to gracefully accept those failures, and log the errors.
looking at this https://uscilab.github.io/cereal/assets/doxygen/structcereal_1_1Exception.html
catching cereal::Exception should be enough.

ldilley:
For an example of exception handling, [see the following lines from binConverter.cpp](https://github.com/Segs/Segs/blob/master/Projects/CoX/Utilities/BinConverter/binConverter.cpp#L182-L218). Rather than catch any/all exceptions however, we want to specifically look for `cereal::RapidJSONException` and wrap the necessary statements that might throw such an exception in a try block. We also do not want to exit in all cases, but continue on and log/note the warning as previously mentioned.

nemerle:
As of f9cb381 this is `done`

PR #406, opened 2018-05-10
Issue #388 - Catch Cereal Exceptions

eos428:
Issue #388 - Catch Cereal Exceptions

Every serialize functions in affected files now have try/catch blocks.
1 catch specifically for cereal::RapidJSONException (Gives out qWarning)
1 catch for std::exception (Gives out qCritical)

- Put some unique message in each file's qWarnings and qCriticals?

### DIFF
--- a/Projects/CoX/Common/GameData/anim_serializers.cpp
+++ b/Projects/CoX/Common/GameData/anim_serializers.cpp
@@ -32,10 +32,21 @@ bool loadFrom(BinStore * s, TextureAnim_Data & target)
 template<class Archive>
 void serialize(Archive & archive, TextureAnim_Data & m)
 {
-    archive(cereal::make_nvp("Speed",m.speed));
-    archive(cereal::make_nvp("Scale",m.stScale));
-    archive(cereal::make_nvp("Type",m.scrollType));
-    archive(cereal::make_nvp("flags",m.flags));
+    try
+    {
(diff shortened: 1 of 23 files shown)


nemerle:
This looks great, one comment though, since all those `static void serialize` functions cannot be called from outside, please move try/catch blocks to their callers that 'are' accessible from outside ( the ones that call the 'root' serialize method), this way will reduce the number of try/catch code, and will log the error for the whole serialization attempt, and not for each failed 'step' of serialization

eos428:
Sorry, not quite sure what you mean. Are you talking about things like serializeTo/From in places like Character.cpp and Costumer.cpp?

nemerle:
Exactly, this way you're guarding the gates, and not every step of the way after. Since exceptions are propagated to callers,, 'guarding the gate' will work well enough :)

nemerle:
serialization_common.h  seem like a good central point of doing this, no ? :)

nemerle:
Looking good :)
Should I merge, or do you want to add anything in here ?

eos428:
Ready for merging now if you wish :). Something to note though, while commonSaveTo is called quite often, every serializer seems to have their own version(s) of loadFrom. I guess that may be something to do for the future?

nemerle:
I think all of the loadFrom functions are loading the original bin files, and that code should not throw anything :)

PR #407, opened 2018-05-10
Add player fade in speed to settings.cfg

### LINKED ISSUE #281: Make player fade-in speed configurable.
This would require the `World::effectsStep` to read in a value from a configuration file.
If we wanted to change this depending on the map, then it would have to be in the per-map configuration, otherwise a central server config would suffice.

matthewswar:
Closes #281 

**Additions/modifications proposed in this pull request:**
- Added `player_fade_in` to the [MapServer] group in settings.cfg
- Added struct MapConfig to store `player_fade_in` that gets passed from `MapServer::ReadConfigAndRestart` -> `MapManager::load_templates` -> `MapTemplate` -> `MapInstance` -> `World`
- Default value for `player_fade_in` is `380.0` but will fail reading config if it can't be parsed to a `float`


nemerle:
Welcome to the mad house :smile: 

This PR looks great, but the place you've selected for the variable is not the best one. Might I suggest looking at MapServerData class ? :)

matthewswar:
Thank you for the warm welcome! 😄 I'm definitely interested in contributing to this project in any way I can and am excited to learn more of the codebase.

By the place I've selected, do you mean the `settings.cfg`? I was following the steps laid out in the issue, but I also don't want to pollute the repository with anything unsatisfactory. I'll be without access to a computer for the next 10 days so I'll be able to address this around the 20th (should the PR be closed for now?)

nemerle:
MapServerData is loading data  the whole MapServer is using, so it ( MapServerData) could read the value from settings.cfg, and MapInstances already have access to MapServerData, so this would reduce the amount of variables being passed around :smile: 

Also, please leave this PR open, that way if any of the other current/future project contributors see that work on this has started

matthewswar:
How does this one look? Is there any preferred method beyond reading from `MapServer::ReadConfigAndRestart` and storing it in a public field inside `MapServerData`?

nemerle:
This looks good, at some point we might consider moving this variable to per-server startup script, but that's for the future we have those :)

Issue #271, opened 2018-03-12
Character Database is not updated automatically


broxen:
### Expected behavior
Character Dbase should update periodically (perhaps on the MapEvent::CharacterUpdate, or World Update?)

### Actual behavior/symptoms
Currently the only way to update your character in the database is to use the slash command `/save` or `/charupdate`

broxen:
I have commented code in World Update, but it's commented out, because it hangs the client when multiple entities are logged in.

PR #409, opened 2018-05-12
Issue #271 - Automatically Updating Characters in Database

eos428:
Issue #271

Goal: Characters can be automatically updated without slash commands /save and /charupdate

Files changed:

**WorldSimulation.cpp**

- In World::update, inside the for loop with entities, update the player characters every n seconds (in this case, 5 secs)
- charUpdateDb(e) is a placeholder atm

**Entity.cpp and .h**
- Added a function to check the if the player has a valid DbStoreFlag

To-do and/or wishlists:
- (To confirm with others) Call GameDbSyncContext::performUpdate based on the DbStoreFlag, and put the required data into the structs as parameters
- Modify slashCommand.cpp so that they actually update the database


broxen:
Previously calling DB updates inside of World::Update with multiple characters logged into the world would cause the server to hang and subsequently the client to crash.

This is just a reminder to test with multiple players logged in and flagged to update.

I can pull down later this week when our house guests head home, unless someone else can test sooner.

eos428:
Currently, I am not calling to GameDbSyncContext functions, where the actual updating happens, so a test using multiple characters doesn't do anything atm.

This brings me to my next question though - How do I get the DbSyncContext? I have a few ideas in mind. What I am doing so far is using HandlerLocator::getAuthDB_Handler()) to get GameDbAuthHandler, who in turn has the db we are looking for. So here's the ideas I have:
1. Create a function in GameDbSyncHandler to get just the db, then Entity.cpp calls in performUpdate directly from the db (does not need casting from CharacterUpdateData to CharacterUpdateMessage and so on)
2. Use GameDbSyncHandler's on_character_update. Entity.cpp does not have access to the db, but I have to cast to CharacterUpdateMessage (and Idk how to do that...) for example to call SyncHandler's function.

nemerle:
Accessing the correct GameDbSyncHandler should be done using HandlerLocator

Projects/CoX/Servers/GameDatabase/GameDBSyncService.cpp
```diff
@@ -55,110 +59,115 @@ void GameDBSyncService::on_update_timer(const ACE_Time_Value &tick_timer)
         // update the player characters in DB every n seconds
         if ((int)tick_timer.sec() % m_update_interval == 0)
         {
-            if (e->hasValidDbStoreFlag())
+            switch(e->m_db_store_flags)
             {
-
+            case (uint32_t)DbStoreFlags::Gui: sendGuiUpdateToHandler(e); break;
```
nemerle: m_db_store_flags is a flags value, so it can have multiple flags set, and should be handled like this
```cpp
if(e->m_db_store_flags & uint32_t(DbStoreFlags::Full))
{
  // do all the things.
  return;
}
if(e->m_db_store_flags & uint32_t(DbStoreFlags::Gui))
{
   // do gui thing
}
...
```

broxen:
This is failing CI because QVariant has no direct 'long unsigned int' support. So everywhere you've passed uint64_t to `bindValue()`
```
error: conversion from ‘const uint64_t {aka const long unsigned int}’ to ‘const QVariant’ is ambiguous
     m_prepared_gui_update->bindValue(QStringLiteral(":id"), data.m_id);
```

Try `QVariant::fromValue(data.m_id)`

broxen:
Looks like you're passing clang now but suffering from #424 on Mac.

I'll pull down for testing.

Projects/CoX/Servers/GameDatabase/GameDBSyncContext.cpp
```diff
@@ -190,13 +212,41 @@ bool GameDbSyncContext::performUpdate(const CharacterUpdateData &data)
...
     m_prepared_costume_update->bindValue(QStringLiteral(":costume_index"), data.m_costume_index);
     m_prepared_costume_update->bindValue(QStringLiteral(":skin_color"), data.m_skin_color);
     m_prepared_costume_update->bindValue(QStringLiteral(":parts"), data.m_parts);
     return doIt(*m_prepared_costume_update);
 }
 
+bool GameDbSyncContext::performUpdate(const GuiUpdateData &data)
+{
+    m_prepared_gui_update->bindValue(QStringLiteral(":id"), quint64(data.m_id));
+    m_prepared_gui_update->bindValue(QStringLiteral(":id"), data.m_gui);
```
broxen: This would bind to `:gui` instead of `:id` but in reality we're serializing `gui` `options` and `keybinds` to player_data now, so these binds aren't doing anything. 

We need to compare the current db player_data against the new gui data, and update db with a newly modified player_data... OR, we check if any of the three flags have been toggled (`GUI, Options, Keybinds`) and update all of player_data if so.

It seems to me that we should just update all of player_data, as we're sending it as one big serialized string, and the operations are probably not that resource intensive. I'm no expert though...

@nemerle may have more direction here.

Projects/CoX/Servers/GameDatabase/GameDBSyncService.cpp
```diff
@@ -0,0 +1,193 @@
...
+    else
+        postGlobalEvent(new ServiceStatusMessage({"GameDbSyncService loaded/configured",0}));
+    return result;
+}
+
+void GameDBSyncService::dispatch(SEGSEvent *ev)
+{
+    // We are servicing a request from message queue, using dispatchSync as a common processing point.
+    // nullptr result means that the given message is one-way
+    switch (ev->type())
```
broxen: Are we anticipating multiple types?

broxen:
Pulled c960f0a down for testing. Currently periodic saves, and forced saves `/save` appear to do nothing. Tested with single character logged in, and multiple characters logged in simultaneously.

Projects/CoX/Servers/GameDatabase/GameDBSyncService.cpp
eos428: I think this is an inherited function from SegsEvent and I must override this one IIRC. As far as how I work this one out, dispatch is essentially useless here.

Projects/CoX/Servers/GameDatabase/GameDBSyncContext.cpp
eos428: Oops, the second bindValue line was a typo on my part, my bad.

If we want to go away from updating the individual things, I guess we can also remove the flags for Gui, Options and Keybinds, since they are now encapsulated in PlayerData.

broxen: Yeah, this is my fault. I misled you here at the beginning, because I forgot that we had serialized everything together as PlayerData.

eos428:
Just had a test myself, yeah it's weird that the saves don't actually work, despite it going all the way from m_sync_service->update to DbContext and executing DoIt for their respective performUpdate functions.

broxen:
Testing this. Some outstanding issues and feedback:

- The periodic updates aren't working for me, but it does correctly save changes after adjusting trays, settings from the menu, or typing `/save`
- Logging back in after having saved a char to the database fails. There is no error message, just reaped stale links. My theory is that it's finding the wrong mapname and not loading properly?
- Several float values are saved in E notation, this may or may not matter.
- Since I can't log back in, I can't test the permanence of position updates or gui changes, but these values appear to change the in database.
- LastCostumeID and LastOnline aren't saving to the database. Maybe they were removed (or moved) during the architectural changes Nem made recently?

Regarding floats:
I'm not sure where these values came from? Maybe there were set automagically with the level change? The numbers don't seem to make sense though.
```
            "DamageType10": 7.2436189140602837e+22,
            "DamageType11": 1.5766134853564306e-19,
            "DamageType12": 4.367735527921468e-05,
            "DamageType13": 2.6512450013171929e-09,
```

eos428:
So far I think it is alright to save the ids in context with explicitly casting it to uint32_t, because they are already that (would make the cast redundant).

Now in response to your feedback...
1. The periodic updates are not supposed to work without typing in /save or likewise. All the entities are unmarked until one of those commands are activated. The to-do would be to create a check every time m_sync_service->update() is fired from MapInstance.
2 and 4. Ah, so that was the issue I had as well.
3. That is how floats are saved in the database. To save it in decimal form, use double or decimals instead (That was what I remembered in my previous work).
5. CharacterUpdateData does not have LastOnline or LastCostumeID. The values related to costume are merely m_body_type, m_physique and m_height.

With how float-point numbers are saved (they are not precise), they could lead to some miscalculations. I would suggest to save them with another type of value instead. This makes for good reading: https://stackoverflow.com/questions/30216575/why-float-epsilon-and-not-zero

Fun fact: The code in that question is from Unity's 2D Roguelike tutorial.

broxen:
> The to-do would be to create a check every time m_sync_service->update() is fired from MapInstance and mark/unmark the entities accordingly.

This makes the most sense. Ideally let's try to fit it into this PR, as that will truly close #271 

> With how float-point numbers are saved (they are not precise), they could lead to some miscalculations.

I'll defer to you and @nemerle on how best to proceed with this issue. IIRC the client stores values with 6 significant digits (so float) and I can't think of a concrete example where storing database values beyond 6 digits will impact the math significantly, but maybe I'm just not thinking far enough ahead (powers)?

eos428:
Thinking this again - checking for changes every time m_sync_service->update() is called might be a bad idea. We'd need to store its records from the previous update cycle (either in an array or querying the db for them - the latter an even worse idea) and compare all of the values inside individually, which is really inefficient.

Immediately marking the entity for db when something has changed (like HP, costume, etc... we'd done that already for PlayerData) is better for me.

As for the mapname issue, it might have to do with EntityData->m_map_idx, which is initialized as 1 atm.

eos428:
How about looking at it the other way around?

Update if: Player is affected by world (moved, took damage, etc) OR takes an action (earn xp, use powers, etc)
Else: Do not update.

Not updating if player is AFK or idle is also a good approach I think, and simpler to implement overall.

broxen:
Yeah either way.. I was just looking for the most elegant (simpliest) approach. Seems like we would update more often than not?

eos428:
If moving a character's position makes it eligible for an update, then I'd say yes :P

broxen:
I assume it would. In the original game just running around (traversing a map) would save in-between sessions.

Anecdotally, traversing Atlas Park and losing connection, my position would be saved (at the post before I timed out)

But maybe this is an argument for Flagging EntityData and CharData separately? EntityData contains Position and Orientation, CharData contains HP etc?

Projects/CoX/Servers/GameServer/GameHandler.cpp
```diff
@@ -454,4 +465,15 @@ void GameHandler::reap_stale_links()
                                      });
 }
 
+MapName GameHandler::checkMap(QString map_path)
+{
+    if (map_path == "segs_maps/City_00_01")
```
broxen: This will not work if a user has configured their map directory differently from you, and will fall through to "Outbreak".

The default is `maps` and mine (for instance) is `MapInstances`.  Maybe we should simply check against the string after the final slash `/`, or compile our own path using the stored mappath config value.

Projects/CoX/Servers/GameServer/GameHandler.cpp
```diff
@@ -365,18 +365,29 @@ void GameHandler::on_map_req(MapServerAddrRequest *ev)
...
-        break;
+        switch(checkMap(map_path))
+        {
+            case Outbreak: map_path = "maps/city_zones/city_00_01/city_00_01.txt"; break;
+            case AtlasPark: map_path = "maps/city_zones/city_01_01/city_01_01.txt"; break;
+        }
+    }
+    else
+    {
+        switch(ev->m_mapnumber)
```
broxen: In the future we'll have some sort of global vector of map values sorted by IDX that we can switch against.

broxen:
Aside from updating periodically (which we know is incomplete) this is now working as expected.

`/save` works as intended. Logging into the same character is successful.

eos428:
That's good news to hear. Now to resolve the remaining issues:

> But maybe this is an argument for Flagging EntityData and CharData separately? EntityData contains Position and Orientation, CharData contains HP etc?

Sure, we can make a separate flag to save just EntityData. Since I didn't play the original game, I'll defer to others on this.

> So maybe if player is AFK or Idle don't update, otherwise periodically do?

We could check if CharacterData.m_afk is true or not, but is having two different kinds of check (this and db store flags) necessary? I could create a temporary solution, in which if m_afk is false, we do a full update, but otherwise we check for the flags.

Another thing I'd like to add - we should also call removeEntityFromActiveList for World and SyncService's Entity managers every time a client exits the game (from one of MapInstance::on_client_quit, on_link_lost, on_disconnect AFAIK).

eos428:
> I guess I was thinking that when afk you would flag dbupdates to off, and when you return from afk we'd turn them back on?

I was thinking in the lines of an if clause with if (e->m_afk) then bla bla bla

> I'd say keep it simple for now, and we can always make updates more granular later.

True, this PR already does a lot of things imo. We can always continue later. I think this PR is ready for merge now, take a final look and see if there's anything else to add.

broxen:
I agree. The only thing that I feel like is missing is automatic periodic updates. Those still aren't working for me.

We don't need to check for any conditions, like afkness. I feel that the issue this PR addresses is only aimed at providing automatic updates to the database.

I'd like to achieve periodic database saves at the most basic level before merging. The granularity and/or other features we can add later.

Thoughts?

eos428:
Automatic periodic updates are not working until some update functions (like maybe in World) are able to change the flags for the entities. So at the moment I will force a full update until the other parts are done.

Also, sync service now has its own entity manager (not a reference to m_entities anymore). Trying to update the NPCs to the database is segmentation fault galore :D

broxen:
Pulled this down for testing. After your latest Push, I receive the following error SEGFault when logging out with a character:

```
authserver: /Segs/Projects/CoX/Servers/MapServer/EntityStorage.cpp:50: void EntityStore::release(Entity*): Assertion `src>m_map_entities.data() && src<m_map_entities.data()+m_map_entities.size()' failed.
```

```
void EntityStore::release(Entity *src)
{
    assert(src>m_map_entities.data() && src<m_map_entities.data()+m_map_entities.size()); // <---- SEGFAULT here
    src->m_destroyed = true;
    m_free_entries.push_back(src->m_idx);
}
```

That said, periodic updates work as expected!

PR #411, opened 2018-05-13
Fixed green reticle issue and corrected Target Assist

### LINKED ISSUE #392: Flashing green colored reticle artifact
### Summary
- Log into Outbreak
- ~~Select Police Drone (or other NPC)~~ also appears to happen even without selecting anything.
- Green Reticle artifact appears far away. You can move towards it, but it isn't "attached" to anything as far as I can tell

### Screenshot
screenshot from 2018-04-30 18-12-32

### Screencast
screencast from 04-30-2018

### SEGS version
build 17c664b

broxen:
Fixed green reticle issue and corrected Target Assist IDX

Closes #392 

- Fixed issues with sending Targets to the Client (fixing #392)
- Set `m_assist_target_idx` server side since the client never sends us this information

Outstanding:
- TODO: Outstanding issue with `/assist` because target is overwritten by client to old value

### DIFF
--- a/Projects/CoX/Servers/MapServer/SlashCommand.cpp
+++ b/Projects/CoX/Servers/MapServer/SlashCommand.cpp
@@ -1014,12 +1014,19 @@ void cmdHandler_MakeLeader(const QString &cmd, MapClientSession &sess)
 
 void cmdHandler_SetAssistTarget(const QString &/*cmd*/, MapClientSession &sess)
 {
-    QString msg = "Setting assist target.";
+    Entity *target_ent = getEntity(&sess, getTargetIdx(*sess.m_ent));
+    if(target_ent == nullptr)
+        return;
 
-    setAssistTarget(*sess.m_ent);
+    uint32_t new_target = getTargetIdx(*target_ent);
+    if(new_target == 0)
+        return;
 
+    setTarget(*sess.m_ent, new_target);
+
+    QString msg = "Now targeting " + target_ent->name() + "'s target";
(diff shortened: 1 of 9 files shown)


Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -1014,12 +1014,19 @@ void cmdHandler_MakeLeader(const QString &cmd, MapClientSession &sess)
...
+    Entity *target_ent = getEntity(&sess, getTargetIdx(*sess.m_ent));
+    if(target_ent == nullptr)
+        return;
+
+    uint32_t new_target = getTargetIdx(*target_ent);
+    if(new_target == 0)
+        return;
 
-    setAssistTarget(*sess.m_ent);
+    setTarget(*sess.m_ent, new_target);
```
broxen: Somehow this is overwritten by the client prior to the server-side update.

For the life of me I can't figure out why. So any input is appreciated.

broxen:
I meant to post this here...

Unfortunately @nemerle, our theory about unqueue_all did not resolve the /assist issue. It didn't have any impact.

In fact, unqueue_all only fires when switching target, so maybe we do want to reset m_target_idx and m_assist_target_idx there? I'll leave them for now. Unless you think it's better to comment them out.

broxen:
Since this PR resolves the original targeting issue #411  I'm going to go ahead and merge this, and open a new issue for the `/assist` problem. Especially since without any sort of powers, using `/assist` is really low priority anyway.

Issue #402, opened 2018-05-03
We need to run full_update at least once


broxen:
### Expected behavior
Looks like disabling full update results in new characters spawning at 0, 0, 0 even though their position is correctly set to spawn points.

Toggling full_update fixes the issue, so it looks like we need to run full_update at least once.

broxen:
Actually further testing shows that this doesn't necessarily solved the issue.

Needs more testing.

PR #412, opened 2018-05-13
Send full_update on non-incremental entity packets.

nemerle:
Might work for #402

Needs more testing


broxen:
This is only working for me about 25% of the time. The rest of the time, any new (or existing) characters are spawning at <0, 0, 0> still

toggling fullupdate will set them to the correct position on the map.

PR #416, opened 2018-05-15
Make NetCommandManager send all commands to the client

### LINKED ISSUE #376: NetCommandManager should build per-access-level command lists
Currently all commands in `void NetCommandManager::addCommand( NetCommand *cmd )` are added to the same `m_commands_level0` array, we should have per-access-level arrays, or better yet a:
```cpp
std::map<int, std::vector<NetCommand *cmd> > m_access_level_commands;
```
and then in `NetCommandManager::SendCommandShortcuts` we should send all commands from the player's access-level and those from levels below.

dracc:
Closes #376 & #420 .

- NetCommandManager::SendCommandShortcuts now sends all known netcommands.

### DIFF
--- a/Projects/CoX/Common/GameData/clientoptions_definitions.h
+++ b/Projects/CoX/Common/GameData/clientoptions_definitions.h
@@ -62,16 +62,21 @@ public:
 static const constexpr uint32_t class_version = 1;
 
     // Other Options
-    int32_t control_debug       = 0;
+    int32_t controldebug       = 0;
     int32_t no_strafe           = 0;
     int32_t alwaysmobile        = 0; // 1 - player is always mobile (can't be immobilized by powers)
     int32_t repredict           = 0; // 1 - client is out of sync with server, asking for new physics state info.
     int32_t neterrorcorrection  = 0;
     float   speed_scale         = 0;
 
-    int32_t svr_lag,svr_lag_vary,svr_pl,svr_oo_packets,client_pos_id;
+    int32_t svr_lag,svr_lag_vary,svr_pl,svr_oo_packets,client_pos_id,pause;
     int32_t atest0,atest1,atest2,atest3,atest4,atest5,atest6,atest7,atest8,atest9;
-    int32_t predict,notimeout,selected_ent_server_index;
+    int32_t predict,notimeout,selected_ent_server_index,disablegurneys,canlook;
+    int32_t camrotate,forward,backward,up,down,nocoll,nosync,turnleft,turnright;
+    int32_t zoomin,zoomout,lookup,lookdown,third,first,mouse_invert,autorun;
+    int32_t left,right,nostrafe;
+
+    float time,timescale,timestepscale,speed_turn,velscale,yaw,mouse_speed;
 
     // Options
     bool    m_first_person_view = false;
--- a/Projects/CoX/Servers/MapServer/NetCommandManager.cpp
+++ b/Projects/CoX/Servers/MapServer/NetCommandManager.cpp
@@ -14,6 +14,7 @@
 #include "MapClientSession.h"
 
 #include <vector>
+#include <algorithm>
 
 static void FillCommands()
 {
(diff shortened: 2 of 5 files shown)


dracc:
Would someone please look at my code to see what I'm doing wrong? I get stuck in what seems like an infinite loop when trying to exit.
<a href="https://i.imgur.com/YZKP9vu.png">
  
</a>
I've been waiting for shutdown close to 10 minutes now.

Projects/CoX/Common/GameData/clientoptions_definitions.h
```diff
@@ -62,7 +62,7 @@ class ClientOptions
 static const constexpr uint32_t class_version = 1;
 
     // Other Options
-    int32_t control_debug       = 0;
+    int32_t controldebug       = 0;
```
broxen: Watch this. It's probably used elsewhere in the code base. Make sure we rename all instances

dracc: `grep -ri control_debug` only returned that line. :open_mouth:

broxen: :+1:

dracc:
```
[log.mapevents]Warning : Unhandled command event type 3
[log.mapevents]Debug   : bits:  4
[log.mapevents]Critical: Not all bits were consumed by previous commands:
[log.mapevents]Warning : 11WindowState
[log.mapevents]Warning : 14ConsoleCommand
[log.slashcommand]Debug   : Unknown game command: "×\u007F\u0002\u0007ÿÿÿ\u007FD\u007FÌ\u0099\u007F"
```

This message now shows up. Apart from that, it seems to be working (somewhat).

dracc:
So, current state;
12138 extra bits on "enter paragon city" until an access level 9 account has entered.
After that, it's 398 extra bits (sometimes more(?!)) unless access level 9 which get no extra bits.

broxen:
We have an issue here in MapEventTypes.h, we should only have one of these (seems like evShortcuts is the one) and it should be 4, instead of 2 or 3.

```
EVENT_DECL(evShortcutsRequest           ,2)
EVENT_DECL(evShortcuts                  ,3)
```

Projects/CoX/Servers/MapServer/NetCommandManager.cpp
```diff
@@ -149,32 +169,25 @@ void NetCommandManager::serializeto(BitStream &tgt, const vNetCommand &commands,
...
-                client->AddShortcut(i,m_commands_level0[i]);
-            serializeto(tgt,m_commands_level0,commands2);
-            break;
-        default:
-            // add shortcuts to client's shortcut map.
-            for (size_t i = 0; i<m_commands_level0.size(); ++i)
-                client->AddShortcut(i, m_commands_level0[i]);
-            serializeto(tgt, m_commands_level0, commands2);
-    }
+    std::for_each(begin(m_access_level_commands),
```
broxen: You could still use range-based for loop here, like so
```
for (const auto &cmd : *m_access_level_commands)
{
    if(cmd->m_required_access_level > client->m_access_level)
        continue;

    client->AddShortcut(cmd_iterator, cmd);
    cmd_iterator++;
    permitted_cmds.push_back(cmd);
}
```

broxen:
Still looking into this dracc

dracc:
@broxen I feel as if I've hit a dead end. I've no idea of how to move forward.

Projects/CoX/Servers/MapServer/NetCommandManager.cpp
```diff
@@ -24,40 +25,59 @@ static void FillCommands()
...
-
-    cmd_manager->addCommand(new NetCommand(3,"time",fargs));
-    cmd_manager->addCommand(new NetCommand(3,"timescale",fargs));
-    cmd_manager->addCommand(new NetCommand(3,"timestepscale",fargs));
-    cmd_manager->addCommand(new NetCommand(3,"pause",args));
-    cmd_manager->addCommand(new NetCommand(3,"disablegurneys",args));
-//    cmd_manager->addCommand(new NetCommand(9,"nodynamiccollisions",args));
-//    cmd_manager->addCommand(new NetCommand(9,"noentcollisions",args));
-//    cmd_manager->addCommand(new NetCommand(9,"pvpmap",args));
+    cmd_manager->addCommand(new NetCommand(0,"time",fargs));
```
broxen: These were level 3 before? I think they still need to be

dracc: I'd very much like them to be, but setting them to level 3 gives lower access level clients a crash on world entry.

broxen: That seems unrelated because Client side code expects these as level 3. Also we sent them as level 3 before? Or just remove them all together and see what happens.

Also, let's run through the entire list and check access_levels.

broxen: Or maybe comment all but one for level 0 in an effort to test to see if the client expects a certain number, or just can't have more than expected, or what...

dracc:
But if we set them to be level 3, they wont be sent to lower level accounts (unlike before) which then crashes. Ah well, I'll give it another go next time I'm at my computer. :)

Projects/CoX/Servers/MapServer/NetCommandManager.cpp
dracc: Wait a minute, if the client expects the commands to have specific access levels but crashes unless the first five are sent... Does the client know the accounts access level from elsewhere? If that's the case, maybe it wants the complete list of commands even though it's not allowed to use all of them?

broxen: It's possible. The client checks against the list I posted in issue #420, but as far as I can tell only checks to see if the sent command exists (names match).

It's also worth noting that that list in #420 came from three locations that all seem to relate to this same recieve code in various ways (You can tell by the indents). So it's possible that some of them should be sent and some sent elsewhere? It's weird because they all relate. I need to dig in further.

broxen: The way to test would be to send all, send some, send one or two, send one per level. Etc.

Also test inside the client each time if they are usable.

dracc:
So, sending all NetCommands to everyone seems to work (somewhat).
Some NetCommands seem to get lost somewhere? (For example, `cmds time` gives nothing even for a level 9 account)
Level 9 access level commands are not reachable for lower level accounts in current client version.

dracc:
So. Further testing has ensued.
`cmds a` gives a shorter list when the client access level goes lower, even when all commands are sent to the client. If a client tries to run a command it does not have the required access level for, the server spits out a `Debug   : Unknown game command: "[commandname]"` warning.

broxen:
@dracc I recall seeing some further discussion on this between you and @nemerle in the discord. What's the status? Is this ready for testing?

dracc:
@broxen The current status is that it's somewhat broken. I need help debugging to see what is supposed to happen. If every NetCommand is sent to everyone, all is fine and dandy. If only "the correct" commands are sent, errors/warnings pop up in the server console and the client itself start acting up. It seems thus far that not every command ought to be sent, even though the client (and server) currently behaves best when they are.

broxen:
@dracc @nemerle as far as I can tell SendCommands should send everything and the client filters by accesslvl on it's side of things using the access_level we've already sent it for the current User/Character.

I feel this way because the client checks incoming commands against `OptionGroup tOptBundle[]` and verifies access level? Why do we feel that we need to trim the list server side? Am I missing something client side?

broxen:
@dracc do you want me to wait for further progress RE: EntityResponse, or test the current improvements?

broxen:
@dracc please filter the list in both sendCmds and EntityResponse. The lists absolutely must match.

AFAICT the client does not NEED us to filter the list server side, as it does a test locally with Access Level. That said, it doesn't appear that the client needs the full list either. My recommendation is to follow @nemerle's direction, and filter the list on both ends, so that the server maintains authority.

Ya know, in case of the jerk hackers! ;)

broxen:
Closing this for now. Let's reopen when ready.

PR #421, opened 2018-05-16
Compiler warning fixes to address #352

ldilley:
This pull request addresses any remaining compiler warnings outlined in #352 with the exception of those stemming from Qt GUI-related casts pertaining to callback functions. As a result, SEGSAdmin and SLAV will remain affected until the Qt developers address the reported casting issues up to 5.9. `Q_UNUSED` is employed as self-documenting way to quell unused var warnings. It is easily searchable and can be used to locate and address each finding in the future. Various casts and format specifiers were also fixed which should help with truncation in certain cases.

### DIFF
--- a/Projects/CoX/Common/NetStructures/Team.cpp
+++ b/Projects/CoX/Common/NetStructures/Team.cpp
@@ -251,6 +251,7 @@ void addSidekick(Entity &tgt, Entity &src)
     Sidekick    &tgt_sk = tgt.m_char->m_char_data.m_sidekick;
     uint32_t    src_lvl = getLevel(*src.m_char);
     uint32_t    tgt_lvl = getLevel(*tgt.m_char);
+    Q_UNUSED(tgt_lvl);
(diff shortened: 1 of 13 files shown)


ldilley:
I just want to ensure that MSVC 19.x does not choke on the `%zu` format specifier prior to merging.

Projects/CoX/Common/NetStructures/Team.cpp
```diff
@@ -251,6 +251,7 @@ void addSidekick(Entity &tgt, Entity &src)
     Sidekick    &tgt_sk = tgt.m_char->m_char_data.m_sidekick;
     uint32_t    src_lvl = getLevel(*src.m_char);
     uint32_t    tgt_lvl = getLevel(*tgt.m_char);
```
broxen: You can even get rid of this here. We don't need it in this method and was probably copypasta from the invite method which checks level range.

PR #422, opened 2018-05-17
Screen resolution setGeometry warning fix & IPLookup

### LINKED ISSUE #414: SEGSAdmin produces setGeometry warning on startup
### Expected behavior
SEGSAdmin runs without warning

### Actual behavior/symptoms
Running SEGSAdmin produces this warning on startup:
`setGeometry: Unable to set geometry 764x849+8+31 on QWidgetWindow/'SEGSAdminToolWindow'. Resulting geometry:  764x822+8+31 (frame: 8, 31, 8, 8, custom margin: 0, 0, 0, 0, minimum size: 596x822, maximum size: 764x16777215).`

Run SEGSAdmin and the warning should display in console. SEGSAdmin still runs normally (afaik) despite the warning.

### SEGS version
Latest master
### Operating system
Windows 10

### LINKED ISSUE #419: Teach SEGSAdmin to list available IP addresses
In cases where the configuration ui needs a list of local addresses, it would be nice to list available addresses ?
```cpp
foreach (const QHostAddress &address, QNetworkInterface::allAddresses()) {
    if (address.protocol() == QAbstractSocket::IPv4Protocol && address != QHostAddress(QHostAddress::LocalHost))
         qDebug() << address.toString();
}```

### PR DESCRIPTION
**Pull request by miles200**
* Restructure of UI to accomodate lower screen resolutions
* Couple of fixes for auth server label change and enable/disable logic
* Add IPLookup functionality, users can now lookup local system IP's
  rather than entering manually

Closes #419 
Closes #414

### CODE REVIEW

**Comment by broxen** [comment]
This is awesome! Great work Miles!

For the IP lookup, maybe we should assume IPs ending in 1 are gateways and ignore them?

I'm just thinking of potential troubleshooting issues "which IP do I choose?"

File: Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp [hunk 1]
```diff
@@ -59,6 +60,10 @@ SEGSAdminTool::SEGSAdminTool(QWidget *parent) :
     connect(m_set_up_data,&SetUpData::dataSetupComplete,this,&SEGSAdminTool::check_data_and_dir);
     connect(m_set_up_data,&SetUpData::getMapsDir,this,&SEGSAdminTool::send_maps_dir);
     connect(this,&SEGSAdminTool::sendMapsDir,m_set_up_data,&SetUpData::create_default_directory);
+    // GetIP Signals/Slots
+    m_get_ip = new GetIPDialog;
```
nemerle: Consider passing `this` to dialog/widget/window constructors, this way when `this` is closed, it will correctly close it's children.

miles200: hey @nemerle sorry I'm not completely sure I understand.. are you saying I should move the signals to the GetIP class rather than sitting it in SEGSAdminTool class?

nemerle: Nope. If you're creating an instance of a class inheriting directly/indirectly from QWidget, call it with a `parent` parameter, in this particular case
```cpp
m_get_ip = new GetIPDialog(this);
```

miles200: Thanks @nemerle that makes sense. Been busy with work last few days will make this change tonight hopefully 😄

broxen:
I'll pull this down tomorrow to test.

CI is failing due to #424

broxen:
This seems to work fine for me. My only suggestion would be to add player fade-in to config creation and the settings menu, to avoid the warning message in the console output about it not being found.

That said, the console output is kind of small and impractical, but a general GUI clean up can fix that, per our prior discussion on discord

miles200:
Thanks @broxen 

I should have some time this weekend at some point to tidy up the GUI and adding fade-in should be easy enough. 

Thanks,
Miles

miles200:
@broxen This can be merged once Travis CI is fixed. I've already incorporated the player_fade_in and UI changes locally, I'll submit in a separate PR once this is gone.

PR #434, opened 2018-06-05
Added busy username error text in dbtool.

### LINKED ISSUE #432: SQL_ERROR upon using SEGSAdmin to create an account with an existing username
1. Run SEGSAdmin
2. In "Add New User" create an account and enter an already existing account's username (entries for password and access level shouldn't matter) 
3. Should produce: SQL_ERROR: QSqlError("19", "Unable to fetch row", "UNIQUE constraint failed: accounts.username") 
### SEGS version
Latest master
### Operating system
Windows 10

Maybe SEGSAdmin should prompt the user that an account with the username already exists and to select another?

dracc:
Closes #432  .

I've started thinking, maybe this could take on a more general approach?
A helper function somewhere which takes a `QSqlError const& q` as argument and depending on the error code prints a prettily formatted error?

### DIFF
--- a/Projects/CoX/Utilities/dbtool/main.cpp
+++ b/Projects/CoX/Utilities/dbtool/main.cpp
@@ -33,6 +33,11 @@
 #include <iostream>
 #include <vector>
 
+// Return value enumeration
+enum { SUCCESS, SETTINGS_MISSING, DBFOLDER_MISSING, NOT_FORCED,
+       NOT_ENOUGH_PARAMS, SQLITE_DB_MISSING, DB_RM_FAILED, DB_CONN_FAILED,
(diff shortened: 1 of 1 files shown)


broxen:
Yes remove abort. qFatal should end the process immediately but gracefully

dracc:
abort() removed from qFatal.
qFatal usage restored.

nemerle:
Gents, I don't think we should be changing the semantics of the code,  qFatal docs state:
```
Calls the message handler with the fatal message message. If no message handler has been installed, the message is printed to stderr. Under Windows, the message is sent to the debugger. On QNX the message is sent to slogger2.

If you are using the default message handler this function will abort to create a core dump. On Windows, for debug builds, this function will report a _CRT_ERROR enabling you to connect a debugger to the application.```

PR #444, opened 2018-06-08
Improve readability in several MapServer files

redseiko:
- Add more detailed startup message in MapServer and some refactoring to use constants. Removed an member field.
- Fix a typo in MapInstance::spin_up_for() and add a startup message.
- Convert qCDebug() statements to qDebug() statements in MapManager to print them in the console.

### DIFF
--- a/Projects/CoX/Servers/MapServer/MapInstance.cpp
+++ b/Projects/CoX/Servers/MapServer/MapInstance.cpp
@@ -197,7 +197,7 @@ bool MapInstance::spin_up_for(uint8_t game_server_id,uint32_t owner_id,uint32_t 
     if (m_endpoint->open() == -1) // will register notifications with current reactor
         ACE_ERROR_RETURN ((LM_ERROR, "(%P|%t) MapInstance: ServerEndpoint::open\n"),false);
 
-    qDebug() << "Spun up MapInstance" << m_instance_id "for MapServer" << m_owner_id;
+    qInfo() << "Spun up MapInstance" << m_instance_id << "for MapServer" << m_owner_id;
 
     return true;
 }
--- a/Projects/CoX/Servers/MapServer/MapInstance.cpp
+++ b/Projects/CoX/Servers/MapServer/MapInstance.cpp
@@ -191,11 +191,13 @@ bool MapInstance::spin_up_for(uint8_t game_server_id,uint32_t owner_id,uint32_t 
     m_instance_id = instance_id;
     if (ACE_Reactor::instance()->register_handler(m_endpoint,ACE_Event_Handler::READ_MASK) == -1)
     {
-        qWarning() << "MapInstance::spin_up_for faile to register_handler, port already open";
+        qWarning() << "MapInstance::spin_up_for failed to register_handler, port already open";
         return false;
     }
     if (m_endpoint->open() == -1) // will register notifications with current reactor
         ACE_ERROR_RETURN ((LM_ERROR, "(%P|%t) MapInstance: ServerEndpoint::open\n"),false);
+
+    qDebug() << "Spun up MapInstance" << m_instance_id "for MapServer" << m_owner_id;
 
     return true;
 }
--- a/Projects/CoX/Servers/MapServer/MapManager.cpp
+++ b/Projects/CoX/Servers/MapServer/MapManager.cpp
@@ -45,7 +45,8 @@ MapManager::~MapManager()
 bool MapManager::load_templates(const QString &template_directory, uint8_t game_id, uint32_t map_id,
                                 const ListenAndLocationAddresses &loc)
 {
-    qCDebug(logSettings) << "Starting the search for maps in" << template_directory;
+    qDebug() << "Starting the search for maps in:" << template_directory;
+
     QDirIterator map_dir_visitor(template_directory, QDir::Dirs|QDir::NoDotAndDotDot, QDirIterator::Subdirectories);
     while (map_dir_visitor.hasNext())
     {
@@ -55,7 +56,8 @@ bool MapManager::load_templates(const QString &template_directory, uint8_t game_
             auto tpl = new MapTemplate(map_dir_visitor.fileInfo().filePath(), game_id, map_id, loc);
             m_templates[s_map_name_to_id[tpl->base_name()]] = tpl;
             m_name_to_template[tpl->client_filename()] = tpl;
-            qCDebug(logSettings) << "Detected a map"<<map_dir_visitor.fileInfo().filePath();
+
+            qDebug() << "Detected map:" << map_dir_visitor.fileInfo().filePath();
         }
     }
     // (template_directory / "tutorial.bin")
--- a/Projects/CoX/Servers/MapServer/MapManager.cpp
+++ b/Projects/CoX/Servers/MapServer/MapManager.cpp
@@ -45,7 +45,7 @@ MapManager::~MapManager()
 bool MapManager::load_templates(const QString &template_directory, uint8_t game_id, uint32_t map_id,
                                 const ListenAndLocationAddresses &loc)
 {
-    qDebug() << "Starting the search for maps in:" << template_directory;
+    qInfo() << "Searching for maps in:" << template_directory;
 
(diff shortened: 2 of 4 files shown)


Projects/CoX/Servers/MapServer/MapServer.h
```diff
@@ -42,10 +45,10 @@ class MapServer : public EventProcessor
 
         std::unique_ptr<PrivateData> d;
 
-        uint8_t                 m_id = 1;
-        uint8_t                 m_owner_game_server_id = 255;
-        QString                 m_serverName;
+        uint8_t                 m_id = {1};
```
dracc: I'm not familiar with this format for assignment/initialization, why not the old format (`m_id = 1;`) or just plain `m_id{1};`? :slightly_smiling_face:

Projects/CoX/Servers/MapServer/MapServer.h
```diff
@@ -42,10 +45,10 @@ class MapServer : public EventProcessor
...
+        uint8_t                 m_owner_game_server_id = {kInvalidGameServerId};
         ACE_INET_Addr           m_base_location; //! this is the base map instance address
         ACE_INET_Addr           m_base_listen_point; //! this is used as a base map listening endpoint
 };
-extern MapServer *g_GlobalMapServer;
+
+extern MapServer *g_GlobalMapServer;
```
dracc: Please make sure there's a newline at EOF.

Projects/CoX/Servers/MapServer/MapServer.h
nemerle: agreed, if using in-class initialization the code should look like `member = value` for basic types/pointers, and ` member { init1,init2 }` for others.

Projects/CoX/Servers/MapServer/MapServer.h
```diff
@@ -21,6 +21,9 @@ class MapInstance;
 class MapServerData;
 class MapManager;
 
+static constexpr uint8_t kInvalidGameServerId= {255};
```
broxen: Let's aim for consistency with variable names. This should be `g_invalid_game_server_id` instead.

Projects/CoX/Servers/MapServer/MapServer.h
```diff
@@ -21,6 +21,9 @@ class MapInstance;
...
+static constexpr char kRuntimeDataPath[] = {"./data/bin/"};
```
broxen: Same here, this should be `g_runtime_data_path`

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -191,12 +191,14 @@ bool MapInstance::spin_up_for(uint8_t game_server_id,uint32_t owner_id,uint32_t
...
     if (ACE_Reactor::instance()->register_handler(m_endpoint,ACE_Event_Handler::READ_MASK) == -1)
     {
-        qWarning() << "MapInstance::spin_up_for faile to register_handler, port already open";
+        qWarning() << "MapInstance::spin_up_for failed to register_handler, port already open";
         return false;
     }
     if (m_endpoint->open() == -1) // will register notifications with current reactor
         ACE_ERROR_RETURN ((LM_ERROR, "(%P|%t) MapInstance: ServerEndpoint::open\n"),false);
 
+    qDebug() << "Spun up MapInstance" << m_instance_id "for MapServer" << m_owner_id;
```
broxen: If this is intended to be end-user facing, then we should use `qInfo()`, otherwise, let's give it a category and use `qCDebug()` instead. Debug categories will allow us to toggle on/off these debug messages in settings.cfg or from in-game using slash commands.

Projects/CoX/Servers/MapServer/MapManager.cpp
```diff
@@ -45,7 +45,8 @@ MapManager::~MapManager()
 bool MapManager::load_templates(const QString &template_directory, uint8_t game_id, uint32_t map_id,
                                 const ListenAndLocationAddresses &loc)
 {
-    qCDebug(logSettings) << "Starting the search for maps in" << template_directory;
+    qDebug() << "Starting the search for maps in:" << template_directory;
```
broxen: Again, `qInfo()` or `qCDebug()`

Projects/CoX/Servers/MapServer/MapManager.cpp
```diff
@@ -55,7 +56,8 @@ bool MapManager::load_templates(const QString &template_directory, uint8_t game_
             auto tpl = new MapTemplate(map_dir_visitor.fileInfo().filePath(), game_id, map_id, loc);
             m_templates[s_map_name_to_id[tpl->base_name()]] = tpl;
             m_name_to_template[tpl->client_filename()] = tpl;
-            qCDebug(logSettings) << "Detected a map"<<map_dir_visitor.fileInfo().filePath();
+
+            qDebug() << "Detected map:" << map_dir_visitor.fileInfo().filePath();
```
broxen: Again, `qInfo()` or `qCDebug()`

Projects/CoX/Servers/MapServer/MapManager.cpp
nemerle: or qCInfo() :P

Projects/CoX/Servers/MapServer/MapServer.h
redseiko: Wouldn't g_ in imply global (which this is just a local constant not visible outside of this .h/.cpp). I googled and read that k_ is the general prefix for constants.

broxen: True `g_` indicates global. I thought this was a global variable because you defined it in the header, outside of a class or namespace, where it can be accessed by any file that includes `MapServer.h`. If it's a static local variable then we should put it in the .cpp file and prefix it with `s_` instead, so that the scope is understood. This shares more information than a simple `k` by providing us hints about variable accessibility.

Projects/CoX/Servers/MapServer/MapManager.cpp
broxen: Oh yeah! Or that 😛

Projects/CoX/Servers/MapServer/MapServer.h
nemerle: Hah, @broxen this is a very specific construction, @redseiko is defining a named constant here. In older c++ versions people did something like:
```cpp
enum
{
  INVALID_GAME_SERVER_ID = 255,
};
```
to `name` a constant value.
since in some ways the construction @redseiko  used is replacement of that, I'd suggest:
```cpp
static constexpr uint8_t INVALID_GAME_SERVER_ID = 255;
```

broxen: Ahh cool. TIL

redseiko: So... what's the consensus on the constants variable naming?

What @nemerle proposed is pretty much how I do it in Java land. I did research for C++ and they said to use k_ or something like that.

nemerle: `k_` prefix is just a way some projects do things like that.
UPPERCASE_UNDERSCORE_SEPARATED_WORDS would be the most consistent with existing code

redseiko:
Added a second commit with changes as discussed.

Projects/CoX/Servers/MapServer/MapServer.h
```diff
@@ -21,6 +21,9 @@ class MapInstance;
 class MapServerData;
 class MapManager;
 
+static constexpr uint8_t INVALID_GAME_SERVER_ID = 255;
+static constexpr char RUNTIME_DATA_PATH[] = "./data/bin/";
+
```
nemerle: one last informational comment:
Remember that the c++ standard says, that you are able to retrieve an address of constexpr values, so if it can't be sure that the address of the variable is never taken, it will be forced to create a place to store the constexpr in the resulting executable.

nemerle:
This looks good, thank you @redseiko

PR #446, opened 2018-06-08
Changing CharData::m_mapName to EntityData::m_current_map

### LINKED ISSUE #302: Serialize MapName (and other MapInfo?) and move mapname to EntityData
### aka, MapName shows `City_00_01` instead of `Atlas Park`
Feel free to provide feedback or input on direction for this Enhancement.

### Summary
Unless the Client has map display names hardcoded, there must be a way for us to pull that map name out for use throughout the server (e.g. FriendsList, TeamsList, Char Select, etc)

There may also be other info that we'll want to store in some kind of new MapStruct for use as MapInstance begins to recognize multiple zones.

Additionally, currently `m_mapname` is stored in `CharacterData` under Character. In reality, we should move this to `EntityData` inside Entity, as every Entity should know what map it's in (besides every Character is an Entity). It's possible that we'll want to rename this to `m_current_map` or something instead.

There are some comments in `map_serializers.cpp` that may provide insight into what's stored in maps.bin
```
//static Serialization_Template Map_Tokens[6] = {
//{ "Name", TOKEN_STRING, 0x10},
//{ "Icon",  TOKEN_STRING, 0x14},
//{ "Location",  TOKEN_VEC2, 0},
//{ "TextLocation",  TOKEN_VEC2, 8},
//{ "End",  TOKEN_END},
//{ },
//};

//static const Serialization_Template Map_Token[2] = {
//{ "Zone", TOKEN_SUB_TABLE, 0, 0x18, Map_Tokens},
//{},
//};
```
### Tasks to complete this issue:
- [ ] Research what is stored client side related to map data/info (maps.bin, maybe others)
- [ ] Serialize relevant map data in map_serializers.cpp using Cereal
- [ ] Rename `m_mapname` to `m_current_map`
- [ ] Move `m_current_map` to `EntityData` struct

eos428:
I didn't expect to make so many changes... To summarize, this is about removing m_mapName from CharData and creating m_current_map in EntityData. I might be making too much changes at one go here.

This is compilable but untested, I will be awaiting your inputs before going in further.
- Updated serializers for CharData and EntityData
- Incremented class versions for affected structs
- Character::from/toActualCharacter now takes in EntityData as parameter, where it is going to be serialized to/from QString
- Character::reset and getCurrentCostume no longer deals with map names
- Created uniquePtr for EntityData in Entity
- GameAccountResponseCharacterData now has m_serialized_entity_data, which will be used by GameHandler
- GameHandler::on_map_req now gets the map name from EntityData.m_current_map
- DataHelpers::getMapName and setMapName now takes in Entity as parameter

Some things to do off the top of my head:

- Test da ting (I feel a little scared to test this now tbh, with all the changes made :P)
- Fill up EntityData::reset and dump functions with something
- DataHelpers::getMapName is now unused, call it from somewhere else

Closes #302 .

### DIFF
--- a/Projects/CoX/Common/GameData/chardata_serializers.cpp
+++ b/Projects/CoX/Common/GameData/chardata_serializers.cpp
@@ -42,6 +42,7 @@ void serialize(Archive &archive, Friend &fr, uint32_t const version)
     archive(cereal::make_nvp("FriendClass", fr.m_class_idx));
     archive(cereal::make_nvp("FriendOrigin", fr.m_origin_idx));
     archive(cereal::make_nvp("FriendMapIdx", fr.m_map_idx));
+    archive(cereal::make_nvp("FriendMapName", fr.m_mapname));
 }
 
 template<class Archive>
--- a/Projects/CoX/Common/GameData/chardata_serializers.cpp
+++ b/Projects/CoX/Common/GameData/chardata_serializers.cpp
@@ -42,7 +42,7 @@ void serialize(Archive &archive, Friend &fr, uint32_t const version)
     archive(cereal::make_nvp("FriendClass", fr.m_class_idx));
     archive(cereal::make_nvp("FriendOrigin", fr.m_origin_idx));
     archive(cereal::make_nvp("FriendMapIdx", fr.m_map_idx));
-    archive(cereal::make_nvp("FriendMapName", fr.m_mapname));
+    archive(cereal::make_nvp("FriendMapname", fr.m_mapname));
 }
 
 template<class Archive>
--- a/Projects/CoX/Common/GameData/chardata_serializers.cpp
+++ b/Projects/CoX/Common/GameData/chardata_serializers.cpp
@@ -42,7 +42,6 @@ void serialize(Archive &archive, Friend &fr, uint32_t const version)
     archive(cereal::make_nvp("FriendClass", fr.m_class_idx));
     archive(cereal::make_nvp("FriendOrigin", fr.m_origin_idx));
     archive(cereal::make_nvp("FriendMapIdx", fr.m_map_idx));
-    archive(cereal::make_nvp("FriendMapname", fr.m_mapname));
 }
 
 template<class Archive>
@@ -101,7 +100,6 @@ void serialize(Archive &archive, CharacterData &cd, uint32_t const version)
     archive(cereal::make_nvp("LastOnline",cd.m_last_online));
     archive(cereal::make_nvp("Class",cd.m_class_name));
     archive(cereal::make_nvp("Origin",cd.m_origin_name));
-    archive(cereal::make_nvp("MapName",cd.m_mapName));
     archive(cereal::make_nvp("SuperGroupCostume",cd.m_supergroup_costume));
     archive(cereal::make_nvp("UsingSGCostume",cd.m_using_sg_costume));
     archive(cereal::make_nvp("SideKick",cd.m_sidekick));
--- a/Projects/CoX/Common/GameData/entitydata_definitions.h
+++ b/Projects/CoX/Common/GameData/entitydata_definitions.h
@@ -13,11 +13,18 @@
 
 struct EntityData
 {
-static const constexpr  uint32_t    class_version       = 3;
+static const constexpr  uint32_t    class_version       = 4;    // v4: adds m_current_map
                         uint32_t    m_access_level      = 0;
                         uint8_t     m_origin_idx        = {0};
                         uint8_t     m_class_idx         = {0};
                         glm::vec3   m_pos;
                         glm::vec3   m_orientation_pyr;          // Stored in Radians
                         uint32_t    m_map_idx           = 1;    // only 1 map instance
+                        QString     m_current_map;
+
+                        void reset()
+                        {}
+
+                        void dump()
+                        {}
 };
--- a/Projects/CoX/Common/GameData/entitydata_definitions.h
+++ b/Projects/CoX/Common/GameData/entitydata_definitions.h
@@ -21,21 +21,4 @@ static const constexpr  uint32_t    class_version       = 4;    // v4: adds m_cu
                         glm::vec3   m_orientation_pyr;          // Stored in Radians
                         uint32_t    m_map_idx           = 1;    // only 1 map instance
                         QString     m_current_map;
-
-                        void reset()
-                        {}
-
-                        void dump()
-                        {}
-
-                        QString getMapName()
-                        {
-                            if (m_current_map.contains("City_00_01", Qt::CaseInsensitive))
-                                return "Outbreak";
-                            if (m_current_map.contains("City_01_01", Qt::CaseInsensitive))
-                                return "AtlasPark";
-
-                            // let's default to Outbreak in case things go wrong
-                            return "Outbreak";
-                        }
 };
--- a/Projects/CoX/Common/GameData/entitydata_definitions.h
+++ b/Projects/CoX/Common/GameData/entitydata_definitions.h
@@ -27,4 +27,15 @@ static const constexpr  uint32_t    class_version       = 4;    // v4: adds m_cu
 
                         void dump()
                         {}
+
+                        QString getMapName()
+                        {
+                            if (m_current_map.contains("City_00_01", Qt::CaseInsensitive))
+                                return "Outbreak";
+                            if (m_current_map.contains("City_01_01", Qt::CaseInsensitive))
+                                return "AtlasPark";
+
+                            // let's default to Outbreak in case things go wrong
+                            return "Outbreak";
+                        }
 };
--- a/Projects/CoX/Common/NetStructures/Friend.cpp
+++ b/Projects/CoX/Common/NetStructures/Friend.cpp
@@ -44,7 +44,7 @@ void addFriend(Entity &src, Entity &tgt)
     f.m_class_idx       = tgt.m_entity_data.m_class_idx;
     f.m_origin_idx      = tgt.m_entity_data.m_origin_idx;
     f.m_map_idx         = tgt.m_entity_data.m_map_idx;
-    f.m_mapname         = tgt.m_entity_data.m_current_map;
+    f.m_mapname         = getEntityMapName(tgt.m_entity_data);
 
     // add to friendlist
     src_data->m_friends.emplace_back(f);
--- a/Projects/CoX/Common/NetStructures/Friend.cpp
+++ b/Projects/CoX/Common/NetStructures/Friend.cpp
@@ -44,7 +44,7 @@ void addFriend(Entity &src, Entity &tgt)
     f.m_class_idx       = tgt.m_entity_data.m_class_idx;
     f.m_origin_idx      = tgt.m_entity_data.m_origin_idx;
     f.m_map_idx         = tgt.m_entity_data.m_map_idx;
-    f.m_mapname         = tgt.m_char->m_char_data.m_mapName;
+    f.m_mapname         = tgt.m_entity_data.m_current_map;
 
     // add to friendlist
     src_data->m_friends.emplace_back(f);
--- a/Projects/CoX/Servers/MapServer/DataHelpers.cpp
+++ b/Projects/CoX/Servers/MapServer/DataHelpers.cpp
@@ -328,9 +328,9 @@ const QString &     getAlignment(const Character &c) { return c.m_char_data.m_al
 
 static const std::vector<MapData> g_defined_map_datas =
 {
-    {0, "City_00_01", "maps/city_zones/city_00_01/city_00_01.txt", "Outbreak", {0, 0, 0}},
-    {1, "City_01_01", "maps/city_zones/city_01_01/city_01_01.txt", "Atlas Park", {0, 0, 0}},
-    {23, "City_23_01", "maps/city_zones/city_23_01/city_23_01.txt", "Galaxy City", {0, 0, 0}}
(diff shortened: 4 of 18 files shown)


eos428:
So I was like, well let's just expect the unexpected with testing and jumped in.

This is what happens when on a newly created account (minus the created admin character):
blob:https://imgur.com/5113b582-75bd-4da7-8798-ca8dce56ea93

Every character slot (except the ones created by the player) are now treated as bad data O_O. You can try to enter the game with them, but MapInstance.cpp line 660 will give out an assertion failure. The created admin character can go into the game without a hitch.

Projects/CoX/Common/GameData/entitydata_definitions.h
```diff
@@ -13,11 +13,29 @@
...
                         uint32_t    m_map_idx           = 1;    // only 1 map instance
+                        QString     m_current_map;
+
+                        void reset()
+                        {}
+
+                        void dump()
+                        {}
+
+                        QString getMapName()
```
broxen: I think we should move this out of the struct, and into DataHelpers as a generic getMapName method that we can access from anywhere, as we'll use this for CharSel Screen, Team List and SG List as well.

Projects/CoX/Common/NetStructures/Friend.cpp
```diff
@@ -44,7 +44,7 @@ void addFriend(Entity &src, Entity &tgt)
     f.m_class_idx       = tgt.m_entity_data.m_class_idx;
     f.m_origin_idx      = tgt.m_entity_data.m_origin_idx;
     f.m_map_idx         = tgt.m_entity_data.m_map_idx;
-    f.m_mapname         = tgt.m_char->m_char_data.m_mapName;
+    f.m_mapname         = tgt.m_entity_data.m_current_map;
```
broxen: We can utilize the aforementioned getMapName() method from DataHelpers here.

We should probably also check if the Character is offline and show "OFFLINE" instead of map name.

Projects/CoX/Common/GameData/chardata_serializers.cpp
```diff
@@ -42,7 +42,7 @@ void serialize(Archive &archive, Friend &fr, uint32_t const version)
     archive(cereal::make_nvp("FriendClass", fr.m_class_idx));
     archive(cereal::make_nvp("FriendOrigin", fr.m_origin_idx));
     archive(cereal::make_nvp("FriendMapIdx", fr.m_map_idx));
-    archive(cereal::make_nvp("FriendMapname", fr.m_mapname));
+    archive(cereal::make_nvp("FriendMapName", fr.m_mapname));
```
broxen: Make sure to iterate the version for this Friend struct.

Projects/CoX/Common/GameData/entitydata_definitions.h
```diff
@@ -13,11 +13,29 @@
...
                         uint32_t    m_access_level      = 0;
                         uint8_t     m_origin_idx        = {0};
                         uint8_t     m_class_idx         = {0};
                         glm::vec3   m_pos;
                         glm::vec3   m_orientation_pyr;          // Stored in Radians
                         uint32_t    m_map_idx           = 1;    // only 1 map instance
+                        QString     m_current_map;
+
+                        void reset()
+                        {}
```
broxen: Since we're not using these methods, they can be removed. I personally prefer structs to be POD containers, and utilize classes when adding logic such as this. In this case we're don't use them either way.

broxen:
I will pull this down for testing after the next round of changes.

eos428:
To add to my commit message, I also removed MapName enum in GameHandler, as those are now stuff handled in DataHelpers' function (For GameHandler's part, specifically the new getMapPath() function).

But here's the bad thing - all of a sudden I'm reaping stale links again when trying to enter the map. I have no idea why since I didn't do a test run after merging with upstream/develop (my bad). Can you help check for me?

broxen:
Will pull this down for testing tonight

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -327,6 +326,42 @@ const QString &     getDescription(const Character &c) { return c.m_char_data.m_
 const QString &     getBattleCry(const Character &c) { return c.m_char_data.m_battle_cry; }
 const QString &     getAlignment(const Character &c) { return c.m_char_data.m_alignment; }
 
+const QString getMapName(const QString &map_name)
+{
+    if (map_name.contains("City_00_01", Qt::CaseInsensitive))
+        return "Outbreak";
+    if (map_name.contains("City_01_01", Qt::CaseInsensitive))
+        return "AtlasPark";
```
broxen: This should have a space "Atlas Park"

broxen: Also, you could add here:
```
if (map_name.contains("City_23_01", Qt::CaseInsensitive))
        return "Galaxy City";
```

broxen:
@eos428 everything looks good to me. Those two comments are my only real input.

Friends online status doesn't work, but that's not because of your code, it's because we never update the online status or mapname of players in your friendslist. Truth is, we should probably look those values up realtime instead of saving them to the dbase. That's another issue altogether.

broxen:
The Enum and helper functions are a fine solution, but spread the relevant data across the code base. What about using a vector of a Map data struct, and storing those values like an array?

This would be similar to how we store SlashCommands.

`std::vector<int m_map_idx, MapData m_map_data>` or something.

Would make it easier to add new maps and update existing with changes, and we can reference maps by their map idx without entering them sequentially

Projects/CoX/Servers/MapServer/DataHelpers.h
```diff
@@ -20,9 +20,13 @@ struct Friend;
 struct FriendsList;
 struct MapClientSession;
 
-enum MapData
+struct MapData
 {
-    Outbreak, AtlasPark, GalaxyCity
+    uint32_t m_map_idx;
```
broxen: @nemerle do we assign IDX ourselves? Or does the client expect certain maps to have precise IDXs?

I don't see anything in the client to suggest either. So maybe we can assign our own?

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -326,30 +326,23 @@ const QString &     getDescription(const Character &c) { return c.m_char_data.m_
...
-        return AtlasPark;
-    if (map_name.contains("City_23_01", Qt::CaseInsensitive))
-        return GalaxyCity;
-
-    // let's default to Outbreak in case things go wrong
-    return Outbreak;
-}
+    {0, "City_00_01", "maps/city_zones/city_00_01/city_00_01.txt", "Outbreak", {0, 0, 0}},
+    {1, "City_01_01", "maps/city_zones/city_00_01/city_01_01.txt", "Atlas Park", {0, 0, 0}},
+    {2, "City_23_01", "maps/city_zones/city_00_01/city_23_01.txt", "Galaxy City", {0, 0, 0}}
```
broxen: This might have an IDX of 23. That's a supposition based upon the city name "City_23_01" and may or maybe not be true. I dont see anything in the client to suggest we __have__ to do it this way, and the server architecture may not support out of sequence IDXs. So I'm going to punt this over to @nemerle for review.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -326,30 +326,23 @@ const QString &     getDescription(const Character &c) { return c.m_char_data.m_
...
-    if (map_name.contains("City_01_01", Qt::CaseInsensitive))
-        return AtlasPark;
-    if (map_name.contains("City_23_01", Qt::CaseInsensitive))
-        return GalaxyCity;
-
-    // let's default to Outbreak in case things go wrong
-    return Outbreak;
-}
+    {0, "City_00_01", "maps/city_zones/city_00_01/city_00_01.txt", "Outbreak", {0, 0, 0}},
+    {1, "City_01_01", "maps/city_zones/city_00_01/city_01_01.txt", "Atlas Park", {0, 0, 0}},
```
broxen: These paths may have a typo "City_01_01", "maps/city_zones/__**city_00_01**__/city_01_01.txt"

Projects/CoX/Servers/MapServer/DataHelpers.h
```diff
@@ -20,9 +20,13 @@ struct Friend;
...
+    QString m_map_name;             // City_00_01, City_01_01, etc...
+    QString m_map_path;             // The ones ending with .txt
+    QString m_display_map_name;     // Outbreak, Atlas Park...
+    glm::vec3 m_spawn_location;
```
broxen: This is calculated using the "spawners" in the map itself. We don't need to include it here.

broxen: Just saw your commit comment about it. Is there a reason I may be overlooking to include them? Perhaps some kind of fallback?

eos428: This was your comment on the issue page:

> -- Add this zone to the zones table
outbreak = {
  zone_id = 1,
  name = "City_00_01",
  displayname = "Outbreak",
  customspawn = {x = -61, y = -0, z = 161}, -- y is the elevation
}
zones.insert(outbreak, outbreak.zone_id)

m_spawn_location is based on customspawn. I guess it may be needed if you make a custom map?

broxen: Ohh yes. I was thinking in the future we may want to override the default spawn points.

My vote would be to exclude it for now, as it may cause confusion, and when we add zone script overrides we add it then.

eos428: Oh sure, no problem. Removed it by latest commit.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -327,6 +326,49 @@ const QString &     getDescription(const Character &c) { return c.m_char_data.m_
...
+    return getMapName(f.m_mapname);
+}
+
+const QString getMapPath(const EntityData &ed)
+{
+    for (uint32_t i = 0; i < g_defined_map_datas.size(); i++)
+    {
+        if (ed.m_current_map.contains(g_defined_map_datas[i].m_map_name))
+            return g_defined_map_datas[i].m_map_path;
+    }
```
nemerle: range-for loop could be used here:
```cpp
for(const auto &map_data : g_defined_map_datas)
{}
```

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -327,6 +326,49 @@ const QString &     getDescription(const Character &c) { return c.m_char_data.m_
 const QString &     getBattleCry(const Character &c) { return c.m_char_data.m_battle_cry; }
 const QString &     getAlignment(const Character &c) { return c.m_char_data.m_alignment; }
 
+static const std::vector<MapData> g_defined_map_datas =
+{
+    {0, "City_00_01", "maps/city_zones/city_00_01/city_00_01.txt", "Outbreak"},
+    {1, "City_01_01", "maps/city_zones/city_01_01/city_01_01.txt", "Atlas Park"},
+    {23, "City_23_01", "maps/city_zones/city_23_01/city_23_01.txt", "Galaxy City"}
+};
```
nemerle: The idx does not matter much.
Note:
This whole `g_defined_map_datas` thing looks like a prime candidate on being 'cerealized' from json.  That way people can add 'support' for new maps using fairly readable json files.
Feel free to add a new **Issue** regarding that.

nemerle:
Overall, this looks good and makes things more robust.

What is interesting to me, is the reason for moving map name to EntityData, are we planning on allowing NPC to roam between maps ? :wink:

broxen:
The move to entity data was at my direction. Of course we don't anticipate NPCs to roam between maps, but it makes more sense to me because that's where we store positional data? All Entities will have position and be assigned to a map, but not all entities will have character data?

nemerle:
Admit it, you just want the mob trains to follow some poor soul through the whole world :D

broxen:
CHOOO CHOOO! 🚂 🚋 🚋 🚋

Projects/CoX/Servers/MapServer/DataHelpers.cpp
eos428: I thought of making this a key-value pair, where idx is the key, but that probably would complicate matters - both in getting the data (although there's optimization, albeit minimal since this array wouldn't be so big) and cerealizing them. But as usual, feel free to decide how you want to structure it :P

eos428:
NPCs that roam between maps actually sound like a fun concept :D Think of a bounty hunter or bodyguard system, for example. If LUA scripting can implement that, then I'd like to try that!

broxen:
Testing this...

Currently it's not showing the correct map_display_name on CharSel screen or in FriendsList (or TeamsList). It's still showing the full map path. Perhaps we should save IDX to the database, instead of name, and reference your `g_defined_map_data` where applicable?

I don't want to expand the scope more than necessary,  as there are valuable changes here, but as a baseline I think the CharSel screen should show the proper mapname prior to merge, as this was the original focus of #302 

As an aside, I'll open a new issue for implementing this array in JSON per Nem's recommendation.

eos428:
Oops, my mistake. I returned m_map_name instead of m_display_map_name. We do have MapIdx value in EntityData and we could use that for reference instead if you like.

broxen:
I guess ultimately it doesn't matter if we use IDX or mapname, as long as the server knows what's what

broxen:
7ec5734 works great!

PR #450, opened 2018-06-09
[WIP] bring the client patch up to date

nemerle:
moving local changes to repository


dracc:
OMG it's happening! :smile:

mobbyg:
I thought this was abandoned? LOL

dracc:
Are there any builds of this available?
I have a version of it from a month or so ago which gives me "Failed to create OpenGL rendering context." and then it exits. No further output given when run from a terminal. All this when started through wine, as I don't use Windows.

dracc:
Okay, so, currently this sits at 12000+ added lines. Something tells me not many will proof-read this before it gets merged. :sweat_smile:
Looking forward to testing it tomorrow though!

nemerle:
I expect we will work on it some more before considering a merge anyway

dracc:
I fetched your branch, get an error about a missing TextureGenerator.h when running cmake for the client. Any ideas?

broxen:
I don't have an Intel card. How can I help test and merge this patch?

broxen:
I think we should resolve merge conflicts here and merge this in. This is significant progress toward support for other cards, and by merging this in we can establish a foundation in the code base if someone new comes along looking to help complete/bug test.

Thoughts?

PR #454, opened 2018-06-12
Added missing "libwinpthread-1.dll" library to MINGW build.

deciton666:
This change makes it possible to run the executables in the "out" directory without using QtCreator.

Closes nothing as I have not opened any issue for this.
If you want me to create an issue first, then I will do so.

- When I tried to run any built executable in my Win8/MINGW development environment it failed with an error message telling me that it could not find the "libwinpthread-1.dll" library. Running from QtCreator works fine though. This change adds the missing library to the "out" directory, just as the other MINGW specific files are.


nemerle:
Let me check msys2 libs, they might have different dll names.

nemerle:
It's all good, thank ye :)

PR #455, opened 2018-06-12
Fixed a crash in "addNpc" slash command when executing it without any parameters.

deciton666:
Closes nothing as I have not opened any issue for this.
If you want me to, then I will do so.

- Fixed a crash in "addNpc" slash command when executing it without any parameters. There was a missing "return" statement after a sanity check.


nemerle:
Good catch :)

PR #465, opened 2018-06-14
License fix

dracc:
Update license text to reflect the change from LICENSE.txt -> LICENSE.md , and add what kindnof license it is.


broxen:
As part of this, can you change all the Authors.txt to AUTHORS.md as I renamed that as well.

You can add `[skip ci]` to the first line of your next push to skip Travis and avoid those errors

PR #467, opened 2018-06-17
SEGSAdmin Field Validation Update

### LINKED ISSUE #433: 'Player Fade In' field for SEGSAdmin should be given default value of 380.0 to avoid error message
1. Run SEGSAdmin
2. Start Authserver
3. Should produce: `Critical: Badly formed float for 'player_fade_in':  ""`
### SEGS version
Latest master
### Operating system
Windows 10

As I understand it, leaving the field blank will just make the server default to 380.0, but it would be nice to avoid any panic the user may have from receiving an error message and not knowing what value to put in.

### LINKED ISSUE #437: SEGSAdmin Account Creation allows for username to exceed the amount of characters the client allows
1. Add New User in SEGSAdmin with a username of 15+ characters
2. When logging into client, you will be limited to 14 characters

Note: You don't need to run AuthServer to observe that the client limits the username to 14 characters.
### SEGS version
Latest master
### Operating system
Windows 10

miles200:
Closes #437 
Closes #433 

**This update to SEGSAdmin introduces the following**

- Validation on the following fields

server name - 32
max players - counter, max 200
max char slots - counter, 8
map location - 32	
account username - 14
account password - 14

- Add validation on blank fields on saving Settings. If blank, user is shown the problem fields by use of highlight and a warning message
- Fixes a crash on opening 'Settings & Logging' dialog if settings.cfg file does not exist yet
- Add QComboBox for selection of DB Drivers
- player_fade_in now utilises a slider and max value of 1000
- player_fade_in will save as a float ensuring .0 is appended to the saved value

**Update 06/07**

Add spin boxes for below port fields, now validate for a value between 1024 and 65536:

char db port
acc db port
auth port
map list port 
map loc port
game list port
game loc port

Add a check on username and password that they are 'Latin' (toLatin1()) and do not contain invalid UTF-8 characters


miles200:
I'm hoping this is ready for merge, something went wrong with my squash and ended up changing some code.

miles200:
Scrap that I've just realised I've forgotten to put the .tolatin check that Nem mentioned. I'll update this shortly

broxen:
I'll wait for your next push to review. Thanks for all of your hard work miles!

dracc:
Port numbers, would it be possible to check that they are less than 2^16 (65536) ?

broxen:
That is a good idea

miles200:
I can :) would validating a value between 00001 and 65536 work? Is there a
lower limit?

miles200:
@dracc @broxen I've added port validation now and the toLatin1 check for username and password, should be good to go, please review and merge when ready :)

broxen:
I'll pull this down for testing.

Projects/CoX/Utilities/SEGSAdmin/SettingsDialog.cpp
```diff
@@ -37,11 +48,30 @@ SettingsDialog::~SettingsDialog()
...
+
 void SettingsDialog::open_settings_dialog()
 {
     QFileInfo config_file("settings.cfg");
     QString config_file_path = config_file.absoluteFilePath();
     SettingsDialog::read_config_file(config_file_path);
+    QList<QLineEdit *> all_line_edits = ui->tab_settings->findChildren<QLineEdit *>();
+    foreach(QLineEdit* le, all_line_edits)
+    {
+        le->setStyleSheet("background-color: rgb(255, 255, 255)");
```
broxen: This causes an unfortunate side effect on dark themes, where the white background blends with the white text to make these fields unreadable.

Instead of making white, can you reset to "default color" somehow?

broxen: screenshot from 2018-07-20 11-35-30

miles200:
Good point, let me see what I can do... there must be a reset or “no
colour” option

broxen:
The port fields only accept 4 characters, so cannot reach the maximum value of `65536`

The lower limit appears to work as expected.

broxen:
Other than these two small UI issues, this appears to be ready to merge.

Once you're satisfied with those two problems, let me know and I'll pull the trigger.

Thank you @miles200! You rock brotha

dracc:
Screenshot
Looks good to me. 👍

miles200:
Thanks guys! Think this can be merged then

PR #476, opened 2018-06-21
Issue #474: Fixed crash after deleting a character and then creating …

### LINKED ISSUE #474: Server-Client character synchronization failure!
### Steps to Reproduce
1. Log in to account with existing characters
2. Delete Existing Character. Do not log out, or log into another character.
3. Create a new character in the same slot where you deleted one previously.
4. Receive Assertion and server crash
```
ACE_ASSERT: file /Segs/Projects/CoX/Servers/GameServer/GameHandler.cpp, line 406 assertion failed for 'selected_slot->m_name == ev->m_char_name || !"Server-Client character synchronization failure!"'.Aborting...
```

deciton666:
…a new one in the same slot.

The cause of this issue was that a copy of the character slot was reset, not the original one (in GameHandler::on_character_deleted).
The fix was to use a reference instead of a copy to access the data.

The same error was done in (GameHandler::on_delete_character), but it did not cause any damage as nothing were written to the slot here.
I changed the copy to a (const) reference here as well to reduce the risk of someone doing the same mistake again.

Closes #474 .

- Fix for character delete+create crash.
- Some minor code improvement to make the modified methods more similar to the "GameHandler::on_map_req" method when it comes to variable names and types.

### DIFF
--- a/Projects/CoX/Servers/GameServer/GameHandler.cpp
+++ b/Projects/CoX/Servers/GameServer/GameHandler.cpp
@@ -309,8 +309,8 @@ void GameHandler::on_link_lost(SEGSEvent *ev)
 void GameHandler::on_character_deleted(RemoveCharacterResponse *ev)
 {
     GameSession &session = m_session_store.session_from_event(ev);
-    auto chr(session.m_game_account.get_character(ev->m_data.slot_idx));
-    chr.reset();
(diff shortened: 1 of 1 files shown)


nemerle:
Awesomly bugfix'd @deciton666

PR #477, opened 2018-06-23
Creating a semi-standalone EmailHandler that inherits from EventProcessor

### LINKED ISSUE #367: Create a semi-standalone EmailHandler
Following the common SEGS pattern, EmailHandler should be a inheriting from EventProcessor, and be accessible from HandleLocator. All email operations and notifications should be processed by that EmailHandler.

eos428:
Closes #367.

This is a very barebones EmailHandler made at the moment. Not sure how to proceed with this, so I'm putting this up now to get some advice.

- Made an EmailHandler that inherits from EventProcessor, and accessible from HandlerLocator
- Created EmailEvents.h to define events to fire to the EmailHandler

Things to do:
- Actually fill up those functions. The email functions at SlashCommand.cpp seem to be good reference on what to do?
- I looked at EmailHeaders, EmailMessageStatus, EmailRead, all of which are derived from GameCommand and have their MapEventTypes. Should we just use those and not use EmailEvents? Also, since they are classes I'm not sure we can use ONE_WAY_MESSAGE because they are made from structs.
- EmailHandler should have a reference to GameDbSyncContext because sending and deleting emails have something to do with the database?
- Need to initialize EmailHandler somewhere


eos428:
Self-memo.

- EmailHandler needs to find the corresponding MapInstance to queue responses to based on the contents of m_stored_client_datas
- Need to get the db id of the recipient when sending email header, I think look into the entity manager of the related MapInstance based on name?
- For EmailWasReadByRecipient event, it should be easy once email is incorporated into the DB (just put the sender's db id into the email). But right now, we should be looking for the sender... from the sender's name in the EmailHeader the recipient received

broxen:
Hi eoS! I know you've been making progress here. I will wait for your prompt before testing for merge.

eos428:
It'll be a while before it's ready, by my own predictions. Hopefully this can be ready in 2 weeks time. Since the email handler will be similar to the friend handler, if that finishes faster I'd be able to copy the homework and adjust it for email handler.

eos428:
Renamed branch to emailhandler

PR #485, opened 2018-07-17
Add all maps to `g_defined_map_datas`

### LINKED ISSUE #483: Fill up Defined Map Datas vector in DataHelpers
Related to #475.

At the moment, this is our g_defined_map_datas vector in DataHelpers.cpp, which stores all the Map information:

> static const std::vector<MapData> g_defined_map_datas =
> {
>     {0, "City_00_01", "maps/city_zones/city_00_01/city_00_01.txt", "Outbreak"},
>     {1, "City_01_01", "maps/city_zones/city_01_01/city_01_01.txt", "Atlas Park"},
>     {23, "City_23_01", "maps/city_zones/city_23_01/city_23_01.txt", "Galaxy City"}
> };

Here's an explanation of all the parameters in this vector, in order from left to right:
1. m_map_idx. The map... index. Currently unused and we could modify the code to use this, but that's not for this PR.
2. m_map_name. This is what the server checks against a character's EntityData's CurrentMap in order to load up the map.
3. m_map_path. The path the server looks for in order to load up the maps (from maps.bin AFAIK). The convention is pretty easy to follow, so I don't feel the need to explain too much.
4. m_display_name. The name of the map displayed to players.

What we should do, is to fill up this vector with information of all the other areas. Basically whatever City_02_01 is, City_03_01, etc etc. The convention is pretty easy to follow, so I think this would be a good first issue.

dracc:
Closes #483 .

- Populate `g_defined_map_datas` in [DataHelpers.cpp](https://github.com/Segs/Segs/blob/develop/Projects/CoX/Servers/MapServer/DataHelpers.cpp) with all cities, hazards and trials
- Re-index "Galaxy City" from 23 to 3, as it seems it's actually located in `City_01_03` rather than `City_23_01`
- Make sure to actually look for Trial and Hazard maps on startup


Projects/CoX/Servers/MapServer/MapTemplate.cpp
```diff
@@ -57,6 +57,14 @@ QString MapTemplate::client_filename() const
 QString MapTemplate::base_name() const
 {
     int city_idx     = m_map_filename.indexOf("City_");
+    if(city_idx == -1)
```
broxen: You could use regex to return the idx of the first underscore to make this nested if statement more efficient.

`m_map_filename.indexOf(QRegExp("^(.+?)"), 0);`

Projects/CoX/Servers/MapServer/MapManager.cpp
```diff
@@ -24,7 +24,30 @@ static QHash<QString,int> s_map_name_to_id =
 {
     {"city_00_01",0}, // Outbreak
     {"city_01_01",1}, // Atlas Park
-    {"city_01_03",29}, // Galaxy City
+    {"city_01_02",2}, // King's Row
```
broxen: I would capitalize these to match the format in g_defined_map_datas

dracc: As both those data entries are available in the `g_defined_map_datas` I was thinking of removing this QHash.

broxen: Yeah that's a better idea.

Projects/CoX/Servers/MapServer/MapTemplate.cpp
```diff
@@ -56,7 +56,7 @@ QString MapTemplate::client_filename() const
 
 QString MapTemplate::base_name() const
 {
-    int city_idx     = m_map_filename.indexOf("City_");
+    int city_idx     = m_map_filename.indexOf('_');
     int end_or_slash = m_map_filename.indexOf("/", city_idx);
     assert(city_idx != -1);
     return m_map_filename.mid(city_idx, end_or_slash == -1 ? -1 : m_map_filename.size() - end_or_slash).toLower();
```
broxen: Here your `.mid` is including city_idx. So your results are `_00_01`... You want city_idx+1 here

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -328,55 +328,145 @@ const QString &     getAlignment(const Character &c) { return c.m_char_data.m_al
 
 static const std::vector<MapData> g_defined_map_datas =
 {
-    {0, "City_00_01", "maps/city_zones/city_00_01/city_00_01.txt", "Outbreak"},
-    {1, "City_01_01", "maps/city_zones/city_01_01/city_01_01.txt", "Atlas Park"},
-    {23, "City_23_01", "maps/city_zones/city_23_01/city_23_01.txt", "Galaxy City"}
+    // City_Zones
+    {0, "City_00_01", "maps/City_Zones/City_00_01/City_00_01.txt", "Outbreak"},
```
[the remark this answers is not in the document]
dracc: The "maps/city_zones" part is not configurable currently, it's part of the `.pigg`s structure as far as I've gathered.

dracc:
Should I add the map folders to be placed in `DefaultMapInstances` in CMakeLists.txt?

broxen:
I can't remember if we add them in cmake or with the SEGSAdmin tool. I say go ahead and add them, and we can use the admintool to ensure they exist.

dracc:
Latest commit adds changes to CMakeLists.txt, trying to create the necessary folders in `DefaultMapInstances`.
But seeing as it's 23 folders, it will become rather ugly. Are there any programmatic ways?
I tried `City_0{0,1,2,3,4,5}_01` as name, but that did not work the same way as in bash.

PR #486, opened 2018-07-18
[WIP] Movement, Velocity, and Interpolation

### LINKED ISSUE #94: [Movement] Rubberbanding movement and Interpolation

broxen:
Closes #94 

THIS IS IN PROGRESS AND SUBMITTED FOR REVIEW ONLY.

I'm putting this out there, because it's a much larger undertaking than I realized initially, and I'm hoping some feedback will be encouraging.

Feedback is greatly appreciated. The code is definitely a mess, so forgive me that.

Currently this contains the following changes:
- Remove temporary physics and update code from #241 and it's brethren.
- Add Velocity calculations
- Add PosUpdates during Entity load, and on Update
- Fixed rotation check in update_rot()
- Correctly handle keypress_release
- Separated InputState and InputStateStorage from InputStateEvent
- Moving toggles AFK
- Added some helper functions for interp testing
- In general movement is better, and movement of other entities is very smooth and mostly accurate position-wise.
- Added seqbits enum for future seq bits =D
- Code cleanup

TODO:
- Oh soooo very much...
- Keypress states still aren't fully saved with each state
- Keypress times aren't fully saved
- There's still a small amount of stuttering when moving
- Players move vertically, instead of staying on the ground. Gravity? Or, more likely, I need to ignore the y axis prior to using direction to update position.
- Interpolation crashes the client? Not sure why, something in bintree.
- Changed PosUpdates for bintree conversion to vector from array size 9. It might need to be a fixed size, and this could relate to the crashes when toggling interpolation
- TimeStates should probably be a part of InputStateStorage
- Movement flags and sequences should probably be saved along the way, but this may be beyond my original scope (eliminate rubberbanding)


Projects/CoX/Servers/MapServer/Events/InputState.h
```diff
@@ -1,29 +0,0 @@
-/*
```
dracc: `git mv` instead of removing and re-creating?

broxen: From the git faq:
> Git has a rename command git mv, but that is just a convenience. The effect is indistinguishable from removing the file and adding another with different name and the same content

dracc: Oh. Til, thanks!

broxen:
Will jump back on this after wrapping up #495

Projects/CoX/Servers/MapServer/EntityUpdateCodec.cpp
```diff
@@ -110,44 +115,60 @@ bool storePosition(const Entity &src,BitStream &bs)
...
-bool update_rot(const Entity &/*src*/, int axis ) /* returns true if given axis needs updating */
+bool update_rot(const Entity &src, int axis ) /* returns true if given axis needs updating */
 {
-    int axis_tmp = axis; // Used to quell tautological comparison warning
-    if(axis==axis_tmp)   // FixMe: var compared against same var.
+    if(src.m_prev_state == nullptr)
         return true;
-    return false;
+
+    if(src.m_prev_state->m_orientation_pyr[axis] == src.m_cur_state->m_orientation_pyr[axis])
```
broxen: After running idle for several minutes, the server will segfault on this line.

Maybe it's due to m_prev_state being a pointer?

broxen: This should be resolved with the latest refactoring 32852f1

broxen:
Closing this until it's closer to completion and ready for actual testing.

PR #496, opened 2018-07-27
Allowing DbHandler to get entity data using character name.

eos428:
It should work almost exactly like GetEntityRequest and GetEntityResponse, the only thing is using a character's name instead of db_id as the search criteria. I think my EmailHandler can use this, and perhaps other handlers that are made for player to player interactions can, as well. In the end, it is possible that we might not use this at all though :P

This is untested atm, and I have made nothing that would use these changes in this PR (so far). To test this, I will change up MapInstance's on_create_map_entity so that it queues GetEntityByNameRequest instead of the standard GetEntityRequest. I don't think I should commit this change to MapInstance, because the default way is already correct, and thus changing it is unnecessary.

### DIFF
--- a/Projects/CoX/Servers/GameDatabase/GameDBSyncContext.cpp
+++ b/Projects/CoX/Servers/GameDatabase/GameDBSyncContext.cpp
@@ -389,15 +389,15 @@ bool GameDbSyncContext::getEntity(const GetEntityRequestData &data, GetEntityRes
 
 bool GameDbSyncContext::getEntityByName(const GetEntityByNameRequestData &data, GetEntityByNameResponseData &result)
 {
-    m_prepared_entity_select->bindValue(0, data.m_char_name);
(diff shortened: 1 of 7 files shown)


eos428:
Ready for testing and merging! Once done, comment MapInstance.cpp line 699 (where GetEntityRequestByName is used) and un-comment line 702 (where the usual GetEntityRequest is used)

broxen:
Will try to pull this down for testing.

@eos428 can you clarify your last comment? Before merging do you need to change those lines back? My understanding is that this method is for future use, correct?

eos428:
Yes, this method is for (potential) future use. So far MapInstance has always used GetEntityRequest to get entitydata from db, and I don't wish to change it. The entire implementation of GetEntityRequestByName in MapInstance is just to give a working example :)

broxen:
This works as expected, and I see no errors after multiple tests.

Go ahead and revert the comments and prepare this for merge. Thank you eoS!

PR #500, opened 2018-07-30
Improvements to Sync Service

### LINKED ISSUE #498: Improve performance and structure of GameDbSyncService
Currently, the Sync Service shares the same entity manager with the entity manager its related MapInstance uses. This is not good, because MapInstance's entity manager contains NPC entity, and so during every update check, the Sync Service has to iterate through those NPCs as well, which are NOT updated to the database.

My idea of changing things up is to do away with the entity manager for Sync Service and instead use a vector of pointers to entities, started from scratch when the Sync Service is created. We don't need the extra functions the entity manager provides, so the vector should be enough for the Sync Service's purpose.

Also, delete the unused functions and set the update functions (except updateEntities(), which is called by MapInstance every few seconds) to private, because we are not supposed to open them for others to use.

eos428:
Closes #498 .

Removed unused variables and functions, moved some of the public ones to private because they shouldn't be accessible to outsiders. And most importantly, improved the sync service's performance by giving it only players' entities.

So far it seems alright in testing, but as always - test again to ensure it's good for merging.

Some note: I cannot delete the pointer to entity because entity's destructor is private. Perhaps we should use a vector of smart pointers instead?


Projects/CoX/Servers/GameDatabase/GameDBSyncService.cpp
```diff
@@ -41,38 +46,35 @@ void GameDBSyncService::set_db_handler(const uint8_t id)
...
+    // point the pointer to null
+    e = nullptr;
 }
 
 void GameDBSyncService::updateEntities()
 {
-    ACE_Guard<ACE_Thread_Mutex> guard_buffer(ref_entity_mgr.getEntitiesMutex());
-
-    for(Entity * e : ref_entity_mgr.m_live_entlist)
+    for(int i = 0; i < m_entities.size(); ++i)
```
dracc: Nit: Why not a range based for-loop anymore?

nemerle:
Not at this point though, all Entity instances are be 'owned' only at a single point, so making those shared_ptr would muddle things a bit, I would even consider making most of `Entity *` into EntityHandles that are accessing EntityManager::m_store

Projects/CoX/Servers/GameDatabase/GameDBSyncService.h
```diff
@@ -5,37 +5,33 @@
...
-
+#include <QVector>
 
 class GameDBSyncService : public EventProcessor
 {
 private:
-    EntityManager& ref_entity_mgr;
-    uint8_t m_game_server_id;
-    int m_update_interval = 5;
+    QVector<Entity *> m_entities;
```
nemerle: I'd very much prefer we don't store `Entity` pointers here, since they can change ( underlying memory is owned by EntityManager, and is part of a std::vector)

Projects/CoX/Servers/GameDatabase/GameDBSyncService.cpp
eos428: i < size/length/count whatever is what i nearly always use :P

eos428:
So I suppose we should make this EntityHandle class that uses the EntityManager's m_store... That makes sense because we'd want the entity to be put into m_free_entries back like some sort of 'entities pool'. The way I look at it, just having access to m_store is not enough, so should we make EntityHandler have a reference instead of a pointer to Entity, and if we have pointers, would the memory no longer be shared between it and EntityManager?

eos428:
Ready for merge. Testing is okay with 2 characters on map. Seems I have to re-create the message in sync service because the original message probably got 'muddied' by being sent as a SEGSEvent to the sync service, and then casted back to a Character/PlayerUpdateMessage.

broxen:
Works great!

screenshot from 2018-08-07 19-12-18

broxen:
@nemerle I'll leave this to you to merge when you've had a chance to review.

Projects/CoX/Common/Servers/ClientManager.h
```diff
@@ -63,6 +63,11 @@ mutable std::mutex m_store_mutex;
         }
 
 public:
+        vClients get_active_sessions()
+        {
+            return m_active_sessions;
+        }
```
nemerle: If the users of this function do not need to modify the active sessions ( the result is only used to find things ), I'd prefer us to use this form here:
```cpp
const vClients &get_active_sessions() const { return m_active_sessions; }
```

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2029,4 +2028,84 @@ void MapInstance::on_interact_with(InteractWithEntity *ev)
...
 }
 
+void MapInstance::on_update_entities()
+{
+    std::vector<MapClientSession *> active_sessions = m_session_store.get_active_sessions();
+
+    // all active sessions are for player, so we don't need to verify if db_id != 0
+    for (size_t i = 0; i < active_sessions.size(); ++i)
+    {
+        Entity *e = active_sessions[i]->m_ent;
```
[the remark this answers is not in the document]
eos428: this is my huge preference for normal for-loops at work here :P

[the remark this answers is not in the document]
nemerle: Just a side-note, for performance work, compilers can have trouble optimizing plain for loops if the condition part has a function call in it ( like `size()` call above )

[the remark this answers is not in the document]
eos428: I see... I'll keep that in mind in the future. Thanks for the heads-up.
On another note, I just made the changes as advised. Should be okay now.

Projects/CoX/Common/Servers/ClientManager.h
```diff
@@ -63,6 +63,11 @@ mutable std::mutex m_store_mutex;
         }
 
 public:
+        const vClients get_active_sessions()
+        {
+            return m_active_sessions;
```
nemerle: `const vClients &` plz,  this still will return a copy of the vector :)

eos428: Oops I forgot to look at the ampersand, updated again

PR #501, opened 2018-07-30
Making PasswordHasher a part of SEGS_Components

### LINKED ISSUE #346: Move PasswordHasher from standalone lib to a part of SEGS_Components
Since PasswordHasher does not depend on any other part of CoX subproject, it could and should be moved to SEGS_Components static library.
The task should be a set of simple git operations (`git mv`, `git rm`), updates to CMakeLists.txt to reflect the changes, and possible modifying a few `#include` statements

eos428:
Closes #346 .

Deleted PasswordHasher folder in Cox/Common and put it in Components folder as a part of SEGS_Components. Things that reference this library but include it as "PasswordHasher/PasswordHasher.h" are modified to reflect this change.


Components/PasswordHasher.cpp
```diff
@@ -5,13 +5,7 @@
  * This software is licensed under the terms of the 3-clause BSD License. See LICENSE.md for details.
  */
 
-/*!
- * @addtogroup PasswordHasher Projects/CoX/Common/PasswordHasher
- * @{
- */
-
```
nemerle: The doxygen comment should not be removed, but modified to match the file's current placement :)

eos428: Done 👍

Issue #499, opened 2018-07-30
Need to remove mapfolder prefix from MapName in database


broxen:
### Issue Summary

Create a new character. Log out and Shutdown Server.
Open Database, character map in entity_data includes map prefix from settings.cfg.

This creates two issues. 
- One, if the map prefix is changed, the user is defaulted to Outbreak (because their map is not found)
- Two, I'm getting errors like this multiple times at the charsel screen. 
```
Warning : No matching map name in g_defined_map_datas to sent map name. Returning Outbreak's display map name as default...
```

### Proposed Fix
Need to remove prefix from MapName, and add it when needed for map_path.

eos428:
For No.2, it seems that character select screen wants to find the map data for all 8 characters in the slots, even if the character is not yet created.

I think this issue is reliant on the results of #494 - if we use MapName, then this issue is prevalent. If we use MapIdx to get map data, then this won't be necessary.

broxen:
Sounds right to me. Good call

broxen:
Closed by #503

PR #503, opened 2018-08-01
Use map idx instead of map name to check for a character's map data

### LINKED ISSUE #494: Server stores mapname and map_idx. We probably don't need both
Now that we have `g_defined_map_datas` we probably don't need to store both mapname and mapidx in the database. We can store one and lookup the other as needed.

I believe we communicate with the client primarily via mapidx, so I would say maybe use that? Also, then we're not storing the map-prefix (`DefaultMapInstances`) in the database. Which is good, in the event that a server operator changes their map prefix at some point after character's have been moving around the world

eos428:
Closes #494 and #499.

This renders the functions to look up map data based on map names unused, but I'd keep it there just in case we need them. 

Also, since an uninitialized integer is 0 because it cannot be null (besides, I made sure I initialized entitydata's m_map_idx to 0 when it's created), the warnings for not finding map data in charsel screen shouldn't be there anymore (The case is the charsel screen asks for map datas for all characters in the slot, even if the characters are not yet created). They will return Outbreak, but not because it is defaulted to that, but because g_defined_map_datas[0] is indeed Outbreak.


broxen:
This looks good and works as expected.

I do appear to be getting a regression in Team-ups where the client fails to load team members correctly in the team list. Something to do with either the db_id sent, or the map name() here in EntityResponse.cpp:
```
        QString member_mapname  = tm_ent->m_client->m_current_map->name();
        bool tm_on_same_map     = true;

        if(member_mapname != src.m_client->m_current_map->name())
            tm_on_same_map = false;
```

**I'm unable t replicate on master.** That said, I'm also not clear how your changes impact this, sooooo.... I'm comfortable merging your PR without a fix, but let this serve as a reminder to open a new issue to track the regression.

Projects/CoX/Common/GameData/entitydata_definitions.h
```diff
@@ -13,12 +13,11 @@
 
 struct EntityData
 {
-static const constexpr  uint32_t    class_version       = 4;    // v4: adds m_current_map
+static const constexpr  uint32_t    class_version       = 5;    // v4: removes m_current_map
```
dracc: Comment version number incorrect?

Projects/CoX/Servers/GameServer/GameHandler.cpp
```diff
@@ -367,14 +367,8 @@ void GameHandler::on_map_req(MapServerAddrRequest *ev)
...
-    QString map_path = ed.m_current_map;
-
-    if (!map_path.isEmpty())
-        map_path = getMapPath(ed).toLower();
-    else
-    {
-        map_path = getMapPath(ev->m_mapnumber).toLower();
-    }
+    // will never be empty because we look based on m_map_idx, and integers cannot be null
+    QString map_path = getMapPath(ed).toLower();
```
dracc: Please consider using `ed.m_map_idx` instead, as it would use a much less complex overloaded function.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -433,7 +435,7 @@ const QString getMapPath(const EntityData &ed)
 {
     for (const auto &map_data : g_defined_map_datas)
     {
-        if (ed.m_current_map.contains(map_data.m_map_name, Qt::CaseInsensitive))
+        if (ed.m_map_idx == map_data.m_map_idx)
```
dracc: This one could as well just be `return getMapPath(ed.m_map_idx);` .

broxen: This is a good recommendation, I'll clean it up in my team-fix branch.

You rock dracc!

Issue #413, opened 2018-05-13
SEGS Admin future features. Ban, Delete, Edit Accounts and more!


broxen:
## Summary
Now that #408 is merged, let's track some future features for the SEGS Admin tool (very low priority) that may require additional infrastructure:

Builds off #404 (closed) and #33 

## Proposed Features
- [x] Clean up UI (see proposed layout below)
- [x] Add player fade-in to settings.cfg creation and (proposed) settings dialog
- [x] Merge Start and Stop Server buttons into a single dual-state button/toggle and Relabel Start/Stop button to "Start Server" or "Stop Server" Dependent upon state.
- [x] Put Settings options into their own class and Dialog (subclass of QDialog) with a save button.
- [x] Add monochromatic symbolic icons to buttons to provide visual clarity and GUI-goodness to UX
- [x] Provide means for updating server releases
- [x] Maybe a simple MOTD editor to edit scripts/motd.smlx

### Features to consider
- Use https://github.com/alex-spataru/QSimpleUpdater framework to serve all our update needs

### Abandoned or Moved Features
- Moved: ~~Provide button to launch game client, perhaps with options for fullscreen/windowed~~ this belongs in SLAV #40 
- Moved: ~Update database schemas after changes (new functionality via dbtool)~ - follow this in Add ability to update databases through dbtool #447

#### *Moved RPC related features to #436*
- Moved: ~Show server stats (version, active connections, uptime, etc)~
- Moved: ~Ban player accounts (new functionality) #33~
- Moved: ~Delete Accounts and associated characters #33~
- Moved: ~Reset player passwords (how to handle this?) #33~
- Moved: ~Rename characters (TheIncredibleHulk needs a new name but is offline) #33~
- Moved: ~Ban based upon IP (for the jerk hackers) #33~

## UI Mockups
![segsadmin](https://user-images.githubusercontent.com/10159505/40417893-b8f8d65e-5e35-11e8-87c0-9eea91c1ef0f.png)

broxen:
For some of these database edits (bans, name edits, reset passwords, etc). How to handle while the server is running?

nemerle:
This is the work for the jsonrpc, methinks

broxen:
Cleaned this up and separated out the RPC related tasks

miles200:
@broxen Everything on this issue is now implemented in SEGSAdmin. Did you want to keep open for future stuff or should we close this one off and create a new one if anything else comes up?

broxen:
[...]

PR #508, opened 2018-08-03
- Add ability to check for updates on start up utilising the github API

miles200:
Contributes towards #413 

- Add github API integration to check for new releases. On new release available alerts and provides detail and download URL to user.
- Add about page to add license information
- UI tweaks, icons added, new splash screen (to be replaced at later stage with new version of splash screen)


miles200:
This version of SEGSAdmin has a dependency on SSL libraries being in the same folder as SEGSAdmin.exe 

- ```libeay32.dll```
- ```ssleay32.dll```

I'm still unsure how / where to document this so we ensure we include these in the next release for users. I could probably document in the readme.MD as ldilley suggested so anyone compiling from the master / dev repo knows these are required.

ldilley:
I've reviewed the diffs and everything seems acceptable. I am building this instance of SEGSAdmin from the PR now and will validate some functionality prior to merging.

ldilley:
The splash screen is a nice touch and the "About" window is perfect in regards to referencing licenses for 3rd-party software used. The update functionality will also come in handy. I've validated that everything is working as intended after dropping both SSL DLLs into the run path of the executable. I'll submit an issue as a reminder to update the guide(s) in regards to the crypto libraries so this does not fall through the cracks.

I must say, this program is truly turning out to be amazing. It has a very professional look to it and is evolving extraordinarily well. Excellent job, @miles200!

PR #516, opened 2018-08-08
Friend list and online status working

### LINKED ISSUE #470: Friends online status doesn't work properly.
Friends online status doesn't work properly.

Currently we save `m_online_status` and `m_mapname` for Friends to the database. Instead we should do a lookup of those values prior to sending to the client in FriendsListUpdate.

Around line 36 of FriendsListUpdate.h
```
            // TODO: Lookup online status, implement: isFriendOnline()
            //bool is_online = isFriendOnline(m_list->m_friends[i].fr_db_id);

            bs.StoreBits(1,m_list->m_friends[i].m_online_status);
```

Classic behavior is that offline friends show "OFFLINE" for zone.

This same logic will probably apply to TeamList too, so we may want to generalize the isOnline method for use anywhere.

supermandrew:
Closes #470.

- Will later need to be adjusted for when characters can zone.
- 
-


Projects/CoX/Servers/MapServer/Events/FriendHandler.cpp
```diff
@@ -0,0 +1,162 @@
...
+        s_friend_map[m_friendslist.m_friends[i].m_db_id].insert(m_char_id);
+    }
+
+    //Iterate through the friends list and update online status accordingly
+    for(Friend& f : m_friendslist.m_friends)
+    {
+        f.m_online_status = is_online(f.m_db_id);
+        f.m_mapname = getFriendDisplayMapName(f);
+        if(!f.m_online_status){
+            //f.m_map_idx = -1;
```
broxen: map_idx is uint, but also, since this is commented out, maybe we should remove the if statement?

Projects/CoX/Servers/MapServer/Events/FriendHandler.cpp
```diff
@@ -0,0 +1,162 @@
...
+                s_map_info_map[msg->m_data.m_char_id].instance_id);
+    //Iterate over map and update friends list of all people who have added this character
+    for(auto const& val : s_friend_map[msg->m_data.m_char_id])
+    {
+        //We need to notify all the people who added this player (if they're online)
+        if(is_online(val)){
+            inst_tgt->putq(new GetPlayerFriendsRequest({s_map_info_map[val].session_token},msg->session_token(),this));
+        }
+    }
+    s_map_info_map.erase(msg->m_data.m_char_id);
```
broxen: We have two "char_ids" on the server, a player's db_id and their m_idx, which is their Entity ID on the MapServer. 

I recommend changing your use of char_id throughout this PR to clarify that it's the characters db_id and not some other identifier.

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -570,6 +585,36 @@ void MapInstance::on_expect_client( ExpectMapClientRequest *ev )
         ->putq(new ExpectMapClientResponse({cookie, 0, m_addresses.m_location_addr}, ev->session_token()));
 }
 
+void MapInstance::on_update_friendslist(SendFriendListMessage *ev)
+{
+    MapClientSession &map_session(m_session_store.session_from_token(ev->m_data.m_session_token));
+    Entity * e = map_session.m_ent;
+    e->m_db_id = e->m_char->m_db_id;
```
broxen: These values should, in theory, already be equal. Why reset e->m_db_id in your friendslist update?

This is outside the scope of FriendsHandler and I recommend removing this. If we ever change one of these members variables elsewhere, having this here could introduce bugs.

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -570,6 +585,36 @@ void MapInstance::on_expect_client( ExpectMapClientRequest *ev )
...
+
+    QString notify_msg = "Your friend " + e->m_char->getName() + " is now online.";
+    sendInfoMessage(MessageChannel::SERVER,notify_msg,&notify_session);
+}
+
+void MapInstance::on_get_player_friends(GetPlayerFriendsRequest *ev)
+{
+    MapClientSession &map_session(m_session_store.session_from_token(ev->m_data.m_session_token));
+    Entity * e = map_session.m_ent;
+    e->m_db_id = e->m_char->m_db_id;
```
broxen: Here again.

Projects/CoX/Servers/MapServer/Events/FriendHandler.cpp
supermandrew: Sure thing, I'll likely rename it to m_char_db_id then.

Projects/CoX/Servers/MapServer/MapInstance.cpp
supermandrew: The only reason I had this in is because I thought it was necessary code when grabbing a character entity.  If you look at on_entity_response inside MapInstance, you'll see that the same thing happens.

    MapClientSession &map_session(m_session_store.session_from_event(ev));
    m_session_store.locked_unmark_session_for_reaping(&map_session);
    Entity * e = map_session.m_ent;
    e->m_db_id              = e->m_char->m_db_id;

But if that's not needed at all, I'd be happy to remove it.  I agree with you, don't want to introduce bugs.

broxen: Yeah, it belongs in on_entity_response only. I would remove it from your friends-related funcs.

supermandrew:
Changes made as requested broxen! :)

supermandrew:
Changes made as requested broxen! :)

Projects/CoX/Servers/GameServer/GameServer.cpp
```diff
@@ -89,6 +90,11 @@ GameServer::GameServer(int id) : d(new PrivateData)
     d->m_handler->start();
     d->m_id = id;
     HandlerLocator::setGame_Handler(d->m_id,d->m_handler);
+
+    qDebug() << "Creating FHandler";
+    FriendHandler *friendHandler = new FriendHandler();
+    friendHandler->set_game_server_id(id);
+    friendHandler->activate();
 }
```
nemerle: Consider storing friendHandler inside the PrivateData class, to allow for easier cleanup on exit.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -578,7 +578,7 @@ void sendFloatingNumbers(Entity *src, uint32_t tgt_idx, int32_t amount)
     src->m_client->addCommandToSendNextUpdate(std::unique_ptr<FloatingDamage>(new FloatingDamage(src->m_idx, tgt_idx, amount)));
 }
 
-void sendFriendsListUpdate(Entity *src, FriendsList *friends_list)
+void sendFriendsListUpdate(Entity *src, FriendsList friends_list)
 {
```
nemerle: FriendsList is not a very small class, ( in termes of bytes-needed-per-instance ),  so here it passing it as `const FriendsList &`   makes more sense.

Projects/CoX/Servers/MapServer/Events/FriendHandler.cpp
```diff
@@ -0,0 +1,159 @@
...
+            break;
+        default:
+            break;
+    }
+}
+
+void FriendHandler::on_player_friends(GetPlayerFriendsResponse* ev)
+{
+    uint32_t &m_char_db_id = ev->m_data.m_char_db_id;
+    FriendsList m_friendslist = ev->m_data.m_friendslist;
```
nemerle: * no need to make `m_char_db_id` a reference to an event's internal value 
* local variables should not have `m_` prefix

Projects/CoX/Servers/MapServer/Events/FriendHandler.cpp
```diff
@@ -0,0 +1,159 @@
...
+     * but won't do any harm.  Inefficient though.
+     *
+     * Iterate over the friends list, and put each friend id as a key,
+     * and push the current character id into the set value.
+     * We don't have to worry about duplicates because we're using a set.
+     */
+    for(uint i=0; i<m_friendslist.m_friends.size(); i++){
+        s_friend_map[m_friendslist.m_friends[i].m_db_id].insert(m_char_db_id);
+    }
+
```
nemerle: maybe use range-for here as well:
```cpp
for(const auto &friend : m_friendslist.m_friends)
    s_friend_map[friend.m_db_id].insert(m_char_db_id);
```

Projects/CoX/Servers/MapServer/Events/FriendHandler.cpp
```diff
@@ -0,0 +1,159 @@
...
+                s_map_info_map[m_char_db_id].server_id, s_map_info_map[m_char_db_id].instance_id);
+    tgt->putq(new SendFriendListMessage({s_map_info_map[m_char_db_id].session_token,
+                                        m_friendslist}));
+}
+
+void FriendHandler::on_client_connected(ClientConnectedMessage *msg)
+{
+    //A player has connected, we need to notify all the people that have added this character as a friend
+    uint32_t &m_char_db_id = msg->m_data.m_char_db_id;
+
```
nemerle: here as well there's no need for reference  + `'m_` prefix is reserved for class's member variables

Projects/CoX/Servers/MapServer/Events/FriendHandler.cpp
```diff
@@ -0,0 +1,159 @@
...
+    EventProcessor *inst_tgt = HandlerLocator::getMapInstance_Handler(
+                s_map_info_map[m_char_db_id].server_id, s_map_info_map[m_char_db_id].instance_id);
+    //Iterate over map and update friends list of all people who have added this character
+    for(auto const& val : s_friend_map[m_char_db_id])
+    {
+        //We need to notify all the people who added this player (if they're online)
+        if(is_online(val)){
+            uint32_t friend_id = val;
+            inst_tgt->putq(new GetPlayerFriendsRequest({s_map_info_map[val].session_token},msg->session_token(),this));
+            inst_tgt->putq(new SendNotifyFriendMessage({s_map_info_map[m_char_db_id].session_token,
```
nemerle: The logic here's a bit strange:
you send 2 request to MapInstance here, one of which is an attempt to `GetPlayerFriendsRequest` ?
1. was `GetPlayerFriendsRequest` not an responsibility of DB ?
2. sending a `request`, and not continuing the process when the response is received means that second message cannot make use of the results of the `request`

supermandrew:
Okay, after talking to nemerle, we decided to take the following course of action:

- change an Entity's FriendsList to just be a set of DB ids (character IDs)
- only request FriendsList when a player logs in
- whenever a change should be made to a player's friends list, construct the FriendsList structure that sendFriendsListUpdate is expecting, send it to MapInstance to send to sendFriendsListUpdate.  

Store a map of Friend structs based on db_id as key, and keep those in FriendHandler.  Use push_back Friend structs from this map in order to build the FriendsList needed.

supermandrew:
Should be ready for review again.  Not sure how to do the missing task nem suggested (moving the handler inside the PrivateData class), but doesn't sound like it'd be too hard.

broxen:
@supermandrew  can you resolve merge conflicts here? I'll test and merge afterward.

supermandrew:
Created a new Segs-470 branch since I basically worked from scratch on changing dev and incorporating my changes.

PR #517, opened 2018-08-14
#509 - Add OpenSSL DDLs to source tree.

### LINKED ISSUE #509: Update relevant guides for SEGSAdmin SSL DLL requirement
With the merger of PR #508, `SEGSAdmin.exe` now makes use of crypto functions defined in the following dynamic link libraries provided by OpenSSL:
* libeay32.dll
* ssleay32.dll

These libraries must be available from the `SEGSAdmin.exe` path at runtime. They can be installed per https://wiki.openssl.org/index.php/Binaries from https://slproweb.com/products/Win32OpenSSL.html. Please select a 1.0.x version considering the SSLeay moniker was renamed in 1.1.x. Both DLLs can simply be copied to the `SEGSAdmin.exe` path after the OpenSSL 1.0.x installer completes.

Any relevant build and install guides should be updated to reflect the above. Lastly, when producing any SEGSAdmin binary releases targeting the Windows platform, it is permissible (licensing requirements are already fulfilled) and advised to include these DLLs in the distribution. This would save end users some work.

deciton666:
Closes #509

- Added OpenSSL DLLs to source tree.
- Build copies the OpenSSL DLLs to the out directory (windows only).

Is this how we want to solve this issue?


nemerle:
While the idea does have merit, most of the time I'm against storing dlls in the git repository

I think the way to solve this would be adding the required dll to the released binaries instead.

deciton666:
I agree. That is a better solution.

PR #518, opened 2018-08-14
SEGSAdmin - Add splash screen and OpenSSL error handling

### LINKED ISSUE #510: Update SEGSAdmin version number to 0.6.0 in line with release
This needs to be done prior to 0.6.0 release so SEGSAdmin doesn’t alert a new version for downloads after release

miles200:
Closes #510 

- Add error handling and user messaging if OpenSSL dll's are missing
- Add updated splash screen from @broxen 
- Moved all resources to it's own folder `SEGSAdmin/Resources`
- Now checking version number from version.h

Something has gone weird with my squash, it's including an appveyor change, hopefully this doesn't cause an issue let me know and I can attempt to sort


broxen:
Appveyor change is ok.  I thought I already pushed that change. Not sure :shrug:

Projects/CoX/Utilities/SEGSAdmin/SEGSAdminTool.cpp
```diff
@@ -120,12 +120,12 @@ void SEGSAdminTool::check_data_and_dir(QString maps_dir) // Checks for data sub
     }
     if(all_data_dirs_exist && all_maps_exist)
     {
-        ui->icon_status_data->setText("<html><head/><body><p><img src=':/icons/check.svg'/></p></body></html>");
+        ui->icon_status_data->setText("<html><head/><body><p><img src=':/icons/Resources/check.svg'/></p></body></html>");
```
broxen: For these icons, can we use colorized fonts from the package we sourced this from instead of HTML with svg? Seems like it would be a cleaner and easily scalable solution?

Something akin to:
```
if (QFontDatabase::addApplicationFont(":/FontAwesome.otf") < 0)
    qWarning() << "FontAwesome cannot be loaded !";

QFont font;
font.setFamily("FontAwesome");
font.setPixelSize(32);

ui->pushButton->setFont(font);
ui->pushButton->setText("\uf083");
```

Projects/CoX/Utilities/SEGSAdmin/resources.qrc
```diff
@@ -1,23 +1,23 @@
...
-        <file>file-plus.svg</file>
-        <file>settings.svg</file>
-        <file>sliders.svg</file>
-        <file>users.svg</file>
-        <file>play.svg</file>
-        <file>square.svg</file>
-        <file>download.svg</file>
-        <file>check.svg</file>
-        <file>alert-triangle.svg</file>
+        <file>Resources/segs.bmp</file>
```
broxen: I'd rather see this bmp as SVG or PNG.

broxen: In fact, what is this used for? The app icon? Perhaps it's best if we made it SVG per new standards.

miles200: @broxen to provide SVG

broxen:
Will test and merge once app icon is in. Please let me know if there is anything else outstanding on this one.

miles200:
I think that's all @broxen .. I'll get that icon pushed up tonight so you can test. 

Thanks!!

broxen:
You're the best!

Projects/CoX/Utilities/SEGSAdmin/NetworkManager.cpp
```diff
@@ -29,11 +30,28 @@ NetworkManager::NetworkManager()
...
+        "ssleay32.dll",
+        "libeay32.dll",
+    };
+
+    bool ssl_files_exist = true;
+    for(const QString &ssl_file : open_ssl_dll)
+    {
+        ssl_files_exist &= QFile(ssl_file).exists();
+    }
+    if(!ssl_files_exist)
```
broxen: Hrmm.. an odd issue with this: A user that already has OpenSSL installed through their package manager will always return False here, because the SSL files aren't in the SEGS directory.

Perhaps there is a way to use the connect() error for those without SSL for the logic here?

miles200: On thinking about this, I developed the GitHub API stuff whilst away from home on my laptop using linux, and the error never occurred, it only came up when I was using a windows machine... Is OpenSSL required for linux / Mac is the question in my mind? Is it just a funky windows thing?

I could do something like: 

 `#ifdef WIN32` check for the SSL dll checks and if not on windows just skip straight to the network manager function... 

Thoughts?

nemerle: It's hard to imagine any widely used linux distro without OpenSSL shipped with it. No idea about Macs though.

miles200: On conversation with @dracc and @nemerle in Discord, the common thinking is that most linux distorts would have OpenSSL installed by default, so I may not need to check for linux systems. I'll do a test on my Mac to see if the API stuff is working correctly without the openSSL check in the code, if so I'll go ahead and add the `#ifdef WIN32` clause

Projects/CoX/Utilities/SEGSAdmin/NetworkManager.cpp
```diff
@@ -53,6 +71,8 @@ void NetworkManager::store_latest_releases()
     }
     qDebug()<<"Api retrieval and stored success";
     qDebug()<<g_segs_release_info[0].github_url;
-    emit releasesReadyToRead();
+    emit releasesReadyToRead(nullptr);
```
miles200: @broxen Just one question I had, is this acceptable to pass 'nullptr' here, I don't need to pass anything back to this method but it requires a string for another scenario (in an error it passes the error back)... Is there a way to make the argument 'optional'?

broxen: You have two options for "optional" arguments

1- You can overload the function so that it doesn't require a variable: https://www.programiz.com/cpp-programming/function-overloading
2- Or you can provide a default value for the argument: https://www.programiz.com/cpp-programming/default-argument

I'd recommend one of those over nullptr

I think this part is particularly important: 
> No matter how you use default arguments, a function should always be written so that it serves only one purpose.
> If your function does more than one thing or the logic seems too complicated, you can use Function overloading to separate the logic better.

broxen:
This looks great. Needs testing and merge

PR #522, opened 2018-08-18
Allow Character to AFK and Auto-Logout (Optional)

### LINKED ISSUE #75: Server doesn't AFK (or disconnect?) idle clients
I accidentally left the client running on my server all night and day, totaling over 18hrs. This shows how rock-solid the server is! However, it probably isn't desired behavior for a live server? Should the server disconnect idle clients?

Regardless, the server should detect when a client goes idle for a period of time, and set the character to [/afk status](https://paragonwiki.com/wiki/Afk_(Slash_Command))

eos428:
Closes #75 .

- Allow players to enter AFK once a certain time period has passed while being idle
- Optionally, allow players to enter auto logout period once another time period has passed while being idle
- Force the player to logout if they didn't do anything during the auto-logout 'grace period' (which is defaulted to 30 secs)

TODO:
- Complete MapServerData::read_settings() to allow MapServerData to read from settings.cfg to put the time thresholds based on a server admin's settings, and whether auto logout is used or not. (Should be this a SegsAdmin addition instead?)
- Display a message to player that they will be auto logged out in 2 mins (configurable) after that m_time_to_logout_msg threshold is passed and the server uses auto-logouts
- Force client to logout in 30 seconds (configurable?) if player is on auto logout period and that they didn't do anything during the 'grace period'
- If the server uses auto logout - Once 18 minutes have passed, give the message to player that they will be logged out in 2 minutes, if no action is done within that time period. And should the player not do anything, then logout the player in 30 secs in which the player cannot reverse the logout.

And obviously when testing, we'd need to slash the time thresholds unless you want to wait the whole 20 mins to see if it works, haha.


broxen:
Will wait for further progress prior to testing.

Thank you for tackling this @eos428 !

eos428:
Sure np.

I think I'll need to ask nem on how we should structure this...  but here's what I did and encountered so far.
- To get all player characters, we'd use getActiveSessions() as we did on Sync Service and then increase m_idle_time based on deltaTime.
- The game should send a message to the player that they are AFK'ed and etc. We'd need the session instead of the entity - so we can't do everything in World::update (which also updates NPC entities, and there was a reason I updated the Sync Service)
- Is there any way for the server to force the client to logout in 30 seconds in which the client cannot reverse the logout decision? (Using opcode?)

broxen:
Adding this to v0.7.0 but we'll gladly merge it in if ready sooner.

Thanks for all your hard work @eos428!!

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -502,8 +505,15 @@ void MapInstance::on_client_quit(ClientQuit*ev)
     MapClientSession &session(m_session_store.session_from_event(ev));
     // process client removal -> sending delete event to all clients etc.
     assert(session.m_ent);
+
     if(ev->abort_disconnect)
-        abortLogout(session.m_ent);
+    {
+        // forbid the player from aborting logout,
+        // if logout is called automatically (through a loooong period of AFK)
```
broxen: Shouldn't a player still be able to abort logout here?

What if a player returns from afk as logout is happening?

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2204,6 +2229,67 @@ void MapInstance::on_recv_new_power(RecvNewPower *ev)
...
+    for (const auto &sess : active_sessions)
+    {
+        Entity *e = sess->m_ent;
+        CharacterData* cd = &e->m_char->m_char_data;
+
+        if (e->m_has_input_on_timeframe == false)
+            cd->m_idle_time += afk_update_interval.sec();
+        else
+        {
+            sendInfoMessage(MessageChannel::EMOTE, "Receiving input from player", sess);
```
broxen: I assume this is for testing. We should probably change to a qCDebug(logInput) instead of sendInfoMessage()

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2204,6 +2229,67 @@ void MapInstance::on_recv_new_power(RecvNewPower *ev)
...
+        }
+
+        const MapServerData &data(g_GlobalMapServer->runtimeData());
+        QString msg;
+
+        if (cd->m_idle_time >= data.m_time_to_afk && !cd->m_afk)
+        {
+            toggleAFK(* e->m_char, true, "Auto AFK");
+            msg = QString("You are AFKed after %1 seconds of inactivity").arg(data.m_time_to_afk);
+            sendInfoMessage(MessageChannel::EMOTE, msg, sess);
```
broxen: We might change the channel here to DEBUG instead. It will be easier to find by the player later when they return from AFK, and is more thematically appropriate.

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2204,6 +2229,67 @@ void MapInstance::on_recv_new_power(RecvNewPower *ev)
...
+        if (data.m_uses_auto_logout && cd->m_idle_time >= data.m_time_to_logout_msg)
+        {
+            // give message that character will be auto logged out in 2 mins
+            // player must not be on task force and is not on mission map for this to happen
+            if (!cd->m_is_on_task_force && !isEntityOnMissionMap(e->m_entity_data) && !cd->m_is_on_auto_logout)
+            {
+                cd->m_is_on_auto_logout = true;
+                msg = QString("You have been inactive for %1 seconds. You will automatically ").arg(data.m_time_to_logout_msg) +
+                      QString("be logged out if you stay idle for %1 seconds").arg(data.m_time_to_auto_logout);
+                sendInfoMessage(MessageChannel::EMOTE, msg, sess);
```
broxen: Same here. DEBUG channel is probably better.

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2204,6 +2229,67 @@ void MapInstance::on_recv_new_power(RecvNewPower *ev)
...
+            }
+        }
+
+        if (cd->m_is_on_auto_logout && cd->m_idle_time >=
+                data.m_time_to_logout_msg + data.m_time_to_auto_logout
+                && !e->m_is_logging_out)
+        {
+            e->beginLogout(30);
+            msg = "You have been inactive for too long. Beginning auto-logout process...";
+            sendInfoMessage(MessageChannel::EMOTE, msg, sess);
```
broxen: Here as well

Projects/CoX/Servers/MapServer/MapServerData.cpp
```diff
@@ -388,6 +390,18 @@ bool MapServerData::read_npcs(const QString &directory_path)
     return true;
 }
 
+// should read from settings.cfg
+bool MapServerData::read_settings(const QString &directory_path)
+{
+    qDebug() << "Loading data from settings:";
```
broxen: This can make a great `good-first-issue` for the issue queue. I'll add it after this merges.

For now, let's change this debug text since we're not actually loading those values from settings yet.

Projects/CoX/Servers/MapServer/MapInstance.cpp
eos428: If simply on AFK, the character can still logout fine.

It's when the character has passed 18 + 2 minutes that they won't be able to do anything anymore. But that's the official implementation of it. I suppose we could customize that and still allow to abort logout regardless on that 30 secs

broxen: Ahh that makes sense. I say leave it like this

broxen:
This looks good to me. Just need someone to test.

Otherwise I'll test early next week.

Great work eoS

broxen:
Hi eoS, when you have time can you please resolve merge conflicts? Thank you!

broxen:
One issue I found with this.

Manually setting afk `/afk Test Message` has stopped working. Looks like you're flagging it off somewhere.

And just a reminder to set the timers to their normal values prior to merge.
> Auto AFK: If there is no keyboard activity in 5 minutes, your character will automatically be set to AFK (“Away From Keyboard”).
> 
> Auto Logout: In order to improve lag and framerate for our players, we have instituted an Auto Logout feature for players who are not actively playing the game. After 18 minutes of inactivity you will hear a chime and receive a message that you will be auto logged out in 2 minutes. If you have not given any inputs to your character in those 2 minutes, you will hear a klaxon and the 30 second logout timer will begin. If you resume activity during those 20 minutes then the Auto Logout will be reset.
> 
> You will NOT Auto Logout under the following circumstances:
> - You are on a Mission Map.
>- You are on a Task Force.
```

**Comment by eos428** [comment]
Can you elaborate this one? Are you supposed to never go AFK when on a team, or this PR makes it impossible to AFK while on a team?

**Comment by broxen** [comment]
This PR makes it impossible to go afk while on a team.

Players should still be able to afk while teaming.

**Comment by eos428** [comment]
/afk does not work anymore since typing is considered an input... The actual toggle works, but then on_afk_update sets it off again, since it detected an input.

File: Projects/CoX/Servers/MapServer/MapInstance.cpp [hunk 7]
```diff
@@ -2301,7 +2301,7 @@ void MapInstance::on_afk_update()
         if (cd->m_idle_time >= data.m_time_to_afk && !cd->m_afk)
         {
             toggleAFK(* e->m_char, true, "Auto AFK");
-            msg = QString("You are AFKed after %1 seconds of inactivity").arg(data.m_time_to_afk);
+            msg = QString("You are AFKed after %1 seconds of inactivity. Player: %2").arg(data.m_time_to_afk);
```
broxen: This QString has no second argument for `%2` to work

eos428:
The fix should be easy (in fact this is incidental behaviour) but i thought there was a discussion in discord in which we should not allow afk while targeting?

broxen:
I must have missed that one, or forgot about it. But it's true that selecting a new target should disable AFK, but simply having a target (previously selected) should have no effect on afk status.

eos428:
AFKing in regards to targeting (remove AFK when selecting/switching target, do AFK if doing nothing with a target selected) should work as you want now

PR #531, opened 2018-08-31
Dead No Gurney changes.

### LINKED ISSUE #529: Dead No Gurney Message
From #519 
Will need to change how health is modified so that it can trigger events.
From Discord - keep in mind possibility for scripts to register to accessor functions in the future.

Magentah:
Closes #529  .

- Health changes are because I initially thought it would be triggered by health reaching 0, but it's triggered by a client event. I can undo them if needed.
- Added client event AwaitingDeadNoGurney and server event DeadNoGurney
- Added test command, which just calls the MapInstance event handler.


Magentah:
I've never actually used C++ before outside of small console apps, so there's probably some issues with what I wrote..

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -526,16 +526,16 @@ void setCombatLevel(Character &c, uint32_t val)
...
-}
-
 void setEnd(Character &c, float val)
 {
     c.m_char_data.m_current_attribs.m_Endurance = std::max(0.0f, std::min(val,c.m_max_attribs.m_Endurance));
 }
 
+void modifyHealth(Entity *tgt, float amount, bool send_events)
+{    
+    tgt->m_char->setHealth(amount);
```
dracc: Afaict, you're not modifying health as much as you're setting it to a value. But you use this function (`modifyHealth`) as if it did `currentHealth += amount`.

Magentah: Ah good catch, missed that.

Projects/CoX/Servers/MapServer/Events/DeadNoGurney.h
```diff
@@ -0,0 +1,26 @@
...
+
+#pragma once
+#include "GameCommandList.h"
+
+#include "MapEvents.h"
+#include "MapLink.h"
+
+#include <QtCore/QString>
+
+class DeadNoGurney : public GameCommand
```
broxen: Can you please add unit test comments to this class and it's members. You can see an example in PowerSystemEvents.h

`// [[ev_def:type]]` above the class, and
`// [[ev_def:field]]` above each member variable (you have none)

Projects/CoX/Servers/MapServer/MapEvents.h
```diff
@@ -259,6 +259,20 @@ class EnterDoor final : public MapLinkEvent
     }
 };
 
+class AwaitingDeadNoGurney final : public MapLinkEvent
```
broxen: Same here please `// [[ev_def:type]]`

Magentah:
Added the comments.

Projects/CoX/Servers/MapServer/MapEvents.h
broxen: Oops. Can we add this ev_def comment here as well? Thanks!

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2204,6 +2207,17 @@ void MapInstance::on_recv_new_power(RecvNewPower *ev)
     addPower(session.m_ent->m_char->m_char_data, ev->ppool);
 }
 
+void MapInstance::on_awaiting_dead_no_gurney(AwaitingDeadNoGurney *ev)
+{
+    MapClientSession &session(m_session_store.session_from_event(ev));
+    session.m_ent->m_client->addCommandToSendNextUpdate(std::unique_ptr<DeadNoGurney>(new DeadNoGurney()));
+}
+
+void MapInstance::on_awaiting_dead_no_gurney_test(AwaitingDeadNoGurney *ev, MapClientSession &session)
```
broxen: I'm wondering if we should move this method to DataHelpers the way we have `sendFloatingInfo()`

Then, in a possible Lua future we can expose that to the scripts without needing MapInstances and MapClientSession

Magentah: `evAwaitingDeadGurney` is a client -> server event, would that still work? `evFloatingInfo` is server -> client, the same as `evDeadNoGurney`, which is already in DataHelpers.

PR #540, opened 2018-09-02
Gracefully handle more than 5col x 4row of inspirations

### LINKED ISSUE #524: Gracefully handle more than 5col x 4row of inspirations
In Powers PR #495 , vInspirations hard codes 4 x 5 into the arrays. This is done because the client will not allow more than 4 rows of 5 cols of inspirations. We need to gracefully handle erroneous input beyond 4x5.

Per discord, nem's suggestion is to convert to nested vectors, and add syntactic sugar for accessing by col x row (the order the client expects)

Something like:
```cpp
struct vInspirations {
  using storage =   std::vector<std::vector<CharacterInspirations>>;
  using iterator = storage::iterator;
  std::vector<std::vector<CharacterInspirations>> m_inspirations;
  vInspirations(size_t rows,size_t cols);
  iterator begin() { return m_inspirations.begin(); }
  iterator end() { return m_inspirations.end(); }
  std::vector<CharacterInspirations> &at(int idx) { return m_inspirations.at(idx); }
};
```

eos428:
Closes #524 .

- We can handle situations where we have more than 5x4 inspirations gracefully
- The struct can contain its own functions, which makes things more flexible
- And since this is a struct, it'll need its own serialization function, which is put in chardata_serializers

A few notes:
- My vector is one-dimensional, and individual inspirations are accessible through m_inspirations[col * row], which we call through &at
- This build fails because &at must be const to follow functions like sendInspirations(), which are const as well. But it seems we cannot bind const when we are returning a reference.

``error: binding 'const value_type {aka const CharacterInspiration}' to reference of type 'CharacterInspiration&' discards qualifiers``

Update: Managed to circumvent the error (see message of 2nd commit).


Projects/CoX/Common/NetStructures/Powers.h
```diff
@@ -58,8 +59,49 @@ static const constexpr  uint32_t    class_version = 1;
...
+    }
+
+    const size_t size()
+    {
+        return m_inspirations.size();
+    }
+
+    CharacterInspiration& at (const size_t col, const size_t row)
+    {
+        // Note: possible arrayIndexOutOfRange exception here
```
broxen: How can we avoid, or gracefully handle this possible exception?

Projects/CoX/Common/NetStructures/Powers.h
```diff
@@ -58,8 +59,49 @@ static const constexpr  uint32_t    class_version = 1;
...
+        return m_inspirations[col * row];
+    }
+
+    CharacterInspiration value(const size_t col, const size_t row) const
+    {
+        return m_inspirations[col * row];
```
broxen: My concern about this implementation is that it cannot distinguish between an inspiration at col 2 * row 3 and one at col 3 * row 2

Projects/CoX/Common/NetStructures/Powers.cpp
```diff
@@ -574,23 +574,23 @@ void addInspirationToChar(CharacterData &cd, CharacterInspiration insp)
     int max_rows = cd.m_max_insp_rows;
     int count = 0;
 
-    for(int i = 0; i < max_rows; ++i)
+    for(int i = 0; i < max_cols; ++i)
```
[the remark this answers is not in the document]
eos428: If we are going to fill from left to right (column-first) first, then we'd have to revert this change since j must refer to a column. I don't get the 2nd question, since I haven't seen it for myself. Apologies for that.

Projects/CoX/Common/NetStructures/Powers.h
broxen: Remember that players start with only 1row*3cols, then progressively get more? Could this inplementation consider an entities max cold and rows?

In your example, 3*2 could be a different idx, depending upon Character level.

From paragonwiki

### Inspiration Slots
A character starts with three inspiration slots in their Inspirations Window at level one, and gradually increases to a maximum of 20 inspiration slots at level 40. The progression is below:

Level 1: 3 slots (1 row of 3 columns)
Level 2: 4 slots (1 row of 4 columns)
Level 3: 8 slots (2 rows of 4 columns)
Level 10: 10 slots (2 rows of 5 columns)
Level 25: 15 slots (3 rows of 5 columns)
Level 40: 20 slots (4 rows of 5 columns)

Projects/CoX/Common/NetStructures/Powers.cpp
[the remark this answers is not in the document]
broxen: No Problem at all. My second question may still apply after reverting this change, due to the conversion to a flat vector.

To phrase it another way:
Using an inspiration on the first row, will automatically move an inspiration from the second row down to fill it's location in the array.

Using [col][row]:
insp[3][2] becomes insp[3][1]

This behavior is intentional, so that using the same inspiration keyboard key uses inspirations from the same column. F1 is always firing column 1 inspirations.

You can see this behavior in action on current develop branch. We simply need to test for this use case.

Projects/CoX/Common/NetStructures/Powers.h
eos428: I didn't play the game so I dunno haha... and yes, depending on character level, 3 * 2 would return a different idx. Now this depends on the client:
- Is the client satisfied with the server 1 row of 3 columns at level 1, or 
- Must all 20 slots be filled regardless of the level?

For the latter, we'd simply have to force m_cols to 5 and m_rows or 4. Which does get into hardcoding territory.

For the former, let's do this with an example of col (2) row (0) and col (1) row (1). And yeah, I just realized it should be (col * m_rows) + row.
- Level 1 and 2, col (2) row (0) will return m_inspirations[2] which is (2 * 1) + 0, col (1) row (1) returns an error since it looks for the 2nd row, which the character does not have
- Level 3 and 10: col (2) row (0) will return m_inspirations[4] which (2 * 2) + 0, col (1) row (1) returns m_inspirations[3] which is (1 * 2) + 1
- Level 25: col (2) row (0) will return m_inspirations[6], col (1) row (1) returns m_inspirations[4]
- Too lazy to calculate for lv40, but you get the gist of it

broxen: So the client takes the max rows and cold for the player and then expects that exact amount of inspiration slots (whether empty or not)

It does this because you can move the inspirations around and stack them in any column you want. So you may have an empty col 2, and insps in column 3.

eos428: We'd be doing the latter option then. We'll hardcode m_cols to 5 and m_rows to 4, then use m_inspirations.reserve(m_cols * m_rows), which will create 20 empty entries, if I'm reading the function right. The easy way would be to push_back empty powers 20 times.

But anyways, let's look to your [3][2] example again.
- Level 1 to 10: Invalid because we are looking for row 3
- Level 25: returns m_inspirations[11] which is 3 (col) * 3 (m_rows) + 2 (row)
- Level 40: returns m_inspirations[14] which is 3 (col) * 4 (m_rows) + 2 (row)

broxen: Ermm I'm getting idx 9 for level 3, and idx 11 for any lvl above 25?

broxen: Also note, based upon I think you're want the former, not latter.

Just to clarify, the client expects the exact number of slots available for your level only. So at level 1, the server must send only 3 inspiration slots. At level 25 we must send exactly 15 slots

eos428: [3][2] is column 4 and row 3.

Up to lv10 we are not supposed to take row 3.
Level 25 - Looks at 12th element
Level 40 - Looks at 15th element

broxen: Ah yes, due to 0s. But still in a flat vector I'm getting idx 13 above level 25? Because you always have the 5th column after that?

eos428: I didn't see the edit. So overall we can just use the current m_rows and m_cols implementation

broxen:
No problem. As the original issue was opened as a result of a conversation with Nem regarding the 4x5 array structure. Let's rope him in here for direction. It's possible he may not have considered these other conditions.

@nemerle can you provide any guidance here?

broxen:
I'm crashing on player load, and new character creation.

Looks like it fails on line 112 of Powers.h
screenshot from 2018-09-05 21-34-02

broxen:
Hi eoS.

I'm not able to spawn more than 4 inspirations with this PR. And moving inspriations to row 0, column 4, results in them disappearing (for good). In fact, moving inspirations around in general makes them disappear.

Steps to replicate:
- Create new char
- `/setlevel 40`
- `/addinsp luck` or `/addinsp insight` about 10 times
- Max 4 inspirations appear
- Move inspirations around to different slots.

eos428:
Something is weird... on addInspirationToChar(), a level 40 character's max_cols is 4 and max_rows is 1. Which is well, way off the mark. That number remains 4 and 1 for even level 1 and level 15 character. I would say the fix is for how character's max_cols and max_rows are calculated.

eos428:
Just cross-checked with master, and it seems these issues are also present there. Can you check if these issues are there on your own master branch?

PR #541, opened 2018-09-02
CharacterPower.m_power_tpl is now CharacterPower.get_power_template()

### LINKED ISSUE #526: Powers Template is not saved between sessions
### Summary
`m_powers_tpl` is a saved copy of the power template from powers.bin. We're currently not saving that to the database between sessions, so things like range, endurance and power type are no longer available upon second log in.

There are a few ways we could resolve this. Here are a couple:
1- Save `m_powers_tpl` to database for powers, enhancements and inspirations.
2- Change `m_powers_tpl` to a method that retrieves a pointer to the data stored in MapServerData, like we do with `CharacterPower getPowerData(PowerPool_Info &ppool)` so that we aren't copying the data dozens of times per character.

Resolution 2 seems like the best idea. But I'm open to alternatives.

eos428:
Closes #526 .

- Using CharacterPower.get_power_template(), which takes the template from MapServerData based on m_power_info's pcat, pset, and pow_idx

The How's:

- Created CharacterPower.get_power_template()
- Changed serialization on CharacterPower
- Removed m_power_tpl, m_name, m_num_charges, m_usage_time, m_activation_time, m_range, m_recharge_time from CharacterPower
- The variables removed above are now obtained from get_power_template()
- MapServerData has power-related functions in which you can grab them without going through the layers in the calling functions

Code architecture in functions that take in the powers from MapServerData is changed a bit (For example, to get a particular powerset from MapServerData, you now call getMapServerData()->get_powerset()) so I hope you don't mind. Also, there's no exact equivalent for m_activation_time in PowerData, so I'll assume it is ActivationPeriod


broxen:
I'll review this in detail later, but one thing I'll note is that m_activation_time will be a time point needed for calculating things like recharge timer, and NetFx for powers (across the network)

For now we can remove it, but eventually we'll need to keep track of when a power was fired.

nemerle:
The functions returning stuff from the templates should most likely return `const &` instead of copies of the values :)

broxen:
With a cursory glance, I really like this implementation. I wish I would have thought of it.

If no one tests before I'm back, I'll review this on Tuesday.

As always, great work eoS

eos428:
For now I'll leave m_activation_time as get_power_template().ActivationPeriod, I guess.
The functions in MapServerData now return const&

Np broxen

broxen:
Looks good. Needs testing and merge

Projects/CoX/Common/NetStructures/Powers.h
```diff
@@ -79,21 +81,19 @@ struct CharacterPower
...
-        float           m_range             = 1.0f;
-        float           m_recharge_time     = 0.0f;
         uint32_t        m_activation_state  = 0;
         uint32_t        m_total_eh_slots    = 0;
         bool            m_active_state_change   = false;
         bool            m_timer_updated         = false;
         bool            m_erase_power           = false;
         std::array<CharacterEnhancement, 6> m_enhancements;
+
+        Power_Data get_power_template() const
```
broxen: To keep with our coding format, let's rename this method to `getPowerTemplate()`

Projects/CoX/Common/NetStructures/Powers.cpp
```diff
@@ -215,7 +215,7 @@ int getPowerSetByName(const GameDataStore &data,const QString &name, uint32_t pc
 {
     int idx = 0;
 
-    for(const Parse_PowerSet &pset : data.m_all_powers.m_categories[pcat_idx].m_PowerSets)
+    for(const Parse_PowerSet &pset : getMapServerData()->get_power_category(pcat_idx).m_PowerSets)
```
broxen: With your changes, we no longer need the GameDataStore variable and it becomes unused. Can we remove it from these functions?

I feel your getMapServerData() method is cleaner.

eos428: done :)

broxen:
Getting linking errors with this one. It appears as though NetStructures can no longer access `DataHelpers.h`

Perhaps @nemerle can provide some guidance.
```
../../../Common/NetStructures/libnetwork_structures.a(Character.cpp.o): In function `Character::finalizeLevel()':
/Segs/Projects/CoX/Common/NetStructures/Character.cpp:102: undefined reference to `getMapServerData()'
../../../Common/NetStructures/libnetwork_structures.a(Character.cpp.o): In function `CharacterPower::getPowerTemplate() const':
/Segs/Projects/CoX/Common/NetStructures/Powers.h:177: undefined reference to `getMapServerData()'
../../../Common/NetStructures/libnetwork_structures.a(Powers.cpp.o): In function `getPowerCatByName(QString const&)':
/Segs/Projects/CoX/Common/NetStructures/Powers.cpp:203: undefined reference to `getMapServerData()'
../../../Common/NetStructures/libnetwork_structures.a(Powers.cpp.o): In function `getPowerSetByName(QString const&, unsigned int)':
/Segs/Projects/CoX/Common/NetStructures/Powers.cpp:219: undefined reference to `getMapServerData()'
../../../Common/NetStructures/libnetwork_structures.a(Powers.cpp.o): In function `getPowerByName(QString const&, unsigned int, unsigned int)':
/Segs/Projects/CoX/Common/NetStructures/Powers.cpp:235: undefined reference to `getMapServerData()'
../../../Common/NetStructures/libnetwork_structures.a(Powers.cpp.o):
/Segs/Projects/CoX/Common/NetStructures/Powers.cpp:249: more undefined references to `getMapServerData()' follow
```

broxen:
Per Nem, maybe we use a singleton here for GameData, like we did with Settings class.

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -98,21 +98,22 @@ void Character::sendTray(BitStream &bs) const
...
 
-    if(m_char_data.m_experience_points < data.expForLevel(m_char_data.m_combat_level))
-        m_char_data.m_experience_points = data.expForLevel(m_char_data.m_combat_level);
+    if(m_char_data.m_experience_points < data->expForLevel(m_char_data.m_combat_level))
+        m_char_data.m_experience_points = data->expForLevel(m_char_data.m_combat_level);
 
-    m_char_data.m_max_insp_rows = data.countForLevel(m_char_data.m_level, data.m_pi_schedule.m_InspirationRow);
-    m_char_data.m_max_insp_cols = data.countForLevel(m_char_data.m_level, data.m_pi_schedule.m_InspirationCol);
-    m_char_data.m_max_enhance_slots = data.countForLevel(m_char_data.m_level, data.m_pi_schedule.m_BoostSlot);
+    m_char_data.m_max_insp_rows = data->countForLevel(m_char_data.m_combat_level, data->m_pi_schedule.m_InspirationRow);
```
broxen: these need to use m_level instead of m_combat_level, otherwise we wind up with the dreaded 1x4 max inspiration slots bug =)

In fact, I strongly suspect that all calculations ran through `countForLevel` should use m_level instead of the adjusted combat_level

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -98,21 +98,22 @@ void Character::sendTray(BitStream &bs) const
     m_char_data.m_trays.serializeto(bs);
 }
 
-void Character::finalizeLevel(const GameDataStore &data)
+void Character::finalizeLevel()
 {
-    uint32_t max_xp = data.expMaxLevel();
-    if (m_char_data.m_level == 4294967295){m_char_data.m_level = 0;}// /levelup can set this negative, need to bring it up
```
broxen: We still need some sort of check here to limit m_level to a range 0...49

Projects/CoX/Common/NetStructures/Character.cpp
broxen: Testing locally confirms. All countForLevel checks should use m_level.

broxen:
Great PR. Thanks eoS!

PR #548, opened 2018-09-05
Creating a Semi-Standalone EmailHandler that derives from EventProcessor (Part 2)

### LINKED ISSUE #367: Create a semi-standalone EmailHandler

eos428:
Closes #367 .

Continuation of the previous PR that has since been pulled down because I changed its name.
At the moment it runs, but the actual MapInstance/DataHelper-EmailHandler interaction is not there yet.


eos428:
Remaining things to do:
- Parse EmailSend correctly in slashCommands. It seems to send the name of the recipient instead of the db_id which does complicate things a bit...
- EmailRead and EmailDelete should be alright, since you'd just need the email id and the handlers will deal the rest
- Create Email column in Database and integrate it with EmailHandler, DbHandler and DbContext

At the moment these are the major things to do. I expect further issues to arise, and then there's the matter of polishing it too.

broxen:
Just a note here. As emails are sure to be updated less frequently than CharData, maybe we should stick them in their own table linked to char by db_id as pk?

eos428:
That's the plan from the beginning. The question is how do I make them?

Once the email table is up and running and integrated with the email handler, I'd say we'd be really close to being done here :D

eos428:
Some questions I have based on my latest commit (And yeah, it's a bit messy and plenty of placeholders but at least it can run):

- Getting emails through sender_id and recipient_id would return multiple emails, should I send an entire vector containing all the emails over a single event?

There are two cases where we'd update an email: To mark the email as read, and when the sender character and/or account is deleted.
- Should we put two separate events for this (Something like EmailMarkAsReadMessage and EmailUpdateOnCharacterDeleteMessage)?
- And should we put m_is_read_by_recipient outside the blob, since sending the entire email data just to mark it as read feels a little cumbersome.

EmailRemove should be okay, since the only case we'd actually delete the email is from the recipient deleting it from his/her inbox.

That's all for now. I am getting confident we'd finish this before 0.6.0 releases :D

Projects/CoX/Data/sqlite/segs_game_sqlite_create.sql
```diff
@@ -13,6 +13,7 @@ INSERT INTO table_versions VALUES(3,'accounts',1,'2017-05-03 12:56:03');
 INSERT INTO table_versions VALUES(4,'characters',8,'2018-05-04 14:58:27');
 INSERT INTO table_versions VALUES(5,'costume',0,'2017-11-11 08:57:43');
 INSERT INTO table_versions VALUES(7,'supergroups',1,'2018-05-03 12:56:43');
+INSERT INTO table_versions VALUES(8,'emails',1,'2018-09-23 08:00:00');
```
dracc: Arrays start at 0. :wink:
Correct row would be
```INSERT INTO table_versions VALUES(8,'emails',0,'2018-09-23 08:00:00');```

File: Projects/CoX/Data/pgsql/segs_game_postgres_create.sql [hunk 2]
```diff
@@ -14,6 +14,7 @@ INSERT INTO table_versions VALUES(3,'accounts',1,'2018-05-03 14:06:03');
 INSERT INTO table_versions VALUES(4,'characters',8,'2018-05-04 14:58:27');
 INSERT INTO table_versions VALUES(5,'costume',0,'2017-11-11 08:57:43');
 INSERT INTO table_versions VALUES(7,'supergroups',1,'2018-05-03 14:06:43');
+INSERT INTO table_versions VALUES(8,'emails',1,'2018-09-23 08:00:00');
```
dracc: Same thing here, regarding versioning

Projects/CoX/Data/mysql/segs_game_mysql_create.sql
```diff
@@ -65,8 +73,8 @@ INSERT INTO `table_versions` (`id`, `table_name`, `version`, `last_update`) VALU
 (3, 'accounts', 1, '2018-05-03 12:52:03'),
 (4, 'characters', 8, '2018-05-04 14:58:27'),
 (5, 'costume', 0, '2017-11-11 08:57:43'),
-(7, 'supergroups', 1, '2018-05-03 12:52:53');
-
+(7, 'supergroups', 1, '2018-05-03 12:52:53'),
+(8, 'emails', 1, '2018-09-23 08:00;00');
```
dracc: Yeah, you guessed it. Here too. :stuck_out_tongue:

dracc:
Apart from the very tiny suggested changes, the SQL part of this PR looks good to me. :+1:

eos428:
Should be working fine except for sending email to names where some other character might have a more extended name of the recipient, such as admin2 sending an email to admin, which instead finds the recipient_id of admin2.

And, I think we can send an email to the same character. It was my oversight of not storing newly-created emails to ``m_stored_email_datas``.

broxen:
Hi eoS! Thank you for tackling the Email system!

I'm getting some errors on this PR though.
- Recipient of email doesn't receive email in their inbox
- Sender appears to have gotten email in their inbox instead? Read receipt fires immediately
- Sender client crashes when clicking on reply
- Writing an email to an offline character causes segfault, and the sender's ID is stored in the database for recipient id.
```
Debug   : Email Sent to recipient: EmailGuy1
authserver: /Segs/Projects/CoX/Common/Servers/ClientManager.h:118: SESSION_CLASS& ClientSessionStore<SESSION_CLASS>::session_from_token(uint64_t) [with SESSION_CLASS = MapClientSession; uint64_t = long unsigned int]: Assertion `iter != m_token_to_session.end()' failed.
```

screenshot from 2018-10-07 00-21-26

broxen:
Thank you eoS! Great PR!

Most of the fixes seemed to work, but I received a segfault when opening an email received while offline.
```
Debug   : Opening Email ID: 1
Debug   : Opening Email ID: 5
authserver: /Segs/Projects/CoX/Common/Servers/ClientManager.h:118: SESSION_CLASS& ClientSessionStore<SESSION_CLASS>::session_from_token(uint64_t) [with SESSION_CLASS = MapClientSession; uint64_t = long unsigned int]: Assertion `iter != m_token_to_session.end()' failed.
```
crashing when session is used here:
```
void MapInstance::on_email_read_by_recipient(EmailWasReadByRecipientMessage *msg)
{
    MapClientSession &map_session(m_session_store.session_from_token(msg->session_token()));
```

### Steps to reproduce:
- Create two characters on two separate accounts. Send emails back and forth. All OK.
- Log out char-2, create new char (char-3). Send email from char-3 to char-2, so far, all OK.
- Log out of char-3, and log in char-2. Select new email to read: segfault =(

### Notes
Database looks correct, in terms of sender_id and recipient_id equating to the db_id of the characters

### Other issues noted:
As mentioned on discord, sending to multiple recipients doesn't seem to work as intended, but I'm comfortable with leaving that for a future PR.

Also, as you've mentioned before, timestamps aren't working, but can be in a followup PR as well.

PR #552, opened 2018-09-06
Map transfer detection -- WIP

Magentah:
Will update soon.

PR to handle detecting entering tunnels/zone boundaries, possibly monorail/doors as well.


Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -461,11 +470,55 @@ void MapInstance::dispatch( Event *ev )
...
 
+void MapInstance::on_initiate_map_transfer(InitiateMapXfer *ev)
+{
+    
+    MapClientSession &session(m_session_store.session_from_event(ev));
+    MapLink *lnk = session.link();
+    MapServer *map_server = (MapServer *)HandlerLocator::getMap_Handler(m_game_server_id);
+    qCDebug(logMapEvents) << "Map Transfer Initiated with token" << lnk->session_token();
+    
+    if (map_server->session_has_xfer_in_progress(lnk->session_token()))
```
nemerle: consider inverting the branches of this if, and using early return:
```cpp
if ( !map_server->session_has_xfer_in_progress(lnk->session_token()) )
{
  qCDebug(logMapEvents) << QString("Client Session %1 attempting to initiate transfer with no map data message received").arg(session.link()->session_token());
  return;
}
```

Projects/CoX/Servers/MapServer/MapSceneGraph.cpp
```diff
@@ -253,4 +253,99 @@ std::vector<glm::mat4> MapSceneGraph::spawn_points(const QString &kind) const
...
+    std::vector<MapSwap> *m_targets;
+    MapSwapLocator(std::vector<MapSwap> *targets):
+        m_targets(targets)
+    {}
+    bool operator()(SceneNode *n, const glm::mat4 &v)
+    {
+        bool found_map_swap = false;
+        if (!n->properties)
+        {
+            if (n->children.size() > 0)
```
nemerle: no need to check if n->children.size() is 0, the for loop will just not run in that case

Projects/CoX/Servers/MapServer/MapServer.cpp
```diff
@@ -192,6 +194,36 @@ void MapServer::on_expect_client(ExpectMapClientRequest *ev)
...
+bool MapServer::session_has_xfer_in_progress(uint64_t session_token)
+{
+    return m_current_map_xfers.find(session_token) != m_current_map_xfers.end();
+}
+
+uint8_t MapServer::session_map_xfer_idx(uint64_t session_token)
+{
+    if (session_has_xfer_in_progress(session_token))
+        return m_current_map_xfers[session_token];
+}
```
nemerle: missing return value in case where xfer in progress is `false`

Projects/CoX/Common/GameData/map_definitions.cpp
```diff
@@ -35,6 +35,11 @@ static const std::vector<MapData> g_defined_map_datas =
     {23, "Trial_05_01", "maps/City_Zones/Trial_05_01/Trial_05_01.txt", "Rikti Crash Site"}
 };
 
+const std::vector<MapData> getAllMapData()
+{
```
nemerle: returning using `const &` might be a better aproach here.

broxen:
Just touching base. This is ready, but waiting on improvements to movement to work correctly?

Is that correct?

Magentah:
@broxen If I remember correctly, it works if you /moveto the transfer coordinated, however I may have broken it when I tried to make some changes last time - I haven't looked at this any more since it relies on the movement working correctly

broxen:
Would it be beneficial to resolve merge conflicts here and get this merged into the code base prior to resolving movement issues? Or will that introduce more bugs than enhancements?

Just working my way through the PR queue.

Magentah:
I think the last things I tried to work on with this actually broke map transfers completely, so I'll probably need to spend some additional time with it.

I think a good point I can get to without fixing the movement issues is loading in at the correct location when transferring maps. Not sure if there's a default spawn location on all maps though. Then when the movement is working, this should then work without any additional changes (hopefully).

broxen:
There is not a PlayerSpawn on every map.

That said, would it be possible to take a random door from the map and spawn there?

We might even be able to use the Door Animation Start (pkt 52) and Door Animation Exit (pkt 53) and maybe make it look all fancy schmancy?

I can provide details on those pkts in discord.

broxen:
Just an update. I went ahead and implemented in DoorAnims in #636 

With this, and some additional work on npc generation, we should be able to spawn from random doorways.

broxen:
Just touching base on this.

Originally we were waiting on some movement improvements, but last messages indicate that we might be able to merge in these improvements. Especially now that some additional work has completed towards loading doors and connecting doors to spawns.

@Magentah, what are your thoughts here? I'd like to get this in, if possible.

broxen:
@Magentah, I'll reach out on discord to follow up here.

PR #573, opened 2018-09-12
 Starting character infromation read from settings.cfg 

### LINKED ISSUE #572: Starting powers/inspirations configured in settings.cfg
### For Issue #525 
starting **powers** to be configured via settings.cfg
starting **inspirations** to be configured via settings.cfg

### Possible additions:
starting **level** to be configured via settings.cfg
starting **inf** to be configured via settings.cfg

In World of Warcraft emulation, there are "fun servers" where people will start at max level, or at some lower level but with exp gain turned off. This is mostly for PvP, but also sometimes challenges of what can be done at a set level. Starting at a higher level would come with starting money. The defaults would be set to level 1 and 0 inf of course.

### Discounted possibilities
enhancements and salvage and other things that haven't been added. Won't be as useful.

cyanpill:
Closes #572.

- moved starting powers and inspirations to settings.cfg
- added starting level and inf

### DIFF
--- a/Components/Settings.cpp
+++ b/Components/Settings.cpp
@@ -169,7 +169,7 @@ void Settings::setDefaultSettings()
         config.setValue("inherent_powers","Brawl");
         config.setValue("starting_temps","EMP_Glove");
         config.setValue("starting_inspirations","Resurgence");
-        config.setValue("level_bonus", 0);
+        config.setValue("starting_level", 1);
         config.setValue("starting_inf", 0);
     config.endGroup();
 
--- a/Components/Settings.cpp
+++ b/Components/Settings.cpp
@@ -165,6 +165,13 @@ void Settings::setDefaultSettings()
         config.setValue("log_npcs","false");
         config.setValue("log_animations","false");
     config.endGroup();
+    config.beginGroup("StartingCharacter");
+        config.setValue("inherent_powers","Brawl");
+        config.setValue("starting_temps","EMP_Glove");
+        config.setValue("starting_inspirations","Resurgence");
+        config.setValue("level_bonus", 0);
+        config.setValue("starting_inf", 0);
+    config.endGroup();
 
     config.sync(); // sync changes or they wont be saved to file.
 }
--- a/Projects/CoX/Common/NetStructures/Character.cpp
+++ b/Projects/CoX/Common/NetStructures/Character.cpp
@@ -28,6 +28,7 @@
 #include "Logging.h"
 #include <QtCore/QString>
 #include <QtCore/QDebug>
+#include "Settings.h"
 
 using namespace SEGSEvents;
 
@@ -127,14 +128,8 @@ void Character::finalizeLevel(const GameDataStore &data)
     m_char_data.m_powers_updated = false;
 }
 
-void Character::addStartingInspirations(const GameDataStore &data)
+void Character::addStartingInspirations(const GameDataStore &data, QStringList starting_insps)
 {
-    // TODO: We can make this configurable in settings.cfg
-    QStringList starting_insps = {
-        "Resurgence",       // 25.2.3
-        "Phenomenal_Luck",  // 25.2.0
-    };
-
     for (QString &name : starting_insps)
         addInspirationByName(data,m_char_data, name);
 }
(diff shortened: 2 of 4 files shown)


Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -168,37 +163,33 @@ void Character::getPowerFromBuildInfo(const GameDataStore &data,BitStream &src)
...
-    };
+    config.beginGroup("StartingCharacter");
+        QStringList inherent_and_preorders = config.value("inherent_powers","Brawl").toString().split(',');
+        QStringList starting_temps = config.value("starting_temps","EMP_Glove").toString().split(',');
+        QStringList starting_insps = config.value("starting_inspirations","Resurgence").toString().split(',');
+        int startlevel = config.value(QStringLiteral("starting_level"), "1").toInt() -1; //combat level is m_level +1, so it gets back to starting_level
+        int startinf = config.value(QStringLiteral("starting_inf"), "0").toInt();
+    config.endGroup();
+
+    if (startlevel<0)
```
broxen: Please see my other comment regarding how the client stores character level starting with 0 (lvl 1)

With that in mind, we should set this check to `< 1`

Components/Settings.cpp
```diff
@@ -165,6 +165,13 @@ void Settings::setDefaultSettings()
         config.setValue("log_npcs","false");
         config.setValue("log_animations","false");
     config.endGroup();
+    config.beginGroup("StartingCharacter");
```
broxen: Just a nit-pick, but I'd put this section above the logging categories as I feel it's more important and logging is sort of the last thought for server-admins (and typically used only when testing/bug-fixing)

Projects/CoX/Common/NetStructures/Character.h
```diff
@@ -78,7 +78,7 @@ const   QString &       getName() const { return m_name; }
         void            serialize_costumes(BitStream &buffer, const ColorAndPartPacker *packer, bool all_costumes=true) const;
         void            serializetoCharsel(BitStream &bs, const QString& entity_map_name);
         void            finalizeLevel(const GameDataStore &data);
-        void            addStartingInspirations(const GameDataStore &data);
+        void            addStartingInspirations(const GameDataStore &data, QStringList starting_insps);
```
broxen: I would pass `starting_insps` as a reference here.

Projects/CoX/Common/NetStructures/Character.cpp
cyanpill: Ah right, I had switched between 1+level bonus, to 0+startlevel, forgot to change this to 1.

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -168,37 +163,33 @@ void Character::getPowerFromBuildInfo(const GameDataStore &data,BitStream &src)
...
+        startlevel = 0;// make sure it isn't negative
+
+    m_char_data.m_level = startlevel;
```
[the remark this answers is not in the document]
cyanpill: The finalize level function is called right after this, and sets combat level to +1. It's the same as in the slashcommand /levelup, where level is set and only when finalized is called does it change the combat level. In the future we will want combat level to be separate for side-kicking and such.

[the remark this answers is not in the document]
broxen: Perfect

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -100,7 +101,7 @@ void Character::sendTray(BitStream &bs) const
 void Character::finalizeLevel(const GameDataStore &data)
 {
     uint32_t max_xp = data.expMaxLevel();
-
+    if (m_char_data.m_level == 4294967295){m_char_data.m_level = 0;}// /levelup can set this negative, need to bring it up
```
broxen: This would work, but perhaps we should just cap level between range 0 and 49, to cover all cases?

cyanpill: `if(m_char_data.m_combat_level >= max_xp)
        m_char_data.m_combat_level = max_xp - 1;`
That runs right after my code and limits the level to 50. I just make sure it isn't level 0. Which could be done after that code, but it's the same effect.

broxen:
Will pull down to test and merge on Monday when I'm back in town. Unless someone gets to it first.

Thanks Cyan!

ldilley:
Enjoy your trip, @broxen. I got this. Nice work, @cyanpill. I believe Appveyor is still borked. This PR builds fine using Clang and GCC and I've validated that the settings are properly read.

PR #574, opened 2018-09-14
Attempt to fix map transfers causing session reaping and disconnects

### LINKED ISSUE #565: Link Reaping During Zone Transfers Causes Timeout
### Expected behavior
Transferring zones occurs seamlessly 
### Actual behavior/symptoms
Upon attempting to switch zones through any means, a loading screen appears and after a few moments, the player is sent back to the title screen with the prompt, "Lost connection to server" 
1. From inside the game, use any means to transfer zones
2. Player will be greeted by a loading screen, and at this point links are reaped
3. After a few moments the player should be sent back to the title screen
### SEGS version
Lastest commit
### Operating system
Windows 10
### C++ compiler and version
QT Creator

Magentah:
Attempts to close #565 .

Attempt to fix the issue that sessions were being reaped sometimes after transferring between maps. This adds a check to the GameHandler disconnect handler to only reap the session if the map server the disconnect is from and the map server the session is connected to are the same.

### DIFF
--- a/Projects/CoX/Common/Servers/InternalEvents.h
+++ b/Projects/CoX/Common/Servers/InternalEvents.h
@@ -204,11 +204,11 @@ struct ClientDisconnectedData
 {
     uint64_t m_session;
     uint32_t m_char_db_id;       // id of the character disconnected
-    uint32_t m_map_server_id;
+    uint32_t m_map_instance_id;
     template<class Archive>
     void serialize(Archive &ar)
     {
-        ar(m_session,m_char_db_id, m_map_server_id);
+        ar(m_session,m_char_db_id, m_map_instance_id);
     }
 };
 //[[ev_def:macro]]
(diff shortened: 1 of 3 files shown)


broxen:
Thank you for Tackling this elusive issue.

Did testing with @dracc and James provide any results?

dracc:
We got a little output, a few kinks were straightened but the bug was still alive and kicking when the last debug-session stopped.

broxen:
If there are other improvements here, and no foreseeable consequences. I'll move this to test/merge.

Otherwise I'll leave it open until the target issue is resolved.

Thoughts?

Magentah:
@broxen I actually don't think any of the changes made had any impact, and feel like this branch can be abandoned/closed. I want to go through it a bit with @nemerle as from what I can see there's not really any out-of-order handling of sessions or anything that would cause the issue.

broxen:
Thanks @Magentah, will wait for you and @nemerle to discuss.

dracc:
Lol, it seems this might be entirely caused by my server not having enough opened ports. Opening 7002-7030 seems to solve the problem. Sorry everyone!

broxen:
I'm closing this, as it looks like the ports are the main issue. Please reopen if there are other meaningful changes here that should be considered.

Thanks team!

Issue #545, opened 2018-09-04
Properly award new powers and additional inspiration slots on level up


eos428:
We already have /levelup so the actual level up process is at least done.

The things that should be done to solve this issue should be:
- Check if character's exp is enough to level up (should be handled by client already?), but this might be for another issue if needed
- On each level up, check if the character level is enough to award additional inspiration slots
- On each level up, check if the character level reaches the amount required to be granted certain powers

broxen:
So some insight into the client side here:

The client already checks for these things, but on the server side we need to verify (as server has authority). There's also a couple of opcodes thrown back to the server when running `/levelup` (possibly not every time?) so we'll need to handle those as well.

Another thing you may want to add to this: we can adjust max-attribs based upon level. It may rely upon #502 to get per-class by level data.

I'll be more helpful with information after I've had some hands-on over the next couple of days.

broxen:
A note on this: we're not currently handling adding new empty enhancement slots to powers correctly, and the server just forces every power to have 3 empty enhancement slots

cyanpill:
We should add something like "trained level" to keep track while you are training. [According to the wiki it was "security level"](https://paragonwiki.com/wiki/Level)

PR #578, opened 2018-09-18
Fix erroneous inspiration counts and fixed messages when adding…

### LINKED ISSUE #577: countForLevel() is broken when level exceeds the value of schedule[i]
### Summary
countForLevel() fails to return the correct value when lvl is greater than the value of all members of schedule array.

broxen:
Fix erroneous inspiration counts and fixed messages when adding inspirations and enhancements using slash commands

Closes #577 and the 1x4 starting insp array issue identified in Discord
Relates to #545 

- Modified countForLevel() to ensure it sends values when lvl is higher than pischedule value
- Modified `/addinsp` and `/addboost` to check for full insp or enhancement trays before sending messages to client
- Modified finalizeLevel() to utilize level instead of combat_level


Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -868,15 +868,23 @@ void cmdHandler_AddInspiration(const QString &cmd, MapClientSession &sess)
     CharacterData &cd = sess.m_ent->m_char->m_char_data;
     int space = cmd.indexOf(' ');
     QString val = cmd.mid(space+1);
-    QString floating_msg = FloatingInfoMsg.find(FloatingMsg_FoundInspiration).value();
-    QString msg = "Awarding Inspiration '" + val + "' to " + sess.m_ent->name();
+    QString msg = "There was an error adding this inspiration!";
```
dracc: nit: Does `msg` need an initialization here, as it's always changed anyways?

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -885,21 +893,28 @@ void cmdHandler_AddEnhancement(const QString &cmd, MapClientSession &sess)
     QVector<QStringRef> args(cmd.splitRef(' '));
     QString name = args.value(1).toString();
     uint32_t level = args.value(2).toInt();
-
-    QString floating_msg = FloatingInfoMsg.find(FloatingMsg_FoundEnhancement).value();
-    QString msg = "Awarding Enhancement '" + name + "' to " + sess.m_ent->name();
+    QString msg = "There was an error adding this enhancement!";
```
dracc: Same here as above

Projects/CoX/Servers/MapServer/SlashCommand.cpp
broxen: Thanks for reviewing!

You're completely right. I can simplify this section

PR #580, opened 2018-09-19
[WIP] Movement

### LINKED ISSUE #78: Movement has a lot of rubberbanding and collision issues
Moving results in a lot of rubberbanding and collision issues. This might be fixed by #44?

screenshot from 2017-10-20 09-58-07

### LINKED ISSUE #94: [Movement] Rubberbanding movement and Interpolation

broxen:
## Summary
I'm putting this out there for review as WIP. It's not pretty, but hopefully with several eyes on it, we can either borrow from this for a new clean branch, or improve what's here to get everything working.

Closes #94

- StateStorage - last 30 states
- MotionState - current calculated motion state of each entity
- TimeState - client provided time states
- Movement.cpp/h - contains all movement related
- Interp/Velocity/entMotion/etc. -- See Movement.cpp. Lots going on there. Most of this is ripped from the client.
- Added SeqBits prefix
- SeqBit serializers
- `/interp` `/collision` `/movement` `/setseq` `/addTriggeredMove` `/setTimeStateLog` slash commands

### This WIP PR has the following issues currently:
- Crashes on Windows `*** stack smashing detected ***: <unknown> terminated` when Interp is calculated
- Entities are rotated 180deg. I can't figure out what is causing this. It appeared after Merging changes made upstream
- Interpolation causes bounciness on other entities.
- Movement is basically borked. You'll see =P

### Note:
- `my_entMoveNoColl(e)` should line up with movement on the client when collision is toggled off `/nocoll 1`
- 'entMotion(...)` should line up with normal movement, and should take the global surface params and modify them based upon the texture you're standing upon.
- you can use `/setu1` to toggle between different movement methods in WorldSimulation.cpp
- The code needs a large amount of clean up.


broxen:
Seems like the client is using movement prediction when an entity is in follow mode `/follow $target`, so perhaps we can look to the logic there for some insight into how the server should be predicting movement between server<-->client updates.

Also, I found this helpful:
http://www.gabrielgambetta.com/client-side-prediction-server-reconciliation.html

broxen:
I might break out two or Three smaller PRs from this:
- AddTimeStateLog
- Movement Logging
- SeqBits

Those might help clean up the focus of this PR.

Also, I noticed a possible typo in StateInterpolator.cpp where I've used glm::mix to replace a former lerp method. That may explain the bounciness when interping.

Projects/CoX/Servers/MapServer/UnitTests/MapEventRegistry.cpp
```diff
@@ -12,8 +12,12 @@ namespace
         "ActivateInspiration",
         "ActivatePower",
         "ActivatePowerAtLocation",
+<<<<<<< HEAD
```
broxen: ooooops

broxen:
So much of this has been reimplemented in other PRs.

Will resubmit after #632 is merged and I've resolved any conflicts.

PR #583, opened 2018-09-20
Code cleanups and Data Helper organization

broxen:
## Summary
Cleaned up DataHelpers, SlashCommand, Team, Friend, etc. Reorganized files to line up with their headers, and removed undefined declarations.

The objective to is to clean up DataHelpers and SlashCommands, and make it easier to find methods relating to their purpose.


broxen:
@nemerle you had originally moved a lot of these methods into DataHelpers. So I'd like your input before merging, as maybe there was a reason for that that I overlooked.

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -108,6 +108,8 @@ void cmdHandler_SendTimeUpdate(const QString &cmd, MapClientSession &sess);
 
 void cmdHandler_SetU1(const QString &cmd, MapClientSession &sess);
 
+void cmdHandler_SetU1(const QString &cmd, MapClientSession &sess);
+
```
nemerle: this looks like a duplicated  `cmdHandler_SetU1` entry

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -207,6 +209,8 @@ static const SlashCommand g_defined_slash_commands[] = {
 
     {{"setu1"},"Set bitvalue u1. Used for live-debugging.", cmdHandler_SetU1, 9},
 
+    {{"setu1"},"Set bitvalue u1. Used for live-debugging.", cmdHandler_SetU1, 9},
+
```
nemerle: duplicated here as well :)

PR #588, opened 2018-09-22
get maxHP from classes.bin and cleaned up combat level

### LINKED ISSUE #502: Pull per class values for BaseAttribs and MaxAttribs from classes.bin
Per-class values for BaseAttribs (Starting attribute values) and MaxAttribs exist in classes.bin

We're already pulling some of this info for use in `m_max_attribs`, but we should be utilizing all of those values where appropriate (for example: starting HP and End, and starting maxHP and maxEnd)

**To Complete:**
- Start with `m_max_attribs` and work your way down the rabbit hole. =D
- Create method for resetting those values based upon level (so when we level up a character we can `getMaxAttribs(ent.m_char);` and set those values to their new max)

One could use bintool to extract a json file of classes.bin for reference.

cyanpill:
Closes #502.

Gets maxHP and maxEnd from AttribMaxTable, and sets based on class/archetype and level.
Also cleared up some more uses of combat_level, and changed it to be 0-49 for consistency with level.

The hp values are a little off from what the wiki lists, but these are the values in classes.bin.

At a latter date, AttribMaxTable could be overwritten from a cfg/lua file to customize the values.

### DIFF
--- a/Projects/CoX/Common/GameData/chardata_serializers.cpp
+++ b/Projects/CoX/Common/GameData/chardata_serializers.cpp
@@ -191,7 +191,7 @@ void serialize(Archive &archive, CharacterData &cd, uint32_t const version)
     }
 
     archive(cereal::make_nvp("Level",cd.m_level));
-    archive(cereal::make_nvp("CombatLevel",cd.m_combat_level+1));
+    archive(cereal::make_nvp("CombatLevel",cd.m_combat_level));
     archive(cereal::make_nvp("XP",cd.m_experience_points));
     archive(cereal::make_nvp("Debt",cd.m_experience_debt));
     archive(cereal::make_nvp("PatrolXP",cd.m_experience_patrol));
--- a/Projects/CoX/Common/GameData/chardata_serializers.cpp
+++ b/Projects/CoX/Common/GameData/chardata_serializers.cpp
@@ -191,7 +191,7 @@ void serialize(Archive &archive, CharacterData &cd, uint32_t const version)
     }
 
     archive(cereal::make_nvp("Level",cd.m_level));
-    archive(cereal::make_nvp("CombatLevel",cd.m_combat_level));
+    archive(cereal::make_nvp("CombatLevel",cd.m_combat_level+1));
     archive(cereal::make_nvp("XP",cd.m_experience_points));
     archive(cereal::make_nvp("Debt",cd.m_experience_debt));
     archive(cereal::make_nvp("PatrolXP",cd.m_experience_patrol));
--- a/Projects/CoX/Common/NetStructures/Character.cpp
+++ b/Projects/CoX/Common/NetStructures/Character.cpp
@@ -62,7 +62,7 @@ Character::Character()
 void Character::reset()
 {
     m_char_data.m_level = 0;
-    m_char_data.m_combat_level = m_char_data.m_level + 1;
+    m_char_data.m_combat_level = m_char_data.m_level;
     m_name="EMPTY";
     m_char_data.m_class_name="EMPTY";
     m_char_data.m_origin_name="EMPTY";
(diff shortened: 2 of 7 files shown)


Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -103,21 +103,18 @@ void Character::finalizeLevel()
     GameDataStore *data = getMapServerData();
     uint32_t max_xp = data->expMaxLevel();
 
-    if (m_char_data.m_level >= 50)
-        m_char_data.m_level = 0;// /levelup can set this negative, need to bring it up
+    if(m_char_data.m_level > max_xp)
```
dracc: Is `max_xp` the correct value to do a comparison with? In that case, one of these variables should probably be treated with a name change.

cyanpill: True, it should be max_level. I was just keeping what already worked before.

Projects/CoX/Common/GameData/chardata_serializers.cpp
```diff
@@ -191,7 +191,7 @@ void serialize(Archive &archive, CharacterData &cd, uint32_t const version)
     }
 
     archive(cereal::make_nvp("Level",cd.m_level));
-    archive(cereal::make_nvp("CombatLevel",cd.m_combat_level));
+    archive(cereal::make_nvp("CombatLevel",cd.m_combat_level+1));
```
broxen: We probably don't need to store this +1

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -101,15 +101,10 @@ void Character::sendTray(BitStream &bs) const
...
-        m_char_data.m_level = 0;// /levelup can set this negative, need to bring it up
-
-    m_char_data.m_combat_level = m_char_data.m_level + 1; // m_combat_level is display level?
-
-    if(m_char_data.m_combat_level >= max_xp)
-        m_char_data.m_combat_level = max_xp - 1;
+    if(m_char_data.m_level > max_level)
+        m_char_data.m_level = max_level - 1;
 
     if(m_char_data.m_experience_points < data.expForLevel(m_char_data.m_combat_level))
```
broxen: This should probably be changed to m_level, as your maxXP will not change when sidekicked.

Projects/CoX/Common/NetStructures/Powers.cpp
```diff
@@ -843,7 +843,7 @@ void reserveEnhancementSlot(CharacterData &cd, CharacterPower *pow)
 
     // Modify based upon level
     pow->m_total_eh_slots = pow->m_total_eh_slots + getGameData().countForLevel(
-                cd.m_combat_level - pow->m_level_bought, getGameData().m_pi_schedule.m_FreeBoostSlotsOnPower);
+                cd.m_level - pow->m_level_bought, getGameData().m_pi_schedule.m_FreeBoostSlotsOnPower);
```
broxen: This may actually be impacted by combat_level? When you Sidekick are some of your powers or enhancement slots disabled?

cyanpill: Right, you do lose access to powers based on level. Enhancements are kept, but reduced in a complicated way: https://paragonwiki.com/wiki/Exemplar_Effects_on_Enhancements

broxen: Oh weird.

Well in that case ignore me 😋

Projects/CoX/Common/NetStructures/Character.cpp
cyanpill: Ah, I see it. Yes, I forgot this when merging the upstream, good catch :D

Issue #519, opened 2018-08-15
[Meta] How Can I Help? aka Future Server Features (and Opcodes to implement)


broxen:
## Summary
We need to implement these items one by one. I'm marking this `good-first-issue` as although these aren't all simple tasks, this can become a great resource for new members of all skill levels looking for how to help in meaningful ways.

When you take on one of the tasks (or groups) below, please [create a new issue](https://github.com/Segs/Segs/issues/new) to track your progress. Thank you!

See FloatingDamage for reference on how to implement server->client pkt communication: [Segs/Projects/CoX/Servers/MapServer/Events/FloatingDamage.h](https://github.com/Segs/Segs/blob/develop/Projects/CoX/Servers/MapServer/Events/FloatingDamage.h)

## Check the Issue Queue
- [Issues labeled `bug`](https://github.com/Segs/Segs/labels/bug)
- [Issues labeled `good first issue`](https://github.com/Segs/Segs/labels/good%20first%20issue)
- [Issues labeled `enhancement`](https://github.com/Segs/Segs/labels/enhancement)
- [Issues labeled `wishlist`](https://github.com/Segs/Segs/labels/wishlist)

### Low Difficulty
- Force Client Logout (boot client) -- useful for GM's to disconn. unruly guests
- Client Console Output -- print to client console
- Client Console PrintF -- printf to client console
- Correlate Map Indexes with Original Servers -- In Review: see #584
- Create "Current Features" list on Wiki -- see #554
- Replace `std::unique_ptr<T>(newT())` with `std::make_unique()` calls -- see #532 
- Normalize basic type arguments -- see #357 
- Update server architecture on wiki -- see #333 

### Moderate Difficulty
- Stores
- SuperGroup Registration *-- follow progress in #549* 
- Bug Report Window and researching a possible backend for linking this to github's issue tracker: #287 #35 
- Initiate Respec - using `/start_menu 13` somehow?
- Implement City Info Kiosk and Stats -- This is partially done (see: StandardDialogCmd, and implement `/kiosk home`). Will need to track Victors, Punishers, Healers, and Nemeses on a daily and monthly basis.
- Database schema migrations for updates to the database
- Map Cells Fog of War (packet 22) - Partially implemented #371
- Control Player (pkt 38)
- *Send Team-removal and Sidekick-removal messages -- see #555 and #642* 

### High Difficulty
- StatisticsService
[...]

PR #596, opened 2018-09-30
Waypoints implementation with slash command to test `/setwaypoint`

broxen:
Relates to #519 

- SendWaypoints event
- `/setwaypoint` to test. NOTE: client will only set waypoint if the idx is equal to the client-side idx. This creates a small issue where you cannot simply set a waypoint without one having been set already. In the future, we'll be able to do with with Tasks or Contacts. Right now, the only mechanism for doing so with in the UI (clicking on the map)
- Cleaned up a couple of small bugs in slashcommands
- Moved SetDestination to Waypoints.h for organization

### DIFF
--- a/Projects/CoX/Common/NetStructures/Entity.h
+++ b/Projects/CoX/Common/NetStructures/Entity.h
@@ -99,6 +99,12 @@ enum class FadeDirection
     Out
 };
 
+struct Destination // aka waypoint
+{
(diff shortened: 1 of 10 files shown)


broxen:
Another interesting thing here: if you don't close the client between sessions, the idx will continue to be iterated on the client side. :man_shrugging:

Issue #525, opened 2018-08-28
[Meta] Powers Roadmap


broxen:
## Summary
Now that #495 has been merged, I'm sure there will be interest in finishing the powers implementation. Here's a run-down on what the current road map for fully functioning Powers might look like. Suggestions very much welcome.

When tackling these, please create a new issue to track your progress, and cross-reference this Issue. Nem and I are available to answer any questions or to provide guidance on any particular issue.

Thank you!

### Simple Features To Implement
- [ ] Allow starting powers to be configured via settings.cfg
- [ ] Allow starting inspirations to be configured via settings.cfg
- [ ] Properly consider PowerType from m_power_tpl when using powers
- [ ] Properly consider Range from m_power_tpl when using powers
- [ ] Properly consider Accuracy from m_power_tpl when using powers
- [ ] Properly consider TargetTypes allowed from m_power_tpl when using powers
- [ ] Properly handle `$$` in commands sent to server (split to new command)

### Complex Features to Implement
- [ ] Power Animations (seqbits, and triggeredmoves)
- [ ] State Mode (needs research; this changes on power use)
- [ ] NetFx aka particles for powers!
- [ ] Active Powers Queue
- [ ] Recharge Timers
- [ ] Store Powers damage and other info not available in client files (maybe as json, or yaml?) on server
- [ ] Fully handle Leveling Up and awarding new powers/etc.
- [ ] Apply Inspiration Benefits when used (maybe using LUA?!?!)
- [ ] Apply Enhancement benefits to powers when used (maybe using LUA?!?!)
- [ ] Apply proper effect when Powers are used (maybe using LUA?!?!)

### Implemented Features
- [x] Grant Starting Powers
- [x] Grant Inherent Powers
- [x] Grant Temporary Powers at Creation (required by client)
- [x] Grant Starting Inspirations
- [x] Load/Save to Database, includes Powers Tray, Enhancements, Inspirations, Owned Powers
- [x] Handle Using/Moving/Trashing Inspirations, Enhancements, Powers
- [x] Handle Adding new Powers

## Related Issues and PRs
- [x] Merge #495 for preliminary powers support!
- [x] This issue should be added to #519 
- [ ] Add new issues as they relate

eos428:
I am interested in doing some more of these. Can you open two issues for:
- Active Powers Queue and Recharge Timers
- Considering PowerType, Range, etc when using powers, should be condensable into 1 issue I believe

[...]

PR #598, opened 2018-10-01
Levelup improvements, powers array improvements, queues and timers, and more

### LINKED ISSUE #190: MapInstance:on_abort_queued_power is unhandled
### Summary
This is one of many events in MapInstance.cpp that is unhandled in one way or another.
### Details
Begins on line 681 of MapInstance.cpp.
When called, it gives "Unhandled abort queued power request" to qWarning().

### LINKED ISSUE #545: Properly award new powers and additional inspiration slots on level up

### LINKED ISSUE #547: Store Powers in a container so that they can be Queued and Recharged
Again pending input from broxen, but here's how I see the problem:

- When a character tries to activate a power, the power is put into the queue, pending activation
- The power must wait until its assorted activation period is over before being casted
- After the 1st power is casted, the power is put into another queue in which it is being recharged.

My suggestion on how to solve this:
- Create two vectors referencing powers ActivatedPowersQueue and RechargingPowersQueue
- ActivatedPowersQueue will handle the timer of only the 1st power inside it (unless you want to cast multiple powers at the same time), and will remove the 1st power after being casted. Then the 2nd power onwards will be put 1 idx in front and so on.
- RechargingPowersQueue stores all powers that are casted and is pending recharge. In this queue, all powers' recharge timer will be handled at the same time.

If a character tries to cast a power that is already inside one of these vectors, then we cannot cast the assorted power.

### LINKED ISSUE #575: Empty m_keybind_profiles vector does not fail gracefully.
### Summary
Sometimes Keybind profiles do not load and server will assert due to vector out of range, aka Empty m_keybind_profiles vector does not fail gracefully.

Here, in `getCurrentKeybinds()` of keybind_definitions.cpp
```cpp
    qCDebug(logKeybinds) << "Could not get Current Keybinds. Returning first keybind profile.";
    return m_keybind_profiles.at(0).KeybindArr;  // <--- fails here
```

Previously this was not an issue, because we would initialize keybinds from MapServerData. This is no longer possible because NetStructures cannot access `g_GlobalMapServer`

It does not happen consistently, but appears to happen most often when the client has been left open between server sessions. It always fails at the same line.

broxen:
## Summary
Started by fixing a couple of bugs in powers array, ended with a laundry list of improvements

Relates to #525 
Closes #545
Closes #547
Closes #575
Closes #190

- Added security threat to Character class
- Added charges and lifetime timer to powers
- Attempted fix for keybinds failing gracefully #575 
- Players now start with the correct Health and End per class
- Powers must be "reset" by the client whenever the array (vector in our case) changes
- Fix for empty power names in settings.cfg
- Automatically add inherent powers based upon level (Rest at lvl 2)
- Lots of enhancement slots fixes, including starting slots of 1, and awarding eh_slots on levelup
- Added methods to set HP, End etc to max
- Powers now queue properly (my apologies to @eos428 )
- Powers now queue recharge timers properly (again, my apologies @eos428 )
- Client sometimes expects us to send it the IDXs of our owned powers in the array we sent it earlier. For this reason, we need to count the array position of our powers, enter: `struct PowerVecIndexes` and `getOwnedPowerIndexes()`
- Some refactoring, especially `countAllOwnedPowers()` and moved power related events to PowerEvents.h
- LevelUpResponse now works. Test with `/level` `/levelup` and `/levelupxp`
- Players are now able to properly buy enhancement slots on levelup. Previously this would crash the client
- Players are now able to set, and unset default power
- DEF, HP, and END inspirations now do their thing (with hardcoded bonuses)
- AbortQueuedPower now works as expected
- Changed uses of target_idx to accept signed values (-1 appears to be "self" in client)
- Buffs work with hardcoded timer, try using Luck inspiration to test

### Some notes:
- ~~`/levelup` appears to have a client bug? It checks against number of powers but does not consider Prestige powers as part of that calculation. Also, it seemingly adds slots and powers automatically? Try using `/level` instead, but make sure your XP is high enough. @nemerle maybe I'm overlooking something in the client here?~~ this is resolved in latest commit. use `/train` and ensure your xp is high enough
- It's really important that we toggle `m_has_updated_powers` to update the UI whenever we "touch" powers. It's also really important to update `m_reset_powers` when we change the powers array.
- There's probably a better way to do the timers. I'm open to ideas.
- UseInspiration and UsePower still need to be refactored out to consider type, target, etc etc
- Sorry if I stepped on anyone's toes with this one, I tried to look for WIP on forks to make sure we weren't duplicating code.


Projects/CoX/Common/NetStructures/Powers.h
```diff
@@ -15,6 +15,7 @@
 #include <QtCore/QString>
 #include <QtCore/QDebug>
 #include <array>
+#include <chrono>
```
nemerle: Does not seem to be used in this file?

Projects/CoX/Common/NetStructures/Powers.h
```diff
@@ -46,6 +47,23 @@ struct EnhancemenSlotEntry
...
+{
+    PowerVecIndexes m_pow_idxs;
+    bool            m_active_state_change   = false;
+    bool            m_activation_state      = false;
+    bool            m_timer_updated         = false;
+    float           m_recharge_time         = 0.0f;
+    float           m_time_to_activate      = 0.0f;
+    float           m_activate_period       = 0.0f;
+};
+
```
nemerle: Try to sort the struct members by size, while putting the most used field at the start :)
This leads to better structure packing, and in case of field 0, better performance

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -450,7 +451,7 @@ void sendTeamOffer(Entity *src, Entity *tgt)
     tgt->m_client->addCommandToSendNextUpdate(std::unique_ptr<TeamOffer>(new TeamOffer(db_id, name, type)));
 }
 
-void sendFaceEntity(Entity *src, uint8_t tgt_idx)
+void sendFaceEntity(Entity *src, int8_t tgt_idx)
 {
```
nemerle: Hmm, if this is tgt_idx, then uint8 or int8 will not be enough to store the value ?

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -942,20 +943,28 @@ void cmdHandler_AddEnhancement(const QString &cmd, MapClientSession &sess)
...
-
-    // levelup command appears to only work 1 level at a time.
-    if(level > sess.m_ent->m_char->m_char_data.m_level)
-        level = sess.m_ent->m_char->m_char_data.m_level + 1;
-
-    setLevel(*sess.m_ent->m_char, level);
-    sess.m_ent->m_char->finalizeLevel();
+    // must adjust level for 0-index array, capped at 49
+    uint32_t level = cmd.midRef(cmd.indexOf(' ')+1).toUInt();
+    if(level > 0)
```
nemerle: so this is basically a `clamp(level-1,0,49)` ?
do we have a clamp func rattling around the codebase ?

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -942,20 +943,28 @@ void cmdHandler_AddEnhancement(const QString &cmd, MapClientSession &sess)
...
 {
-    uint32_t level = cmd.midRef(cmd.indexOf(' ')+1).toUInt() -1; // convert from 1-50 to 0-49
-
-    // levelup command appears to only work 1 level at a time.
-    if(level > sess.m_ent->m_char->m_char_data.m_level)
-        level = sess.m_ent->m_char->m_char_data.m_level + 1;
-
-    setLevel(*sess.m_ent->m_char, level);
-    sess.m_ent->m_char->finalizeLevel();
+    // must adjust level for 0-index array, capped at 49
```
nemerle: Consider pulling out the magic const of MAX_LEVEL to at least file-local `static const int MAX_LEVEL=50` ?

Projects/CoX/Servers/MapServer/SlashCommand.cpp
broxen: not yet. C++ 17 though.

For now I'll `level = std::max(0, std::min(level, 49));`

broxen:
Ok. Fixes pushed.

I created a new slash command `/train` access_lvl = 0, which will require xp be high enough to level. For now I can setxp in that slash command for testing or leave it as is. Let me know.

Also, I fixed all the issues above, and left `/levelupxp` in the system, as it ties into a built-in client slash command.

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -100,9 +100,7 @@ void Character::finalizeLevel()
 
     GameDataStore &data(getGameData());
     uint32_t max_level = data.expMaxLevel();
-
-    if(m_char_data.m_level > max_level)
-        m_char_data.m_level = max_level - 1;
+    m_char_data.m_level = std::max(uint32_t(0), std::min(m_char_data.m_level, max_level));
```
cyanpill: data.expMaxLevel() returns 50, but data.expForLevel() wants 0-49 or it asserts.

cyanpill: m_char_data.m_level, max_level-1));
 or on the previous line
 int max_level = data.expMaxLevel() - 1;

broxen: Ahh good catch. I'll adjust.

The typecast is due to min throwing a fit about mismatched types.

broxen: Anyone opposed to me adjusting expMaxLevel to:
`return m_experience_and_debt_per_level.m_ExperienceRequired.size()-1;`

broxen:
Adjusted expMaxLevel() per comments. Let me know if that will suffice, and about the typecast

broxen:
Will resolve conflicts here, and clean up lookUpPower() (per discussion in #599) tonight. Then this should be ready for merge.

PR #599, opened 2018-10-01
Power use improvements

### LINKED ISSUE #546: Apply PowerType, Range, Accuracy, and TargetType when using powers
Pending input from broxen, but we should apply these variables that are taken from the associated power template (stored on MapServerData) when using powers.

This is related to #541 since it changes how we take in the power data from MapServerData, and in case I am unavailable to do this one, please don't work on this until that PR is done.

cyanpill:
Closes #546 .

**DataHelpers.cpp usePower()**

 
- PowerType, Range, Accuracy, and TargetType are taken from the power's data, and tested against the target.
- Started to implement damage and healing from powers, but it will need much more work. 
- The batman soundeffects are still there for powers that don't have attribute data.

**GameDataStore.cpp**

- Hardcoded values for some powers, to better test the above.

**power_definitions.h**

- Filled in some values from powers.bin

### DIFF
--- a/Projects/CoX/Common/GameData/GameDataStore.cpp
+++ b/Projects/CoX/Common/GameData/GameDataStore.cpp
@@ -417,6 +417,23 @@ bool GameDataStore::read_powers(const QString &directory_path)
     if (!read_data_to<AllPowerCategories, powers_i0_requiredCrc>(directory_path, "powers.bin",
                                                                    m_all_powers))
         return false;
+   StoredAttribMod temp; // just a bit of hardcoding stats to test powers
+   temp.name = "Damage";
+   temp.Magnitude = 5;
+   m_all_powers.m_categories[26].m_PowerSets[0].m_Powers[0].pAttribMod.push_back(temp);//brawl
+   m_all_powers.m_categories[32].m_PowerSets[1].m_Powers[0].pAttribMod.push_back(temp);
+    m_all_powers.m_categories[32].m_PowerSets[1].m_Powers[1].pAttribMod.push_back(temp);
+    m_all_powers.m_categories[30].m_PowerSets[1].m_Powers[0].pAttribMod.push_back(temp);
+    temp.name = "Knockback";
+    m_all_powers.m_categories[32].m_PowerSets[1].m_Powers[0].pAttribMod.push_back(temp);
+
+    temp.name = "Healing";
+    m_all_powers.m_categories[26].m_PowerSets[0].m_Powers[7].pAttribMod.push_back(temp);//rest
+    m_all_powers.m_categories[27].m_PowerSets[0].m_Powers[24].pAttribMod.push_back(temp);//medkit?
+    temp.name = "Speed_Boost";
+    temp.Magnitude = 0.5;
+    m_all_powers.m_categories[26].m_PowerSets[0].m_Powers[6].pAttribMod.push_back(temp);//sprint
+
     return true;
 }
 
--- a/Projects/CoX/Common/GameData/GameDataStore.cpp
+++ b/Projects/CoX/Common/GameData/GameDataStore.cpp
@@ -417,22 +417,23 @@ bool GameDataStore::read_powers(const QString &directory_path)
     if (!read_data_to<AllPowerCategories, powers_i0_requiredCrc>(directory_path, "powers.bin",
                                                                    m_all_powers))
         return false;
-   StoredAttribMod temp; // just a bit of hardcoding stats to test powers
-   temp.name = "Damage";
-   temp.Magnitude = 5;
-   m_all_powers.m_categories[26].m_PowerSets[0].m_Powers[0].pAttribMod.push_back(temp);//brawl
-   m_all_powers.m_categories[32].m_PowerSets[1].m_Powers[0].pAttribMod.push_back(temp);
-    m_all_powers.m_categories[32].m_PowerSets[1].m_Powers[1].pAttribMod.push_back(temp);
-    m_all_powers.m_categories[30].m_PowerSets[1].m_Powers[0].pAttribMod.push_back(temp);
-    temp.name = "Knockback";
-    m_all_powers.m_categories[32].m_PowerSets[1].m_Powers[0].pAttribMod.push_back(temp);
+
+    StoredAttribMod temp; // just a bit of hardcoding stats to test powers
+    temp.name = "Damage";
+    temp.Magnitude = 5;
+    Power_Data *temppower = lookUpPower(26,0,0);  //brawl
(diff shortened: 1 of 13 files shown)


broxen:
Let's try to merge this before my #598 as it'll probably be cleaner that way

cyanpill:
Ok, that way you can fix up anything that we don't catch here :smile:

Projects/CoX/Common/GameData/GameDataStore.cpp
```diff
@@ -417,6 +417,23 @@ bool GameDataStore::read_powers(const QString &directory_path)
     if (!read_data_to<AllPowerCategories, powers_i0_requiredCrc>(directory_path, "powers.bin",
                                                                    m_all_powers))
         return false;
+   StoredAttribMod temp; // just a bit of hardcoding stats to test powers
+   temp.name = "Damage";
+   temp.Magnitude = 5;
+   m_all_powers.m_categories[26].m_PowerSets[0].m_Powers[0].pAttribMod.push_back(temp);//brawl
```
nemerle: we really could use `lookUpPower(cat_idx,set_idx,power_idx)` at this point

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -576,22 +576,112 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, uint32_t tgt_idx
...
+            QString msg = "miss";
+            sendFloatingInfo(*ent.m_client, msg, FloatingInfoStyle::FloatingInfo_Info, 0.0);
+        return;
+        }
+
+    }    //todo check for auto hit
+ //   if (pdata.EffectArea == StoredAffectArea::Character)//single target so do the following once
+ //   else                                                // AOE has to check all valid targets, and do the following
+    if (pdata.pAttribMod.size() > 0)
+    {                       // effects done here
```
nemerle: check if ` !pdata.pAttribMod.empty()` the meaning is clearer and for some containers it will be faster.

Projects/CoX/Common/GameData/GameDataStore.cpp
broxen: We actually have `get_power_template()` for this purpose. Though it returns const. So we'd need to overload the method if you plan to add these values to the global gamedata array

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -576,22 +576,112 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, uint32_t tgt_idx
...
+
+    if(ppower != nullptr && !pdata.m_Name.isEmpty())
         ent.m_queued_powers.push_back(ppower);
 
+    dumpPower(*ppower);
+    QString msg;
+
+    if (pdata.Type != PowerType::Toggle && pdata.Type != PowerType::Click)      //treat toggle as clicks, ignore everything else for now
+        return;
+    Entity *target_ent = getEntity(ent.m_client, tgt_idx);
```
broxen: Unfortunately I didn't write getEntity very well and we need to check to ensure *target_ent != nullptr, or in edge cases you'll get client crashes. I recommend keeping:
```
if(target_ent == nullptr)
    {
        qCDebug(logPowers) << "Failed to find target:" << tgt_idx << tgt_id;
        return;
    }
```

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -576,22 +576,112 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, uint32_t tgt_idx
...
+
+    if (pdata.Range != float(0.0)) {               // self targetting doesn't need these checks
+        if(target_ent == nullptr)
```
broxen: I recommend moving this out of the if statement, so that it can catch a nullptr target_ent for the `sendFloatingInfo(*target_ent->m_client...` at the end of this method.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -576,22 +576,112 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, uint32_t tgt_idx
...
         ent.m_queued_powers.push_back(ppower);
 
+    dumpPower(*ppower);
+    QString msg;
+
+    if (pdata.Type != PowerType::Toggle && pdata.Type != PowerType::Click)      //treat toggle as clicks, ignore everything else for now
+        return;
+    Entity *target_ent = getEntity(ent.m_client, tgt_idx);
+
+    if (pdata.Range != float(0.0)) {               // self targetting doesn't need these checks
```
broxen: Please put brackets on their own line

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -576,22 +576,112 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, uint32_t tgt_idx
...
+            if (!target_ent->m_is_villian){
+                messageOutput(MessageChannel::COMBAT, QString("needs foe, targeting ally"), ent);
+                return;
+            }
+        }
+        else{
+            if (target_ent->m_is_villian){
+                messageOutput(MessageChannel::COMBAT, QString("needs ally, targeting foe"), ent);
+                return;
+                }
```
broxen: Again, please brackets on their own line

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -631,19 +711,17 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, uint32_t tgt_idx
 
     if(target_ent->m_type != EntType::PLAYER)
         return;
+    if(tgt_idx == ent.m_idx) // Skip the rest if targeting self.
```
broxen: nit: you could probably combine this with the if statement directly above:
```
if(target_ent->m_type != EntType::PLAYER || tgt_idx == ent.m_idx)
    return;
```

Projects/CoX/Servers/MapServer/DataHelpers.cpp
cyanpill: I was wondering why this was working without that, the only place usepower() is called from makes sure that it has a target, but that could change in the future.

broxen: Yeah unfortunately, the way I wrote getEntity isn't the best really. It works, but there's not a lot of failsafe built in.

PR #604, opened 2018-10-03
Testing cache theory on appveyor. Iterated version number

broxen:
Appveyor fails due to timeout. See if caching will fix that.

Also, iterate version number to v5, as we're cool like that


broxen:
Blargh.. that didn't work. Maybe because it only caches on success?

PR #606, opened 2018-10-04
Code clean-ups. Couple tiny bug fixes

broxen:
## Summary
Code clean ups and some little bug fixes for powers.

- Code Clean Ups
- Fixed name.compare() issue in usePower()
- Moved toggling of m_has_power_update and m_reset_powers to inside the methods that require them to simplify use. (useInspiration, removeInspiration, etc)


broxen:
I had a weird boot up bug with this (for seemingly no reason) but it disappeared after rebooting my computer. So maybe it's just my environment.

Can someone test this just to make sure? Thanks in advance

dracc:
I could test, but what kind of bug am I looking for?

cyanpill:
Hm, my client crashed a few times, but it could just be a coincidence. One time I saw in the (client)console something about a packet being the wrong size, but I didn't save it in time. On the server side, the only thing wrong is "Critical: Failed to serialize CharacterData, incompatible serialization format version  7"

broxen:
For wrong packet size I'll need to explore further, though it may relate to your second issue.

The second issue is actually a demonstration of one improvement from this PR. Previously the server would not have given you that message. You're getting it because a Character in your database had a db_version other than 7 for CharData schema. We'll need to check that CharData version is set to the correct value, and any Characters you have will have to be remade until we have a database update tool that can migrate versions.

cyanpill:
Deleted all characters, reloaded, created a new character and the first time logging in it locked up, but every time after that it was ok.

I did manage to copy the client console, but I notice it says this every time I connect to the server, so it's most likely a different issue:

```
Connecting to DbServer 192.168.0.10:7002 (UDP) cookie: c0a88a6d..ok   (0.02)
Connecting to mapserver 192.168.0.10:7004 (UDP) cookie: c0a8ddca..Error calling setsockopt, ending socket buffer size is not what we told it! (131072!=262144)
Error calling setsockopt, ending socket buffer size is not what we told it! (131072!=262144)
    (0.18)
```
Note this is in wine 3.03

PR #609, opened 2018-10-05
Added Network Ports on Line 152 in Readme for SEGS

Stitchified:
Closes # .

- 
- 
-

### DIFF
--- a/Projects/CoX/docs/README.md
+++ b/Projects/CoX/docs/README.md
@@ -149,7 +149,7 @@ Enjoy!
 HELP AND MORE INFORMATION
 ------
 
-Mae sure the following Network Ports are open: TCP 443, TCP 2104, TCP 2016, and UDP 7000-7200
+Make sure the following Network Ports are open: TCP 443, TCP 2104, TCP 2016, and UDP 7000-7200
 
 For Help with installation and configuration of your local SEGS, please see visit us on discord at https://discord.segs.io/
 
--- a/Projects/CoX/docs/README.md
+++ b/Projects/CoX/docs/README.md
@@ -149,6 +149,8 @@ Enjoy!
 HELP AND MORE INFORMATION
 ------
 
+Mae sure the following Network Ports are open: TCP 443, TCP 2104, TCP 2016, and UDP 7000-7200
+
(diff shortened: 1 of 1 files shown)


Projects/CoX/docs/README.md
```diff
@@ -149,6 +149,8 @@ Enjoy!
 HELP AND MORE INFORMATION
 ------
 
+Make sure the following Network Ports are open: TCP 443, TCP 2104, TCP 2016, and UDP 7000-7200
```
broxen: Great PR Stitch! Welcome to GitHub!

This is a much needed addition to the docs, so thank you for tackling this.

I think this section you added would be more useful under the `INSTALL & CONFIGURE THE COX CLIENT` header as a new numbered item.

```
2. **Point the CoX client towards your server IP**
// Blah blah

3. **Ensure proper ports are open on any firewalls**
// Your paragraph here

4. **How to correctly run the CoX Client?**
// Blah blah
```

dracc: I'd say it's rather server related than client though? 🤔

broxen: Yeah you're right dracc. We should leave this here under help where Stitch has put it, but also add it to the server setup section of the main readme.

Projects/CoX/docs/README.md
```diff
@@ -124,7 +124,10 @@ If you're looking to compile SEGS from source, or need to do an advanced manual
 
       SEGS requires several CoH files to run. Clicking this button will prompt you to enter the path to your CoX directory. Once selected, it will copy several files from the CoX directory, and extract a portion of those files to your `SEGS/data/` subdirectory. These `.bin` files are required by SEGS to load CoX maps, and spawn NPCs. Once completed, SEGSAdmin will clean up any unnecessary files.
 
-
+  4. **Ensure proper ports are open on any firewalls**
+      Make sure the following Network Ports are open: TCP 443, TCP 2104, TCP 2016, and UDP 7000-7200
+     These are the same ports that the original game client and now, SEGS uses as well.
+      
 4. **(Optional) Add User Account**
```
broxen: Since we've added a number 4, can we iterate this to 5.

Then you should be all set. Thanks!

PR #611, opened 2018-10-05
Added more Lua functions and LogScripts

ZopharShinta:
Added addNpc, faceEntity, faceLocation and printDebug to lua
printDebug calls qCDebug(logScripts) with the string to log.

Added logScripts checkbox to AdminTool as well


Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -174,11 +176,57 @@ void ScriptingEngine::registerTypes()
...
+        {
             FloatingInfoMsgKey f_info_message = static_cast<FloatingInfoMsgKey>(message_type);
             QString message = FloatingInfoMsg.find(f_info_message).value();
             cl->addCommand<FloatingInfo>(cl->m_ent->m_idx, message, FloatingInfo_Attention , 4.0);
+        },
+        "sendInfoMessage", [](MapClientSession *cl, int channel, const char* message)
+        {
+            sendInfoMessage(static_cast<MessageChannel>(channel), QString::fromUtf8(message), *cl);
         }
+    , // Error on line 165 prevents this from working. Not sure how to handle
```
broxen: which error?

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -26,6 +27,7 @@
 #include "Events/InfoMessageCmd.h"
 #include "Common/NetStructures/Entity.h"
 #include "Common/NetStructures/Contact.h"
+#include <iomanip>
```
broxen: It doesn't look like you're using this library anywhere, so let's not include it.

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
ZopharShinta: Old comment. Sorry. Removed

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -174,11 +175,57 @@ void ScriptingEngine::registerTypes()
...
+        }
+        int idx = npc_store.npc_idx(npc_def);
+        Entity *e = cl.m_current_map->m_entities.CreateNpc(getGameData(),*npc_def,idx,variation);
+
+        glm::vec3 loc;
+        const auto& floats = location.source;
+        int count = 0;
+        for (const auto& f : floats)
+        {
+            if(count == 0)
```
broxen: You for really simple one-line if/then/else statements like this, you can actually omit the brackets altogether:
`if(count == 0)
    loc.x = f;
else if(count == 1)
    loc.y = f;
else
    loc.z = f;

++count; // marginally better to iterate this way, over count++
```

File: Projects/CoX/Servers/MapServer/ScriptingEngine.cpp [hunk 4]
```diff
@@ -316,6 +362,35 @@ void ScriptingEngine::registerTypes()
...
+        sendFaceEntity(*cl->m_ent, target);
+    },
+    "faceLocation", [](MapClientSession *cl, sol::as_table_t<std::vector<std::float_t>> location)
+    {
+        glm::vec3 loc;
+        const auto& floats = location.source;
+        int count = 0;
+        for (const auto& f : floats)
+        {
+            if(count == 0)
```
broxen: same here as before

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -378,6 +464,29 @@ std::string ScriptingEngine::callFunc(const char *name, int arg1)
     return result.get<std::string>();
 }
 
+std::string ScriptingEngine::callFunc(const char *name, int arg1, glm::vec3 vec3)
```
broxen: Let's use QString over std::string whenever possible.

ZopharShinta: This is just an overload of the original callFunc to add the vec3.
Same for callFuncWithClientContext
We can change them all if it doesn't break the calls before.

broxen: Ahh I see. If you change them, change all of them. But I have a feeling that if Nem originally implemented those methods that way, it requires the std library for some reason.

broxen:
Maybe @nemerle can provide some insight into why this is failing CI.

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -16,6 +16,7 @@
 #include "DataHelpers.h"
 #include "ScriptingEngine.h"
 #include "MapClientSession.h"
+#include "MapInstance.h"
```
broxen: The linker might be complaining about this here. Do you need to include MapInstance? I don't see what you're using it for below.

ZopharShinta:
Another fail.
Should I change it from a std::vector<float> to 3 separate floats in the function signature or is there another option beside std::vector to store all 3 floats?

Projects/CoX/Servers/MapServer/ScriptingEngine_Null.cpp
```diff
@@ -48,6 +48,12 @@ std::string ScriptingEngine::callFunc(const char *name, int arg1)
     Q_UNUSED(arg1);
     return "";
 }
+std::string ScriptingEngine::callFunc(const char *name, int arg1, std::vector<float> loc)
+{
+    Q_UNUSED(name);
+    Q_UNUSED(arg1);
```
broxen: Make sure to add another Q_UNUSED macro for your new location argument

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -174,11 +175,51 @@ void ScriptingEngine::registerTypes()
...
             FloatingInfoMsgKey f_info_message = static_cast<FloatingInfoMsgKey>(message_type);
             QString message = FloatingInfoMsg.find(f_info_message).value();
             cl->addCommand<FloatingInfo>(cl->m_ent->m_idx, message, FloatingInfo_Attention , 4.0);
+        },
+        "sendInfoMessage", [](MapClientSession *cl, int channel, const char* message)
+        {
+            sendInfoMessage(static_cast<MessageChannel>(channel), QString::fromUtf8(message), *cl);
+        }
+    ,
+    "addNpc", [](MapClientSession &cl, const char* name, sol::as_table_t<std::vector<std::float_t>> location, int variation)
```
nemerle: Those lambdas are pretty big, if a lambda is more then 5 lines long, just make it a standalone function ?

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -174,11 +175,51 @@ void ScriptingEngine::registerTypes()
...
+        {
+            sendInfoMessage(MessageChannel::USER_ERROR, "No NPC definition for:"+npc_name, cl);
+            return;
         }
+        int idx = npc_store.npc_idx(npc_def);
+        Entity *e = cl.m_current_map->m_entities.CreateNpc(getGameData(),*npc_def,idx,variation);
+
+        glm::vec3 loc;
+        const auto& floats = location.source;
+        int count = 0;
```
nemerle: This is a strange way of doing this, can't we pass glm::vec3 as a parameter to this lambda ? Maybe we have to teach sol2 how to convert to it ?

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -378,6 +453,29 @@ std::string ScriptingEngine::callFunc(const char *name, int arg1)
     return result.get<std::string>();
 }
 
+std::string ScriptingEngine::callFunc(const char *name, int arg1, glm::vec3 loc)
+{
+    sol::protected_function funcwrap = m_private->m_lua[name];
+    funcwrap.error_handler = m_private->m_lua["ErrorHandler"];
+
+    sol::table loc_table = m_private->m_lua.create_table_with("x", loc.x, "y",  loc.y, "z", loc.z);
+    sol::table table = m_private->m_lua.create_table_with("id", arg1, "loc" , loc_table);
```
nemerle: if we can't teach sol2 to autoconvert our types by value, we can use the built-in support for pointers ?

ZopharShinta: Sol2 actually will send the glm::vec3 and Lua will read it. So this conversion wasn't needed.

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
ZopharShinta: Done

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
ZopharShinta: Had an issue with creating this functions within the scriptingEngine itself.
Broxen recommend adding them to DataHelpers.

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -213,109 +225,27 @@ void ScriptingEngine::registerTypes()
...
-        if(new_lvl != lvl){
-            cl->addCommand<FloatingInfo>(cl->m_ent->m_idx, FloatingInfoMsg.find(FloatingMsg_Leveled).value(), FloatingInfo_Attention , 4.0);
-            msg += " and LVL to " + QString::number(new_lvl);
-        }
-        qCDebug(logSlashCommand) << msg;
-        sendInfoMessage(MessageChannel::DEBUG_INFO, msg, *cl);
+        sendFaceEntity(*cl->m_ent, target);
     },
-    "sendFloatingDamage", [](MapClientSession *cl, int target, int amount)
+    "faceLocation",  [](MapClientSession *cl, glm::vec3 loc)
```
broxen: can't you simply `"faceLocation", sendFaceLocation` here, instead of using the lambda?

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -213,109 +225,27 @@ void ScriptingEngine::registerTypes()
...
-    },
-
-    "giveHp", [](MapClientSession *cl, float hp)
+    "giveEnd", giveEnd,
+    "giveHp", giveHp,
+    "giveInf",giveInf,
+    "giveInsp", giveInsp,
+    "giveXp", giveXp,
+    "sendFloatingDamage",sendFloatingNumbers,
+    "faceEntity",[](MapClientSession *cl, int target)
```
broxen: same here with `sendFaceEntity` the way you've used `sendFloatingNumbers` above

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -169,42 +180,43 @@ void ScriptingEngine::registerTypes()
...
-    {
-        uint32_t current_debt = getDebt(*cl->m_ent->m_char);
-        uint32_t debt_to_give = current_debt + debt;
-        setDebt(*cl->m_ent->m_char, debt_to_give);
-        QString msg = "Setting Debt to " + QString::number(debt_to_give);
-        qCDebug(logSlashCommand) << msg;
-        sendInfoMessage(MessageChannel::DEBUG_INFO, msg, *cl);
-    },
+    "giveDebt", giveDebt,
     "giveEnhancement",[](MapClientSession *cl, sol::as_table_t<std::vector<std::string>> enhancement)
```
broxen: Can we refactor this out as well? `"giveEnhancement", giveEnhancement`

Projects/CoX/Servers/MapServer/DataHelpers.h
```diff
@@ -149,3 +149,23 @@ void increaseLevel(Entity &ent);
...
+ */
+void addNpc(MapClientSession &cl, const char* name, glm::vec3 *loc, int variation);
+void giveDebt(MapClientSession *cl, int debt);
+void giveEnd(MapClientSession *cl, float end);
+void giveHp(MapClientSession *cl, float hp);
+void giveInf(MapClientSession *cl, int inf);
+void giveInsp(MapClientSession *cl, const char *value);
+void giveXp(MapClientSession *cl, int xp);
+
+
```
broxen: We can remove all these blank lines

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -872,4 +872,134 @@ void findTeamMember(Entity &tgt)
...
+    sendInfoMessage(MessageChannel::DEBUG_INFO, QString("Created npc with ent idx:%1 at location x: %2 y: %3 z: %4").arg(e->m_idx).arg(loc->x).arg(loc->y).arg(loc->z), cl);
+}
+
+void giveDebt(MapClientSession *cl, int debt)
+{
+    uint32_t current_debt = getDebt(*cl->m_ent->m_char);
+    uint32_t debt_to_give = current_debt + debt;
+    setDebt(*cl->m_ent->m_char, debt_to_give);
+    QString msg = "Setting Debt to " + QString::number(debt_to_give);
+    qCDebug(logSlashCommand) << msg;
```
broxen: Let's change all these categories from `logSlashCommand` to `logScripts` as we're calling these methods from the scripts only.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -872,4 +872,134 @@ void findTeamMember(Entity &tgt)
...
+    CharacterData &cd = cl->m_ent->m_char->m_char_data;
+    QString val = QString::fromUtf8(value);
+    QString msg = "You do not have room for any more inspirations!";
+
+    if(getNumberInspirations(cd) < getMaxNumberInspirations(cd))
+    {
+        msg = "Awarding Inspiration '" + val + "' to " + cl->m_ent->name();
+
+        addInspirationByName(cd, val);
+        cd.m_has_updated_powers = true;
```
broxen: You can remove this line, it's already called within the function preceeding it.

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -213,109 +225,27 @@ void ScriptingEngine::registerTypes()
         if(getNumberEnhancements(cd) < 10)
         {
             msg = "Awarding Enhancement '" + name + "' to " + cl->m_ent->name();
-
             addEnhancementByName(cd, name, level);
-            cd.m_powers_updated = true;
-
+            cd.m_has_updated_powers = true;
```
broxen: You can remove this line, it's already called within the function preceeding it.

PR #612, opened 2018-10-06
SEGSAdmin - Settings & Logging Update

### LINKED ISSUE #466: Change SEGSAdmin code to use settings_template.cfg on settings.cfg generation
On discussion with @nemerle and @dracc in Discord, it would be a nice to have for SEGSAdmin and other components that generate a settings.cfg file to utilise a single central function to do this, rather than use duplicate functions in different areas which perform the same actions. 

Make a change to SEGSAdmin to utilise settings_template.cfg to generate the settings.cfg file on user initial setup.

### LINKED ISSUE #564: Programatically add logging categories to SEGSAdmin
## Summary
Currently we have to add new log categories to the settings.cfg manually. Perhaps we can add them programatically by reading from Logging.h, or by reading from the settings_template.cfg

Alternatively, we can continue adding manually with new releases.

### LINKED ISSUE #607: Inform users that they don't need all logs enabled at all times
## Summary
Discussion in discord demonstrated that some users run their servers with all logs toggled on all the time. While this probably isn't harmful, it does produce hundreds of thousands of lines of log output on any given server, and makes troubleshooting issues difficult.

### Proposal
We should add a message to SEGSAdmin's logs window letting users know that log messages only need to be set to true for debugging specific issues. We can enable logMapEvents and logSlashCommands by default, if we must, but the rest can be false.

miles200:
- Logging categories now read from settings_template.cfg
- Add log_mapxfers to settings.cpp
- Add new map names to SetUpData for map directories
- Add AFK and StartingCharacter to SEGSAdmin Settings
- Add info message to Logging tab advising users that usually used for
  debugging

Closes #564 
Closes #466 
Closes #607


miles200:
Cheers @cyanpill

I’ll remove these from default generation, do you want me to update
settings_template.cfg too? They were in there

Thanks
Miles

cyanpill:
> Cheers @cyanpill I’ll remove these from default generation, do you want me to update settings_template.cfg too? They were in there Thanks Miles

hm, shouldn't be there, you might not have updated? https://github.com/Segs/Segs/blob/develop/Projects/CoX/Data/settings_template.cfg

miles200:
Ah must be on an older version. Okay will update cheers!

broxen:
Pending the changes Cyanpill recommended, I think this is ready for merge.

miles200:
@broxen I think this is good to go now pending CI clearing.

PR #631, opened 2018-10-14
log: Put 'duplicate mod trick' in 'SceneGraph' category

dracc:
As title says, move the message ` debug: duplicate model trick!` into a log category.
Said logging category is a new one.


broxen:
:heart: thanks for this dracc!

I recommend we create a new category for loading the SceneGraph, maybe `logSceneGraph`

You can use one of the existing logging categories as an example.

PR #632, opened 2018-10-14
StateStorage, MotionState and Velocity

broxen:
## Summary
Relates to #78 . Specifically this defines a StateStorage for keeping track of InputStates received from the client. It defines a MotionState that stores current derived motion related values for entities. And it eliminates the need for the the movement workaround that was implemented some months back.

- Define StateStorage
- Add incoming inputstates to state storage. Purge states when the vector is larger than 30 (we may need less than this even)
- Add MotionState to entity for storing the derived state at any given time, including flags for movement related effects, such as flying or sliding
- Add setVelocity()
- NOTE: this doesn't impact (improve) movement in any noticeable way, but aims to improve the infrastructure around movement and reduce the size and complexity of my Movement PR #580 to make it more clear what's happening there, and (hopefully) simplify the review process.


broxen:
The meat of this is in Movement.cpp and StateStorage.cpp, the rest is mostly variable renames, etc.

PR #634, opened 2018-10-15
Moving methods to EntityHelpers.cpp/h and cleaned up DataHelpers

broxen:
## Summary
As a result of a discussion in Discord with Nem. We really needed to move some methods to an EntityHelper file in NetStructures.

This one is going to be painful, because it touches a lot of files. It's something that we sort of needed though.

- Moved Entity helpers to their own file
- Cleaned up DataHelpers and it's methods to use MapClientSession where necessary
- Removed a couple duplicate methods
- Removed assert(false) from Team.cpp in favor of a qWarning() to fail gracefully.


Projects/CoX/Common/NetStructures/Entity.cpp
```diff
@@ -188,50 +182,4 @@ void initializeNewNpcEntity(const GameDataStore &data, Entity &e, const Parse_NP
     e.m_char->m_char_data.m_level       = src->m_Level;
 }
```
nemerle: Is there a reason that this was left in here ? :)

Projects/CoX/Common/NetStructures/Trade.cpp
```diff
@@ -31,7 +32,7 @@ static bool hasRoomForTradeInfluence(Entity& src)
     const uint32_t inf = (getInf(*src.m_char) - info_src.m_influence) + info_tgt.m_influence;
     if (inf < info_tgt.m_influence) // Standard overflow check (only works with unsigned integers).
     {
-        messageOutput(MessageChannel::SERVER, "Not enough room to accept that much influence.", src);
+        sendInfoMessage(MessageChannel::SERVER, "Not enough room to accept that much influence.", *src.m_client);
```
nemerle: the code here is quite problematic, since it attempts to send messages, and that's part of MapInstance 
`NetStructures` should not depend on MapInstance.
What we'd likely want here, is functions returning error codes, that callers check, and that send messages accordingly.

That's for the future, though :)

Projects/CoX/Common/NetStructures/Trade.cpp
```diff
@@ -173,7 +166,7 @@ static void finishTrade(Entity& src, Entity& tgt)
     src.m_trade.reset();
     tgt.m_trade.reset();
 
-    sendTradeSuccess(src, tgt);
+    sendTradeSuccess(*src.m_client, *tgt.m_client);
```
nemerle: Once again, we should sendTradeSuccess in the caller of this function.

Projects/CoX/Common/NetStructures/Entity.cpp
broxen: My thought was that we may never need to call it outside of entity and we could make these methods private. Thoughts?

Projects/CoX/Common/NetStructures/Trade.cpp
broxen: Yes. Trade is going to need a rewrite. It's going to be pretty extensive, since it utilizes a lot of messages and addCommand wrappers.

broxen:
Ok, I've refactored anything that relies on MapInstance out of Trade System, and I moved the other Entity methods to EntityHelpers (we can make them private later, if better).

The only thing I wasn't sure how to resolve was the EmailHeaderRequest, EmailReadMessage etc errors when you ENABLE_TESTS.

Any insight into that issue @nemerle ? You mentioned a work-around in Discord, but then also said we may not need to serialize those methods anyway. Should we aim to merge this and resolve that problem separately?

Let me know your thoughts.

PR #636, opened 2018-10-16
Dooranims

broxen:
Relates to #519 

- DoorAnimStart
- DoorAnimExit
- Send RUNIN seqbit


broxen:
Thanks for catching that @cyanpill 

I must have noticed it right as you did

PR #642, opened 2018-10-19
[WIP] Standalone Team Handler

### LINKED ISSUE #555: Send Team-removal and Sidekick-removal messages
**aka: Logging off while on Team, or leaving a Team while Sidekicked asserts false**

## Summary
Logging off while on Team, or leaving a Team while Sidekicked is hard coded to assert(false) because we have no method of accessing other mapserver's entity lists. We need to develop a Team-Removal event, and Sidekick Removal event to be sent to entities when these methods fire in DataHelpers.cpp

DataHelpers.cpp: `removeSidekick()`
```cpp
    assert(false);
    //TODO: this function should actually post messages related to de-sidekicking to our target entity.
    Entity      *tgt            = nullptr; //getEntityByDBID(src_sk.m_db_id);
```
DataHelpers.cpp: `removeTeamMember()`
```cpp
    assert(false);
    // TODO: this should post an Team-removal event to the target entity, since we can't access other server's
    // Entity lists
    Entity *tgt = nullptr; //getEntityByDBID(idx);
```

### Note:
I'd like to get this in before v0.6.0 as it creates an unfriendly user-experience that may result in support-requests from less technical users

crylico:
Closes #555.

- [ ] Route slash commands, connection events, etc. to a contained TeamHandler
- [ ] Extract required Entity IDs from incoming events to centrally manage team relationships
- [ ] Interface with MapInstance to populate Client GUI Windows


Projects/CoX/Servers/GameServer/GameServer.cpp
```diff
@@ -70,8 +72,11 @@ class GameServer::PrivateData
     {
         // tell our handler to shut down
         shutdown_event_processor_and_wait(m_handler);
-        // tell our friendship service to close too
+
+        // tell our child services to close too
+        shutdown_event_processor_and_wait(m_email_service.get());
```
broxen: Good catch!

broxen:
This is a great start! If we can manage to sneak this into v6, great! If not, no worries, we'll push out to v7.

Really appreciate your work @crylico

crylico:
I'm really hoping to get everything in for v0.6, too! If you could take a look at the latest changes, specifically the reworking of CMake stuff for services/handlers that can link against multiple different servers, I'd really appreciate your feedback :)

crylico:
Overview of Interaction between MapInstance slash commands OR zoning/power/disconnect events => Team Handler manages teams, publishes changes => MapInstance updates all changes entities

1) slash commands will extract metadata from the MapInstance required for the team window and save them in the team handler
2) When any slash commands or new data affect the team (change leader status, kick / join), create a message with any ID's that are no longer in a team AND the new copy of the team if it still exists
3) Iterate through all MapInstances using the HandlerLocator and send this message to all of them
4) MapInstances will make sure all current entities involved in the message reflect the team record, including metadata from step 1 in case the entities are on different MapInstances
5) Make sure team handler gets notified about any disconnects, changed active powers, player zone changes, etc - send message as in step 2

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -1823,40 +1797,44 @@ void cmdHandler_Train(const QString &/*cmd*/, MapClientSession &sess)
...
+{    
     // game command: "team_accept \"From\" to_db_id to_db_id \"To\""
 
     QString msgfrom = "Something went wrong with TeamAccept.";
     QString msgtgt = "Something went wrong with TeamAccept.";
     QStringList args;
     args = cmd.split(QRegularExpression("\"?( |$)(?=(([^\"]*\"){2})*[^\"]*$)\"?")); // regex wizardry
 
-    QString from_name       = args.value(1);
+    // QString from_name       = args.value(1);
```
broxen: My suspicion is that the Client sends char_names alongside db_ids here to verify and prevent a malicious user from executing random `/team_accept MyGuy 10 10 YourGuy` cmds.

Projects/CoX/Servers/MapServer/MapServer.cpp
```diff
@@ -168,11 +172,93 @@ void MapServer::dispatch(Event *ev)
...
+    }
+
+    return nullptr;
+}
+
+void MapServer::route_opaque_message(UserRouterMessage *msg) 
+{
+    MapInstance *ins = getMapInstanceForEntityID(msg->m_data.m_target_id);
+
+    if (ins == nullptr) {
```
broxen: Thank you for all of your work on this task (and messagebus)!

Please put brackets on their own line, per our [Coding Style Guide](https://github.com/Segs/Segs/blob/develop/docs/CONTRIBUTING.md#coding-styleguide)

crylico: Will address all these before merge :)

crylico:
1. [SlashCommand => TeamMemberInvitedMessage](https://github.com/crylico/Segs/blob/cf13769d70a285c4e7939abcc5ac159efc21a72b/Projects/CoX/Servers/MapServer/SlashCommand.cpp#L1497)
2. [Handle TeamMemberInvitedMessage](https://github.com/crylico/Segs/blob/cf13769d70a285c4e7939abcc5ac159efc21a72b/Projects/CoX/Servers/GameServer/TeamService/TeamHandler.cpp#L36)
3. [Route TeamMemberInvitedMessage via MapServer](https://github.com/crylico/Segs/blob/cf13769d70a285c4e7939abcc5ac159efc21a72b/Projects/CoX/Servers/GameServer/TeamService/TeamHandler.cpp#L87)
4. [Inject target_id via target_name](https://github.com/crylico/Segs/blob/cf13769d70a285c4e7939abcc5ac159efc21a72b/Projects/CoX/Servers/MapServer/MapServer.cpp#L223)
5. [Route Message to correct MapInstance](https://github.com/crylico/Segs/blob/cf13769d70a285c4e7939abcc5ac159efc21a72b/Projects/CoX/Servers/MapServer/MapServer.cpp#L244)
6. [Present TeamOffer to target MapInstance](https://github.com/crylico/Segs/blob/cf13769d70a285c4e7939abcc5ac159efc21a72b/Projects/CoX/Servers/MapServer/MapInstance.cpp#L2681)

Projects/CoX/Servers/GameServer/TeamService/TeamHandler.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+	std::stringstream stream;
+    to_storage(stream, msg);
+	size_t s = size_of_stream(stream);
+
+	char buf[1024];
+	assert(s <= sizeof(buf));
+
+	stream.read(buf, s);
+	QByteArray payload(buf, s);
```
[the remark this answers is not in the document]
crylico: lol I'll figure out the stdlib + Qt5 sooner or later - this looks a lot simpler than the solution I threw together :)

Projects/CoX/Servers/MapServer/SlashCommand.cpp
crylico: I realized while implementing the `/invite` command that I needed the name explicity so that dialog messages don't require a conversion from ID to Name by the MapServer - maybe I'll find it useful when implementing on this message too :) Thanks for this insight

crylico:
@nemerle had to do some really dodgy stuff with the underlying values for UserRouterService and TeamService enums. Can you take a look and let me know if you have a better idea in mind? Summary: Custom handlers need to have Unique enum values so that any service can handle any kind of event (for example, the TeamHandler needs to handle TeamInvite events AND UserRouterQueryRequest/Response events, so they can't have the same underlying enum value)

nemerle:
> @nemerle had to do some really dodgy stuff with the underlying values for UserRouterService and TeamService enums. Can you take a look and let me know if you have a better idea in mind? Summary: Custom handlers need to have Unique enum values so that any service can handle any kind of event (for example, the TeamHandler needs to handle TeamInvite events AND UserRouterQueryRequest/Response events, so they can't have the same underlying enum value)

Yup I've been thinking about discarding type enums, and using compile-time hashed strings, so that the switches would look like:
```cpp
switch(val)
{
    case ev_id("TeamInvite"):
}
```
The only problem with this ( apart from readability ), is the potential performance hit due to double indirection in switches. Sadly this does not solve the symbol/hash clashes. Possible way to check against that is generating a synthetic cpp file that contains a switch with all event types as cases, this way compiler will complain if it finds any kind of clash.

broxen:
Awesome progress @crylico!

broxen:
Just a note: spoke to crylico on discord about a week ago. This is wrapping up.

Crylico, just let me know when you're ready for testing and review.

PR #643, opened 2018-10-19
Add in Contact Lists / Expose Contacts to Lua

ZopharShinta:
This adds in basic functions for the Contact Lists.
Lua can add/update the contact list as well.


Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -144,6 +144,32 @@ void Character::addStartingInspirations(QStringList &starting_insps)
...
+{
+    Contact startingContact;
+    startingContact.setName(contact_name); // "OfficerFlint
+    startingContact.m_current_standing = 1;
+    startingContact.m_notify_player = true;
+    //startingContact.m_display_name = "Officer Flint";
+    startingContact.m_task_index = 1;
+    startingContact.m_npc_id = 1939; // Npc Id
+    startingContact.m_has_location = true;
+    //startingContact.m_handle = 1; // ?
```
broxen: handle is likely `m_contact_idx; // contact's idx in player's list of contacts`

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -144,6 +144,32 @@ void Character::addStartingInspirations(QStringList &starting_insps)
         addInspirationByName(m_char_data, name);
 }
 
+void Character::addStartingContact(const char *contact_name)
+{
+    Contact startingContact;
+    startingContact.setName(contact_name); // "OfficerFlint
+    startingContact.m_current_standing = 1;
+    startingContact.m_notify_player = true;
+    //startingContact.m_display_name = "Officer Flint";
```
broxen: contact has no member `m_display_name`, is this planned for future use? or?

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -144,6 +144,32 @@ void Character::addStartingInspirations(QStringList &starting_insps)
...
+    startingContact.m_task_index = 1;
+    startingContact.m_npc_id = 1939; // Npc Id
+    startingContact.m_has_location = true;
+    //startingContact.m_handle = 1; // ?
+    startingContact.m_location_description = "Outbreak";
+    //startingContact.m_location_name = "Outbreak Starting";
```
broxen: Contact object should only have one location descriptor. This is probably in reference to the member of Destination m_location_name, which appears to (also) be the Zone name.

Same with m_location_map_name below this, which actually should be a member of Destination struct.

Projects/CoX/Common/NetStructures/Character.cpp
```diff
@@ -144,6 +144,32 @@ void Character::addStartingInspirations(QStringList &starting_insps)
         addInspirationByName(m_char_data, name);
 }
 
+void Character::addStartingContact(const char *contact_name)
```
broxen: This is great for now!

Eventually we're going to watch to make this method generic, and have it automatically fill these values from the entity making itself a contact to the player.

This could be done through scripts by returning a specific value to the on_entity_interact() handler in MapInstance. We can get the Contact's specs from within that handler.

Projects/CoX/Common/NetStructures/Contact.h
```diff
@@ -38,16 +44,91 @@ static QHash<QString,int> contactLinkHash = {
...
+    static const constexpr uint32_t class_version = 1;
+
     QString m_name;
     QString m_display_name;
 
     // for scripting language access.
     std::string getName() const { return m_name.toStdString();}
     void setName(const char *n) { m_name = n; }
+
+    std::string getLocationDescription() const { return m_location_description.toStdString();}
```
broxen: We should return QString here.

Projects/CoX/Common/NetStructures/Character.cpp
ZopharShinta: This was for early testing. The current scripts actually create the contact on the entity_interact/accept mission.
Once we get the waypoint and tasks working I could see the script setting the first contact on spawning in or something.

Projects/CoX/Common/NetStructures/Character.cpp
ZopharShinta: This was old code that was already in the contact class. I can remove it since it doesn't look like its needed.

Projects/CoX/Common/NetStructures/Character.cpp
ZopharShinta: I'll remove them. There seemed to be a few redundant properties in the class.

Projects/CoX/Common/NetStructures/Contact.h
```diff
@@ -7,7 +7,13 @@
 
 #pragma once
 
+
 #include <QString>
```
broxen: You can simplify these includes:
```
#include "CommonNetStructures.h"
#include <glm/vec3.hpp>
#include <cereal/cereal.hpp>
```

Projects/CoX/Common/NetStructures/Contact.h
```diff
@@ -38,16 +44,91 @@ static QHash<QString,int> contactLinkHash = {
...
+       uint8_t          m_confidant_threshold;           // int for confidant status?
+       uint8_t          m_friend_threshold;              // int for friend status?
+       uint8_t          m_complete_threshold;            // int for complete status?
+       uint8_t         m_task_index            = 0;
+       bool            m_notify_player         = false;
+       bool            m_can_use_cell          = false;
+       bool            m_has_location          = false;
+       Destination     m_location;
+
+       QString         m_location_name;
```
broxen: This and `m_location_map_name` should be members of Destination struct.

Projects/CoX/Common/GameData/chardata_definitions.h
```diff
@@ -28,9 +29,11 @@ static const constexpr  uint32_t        class_version   = 1;
                         SidekickType    m_type          = NoSidekick;
 };
 
+using vContactList = std::vector<Contact>;
```
broxen: You can move this to `Contact.h` (after the declaration of Contact struct) where it'll be accessible by anything that needs to use it.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -647,6 +648,34 @@ void sendContactDialogClose(MapClientSession &src)
...
+        }
+    }
+
+    if(!found)
+        contacts.push_back(contact);
+
+    //update database contactList
+    src.m_ent->m_char->m_char_data.m_contacts = contacts;
+
+    auto val = src.m_current_map->m_scripting_interface->callFunc("contact_list",src.m_ent->m_char->m_char_data.m_contacts);
```
broxen: What's the purpose of this here? Does this update the scripting interface with the new contact list?

Since we don't use the return value `val` anywhere, add a macro to ignore it on the next line:
`Q_UNUSED(val);`

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -647,6 +648,34 @@ void sendContactDialogClose(MapClientSession &src)
...
+
+    //Send contactList to client
+    src.m_ent->m_client->addCommandToSendNextUpdate(std::unique_ptr<ContactStatusList>(new ContactStatusList(contact)));
```
broxen: you can shorten this to `src.addCommand<ContactStatusList>(contact);` to make it much more readable.

Projects/CoX/Servers/MapServer/DataHelpers.h
```diff
@@ -8,6 +8,7 @@
 #pragma once
 #include "Events/MessageChannels.h"
 #include "Events/FloatingInfoStyles.h"
+#include "NetStructures/Contact.h"
```
broxen: No need to include this here. Just forward declare `class Contact;` below the includes.

Projects/CoX/Servers/MapServer/Events/ContactList.h
```diff
@@ -0,0 +1,112 @@
...
+
+namespace SEGSEvents
+{
+    // [[ev_def:type]]
+    class ContactStatusList : public GameCommandEvent
+    {
+
+    public:
+        // [[ev_def:field]
+        std::vector<Contact> m_contact_list;
```
broxen: leverage your vContactList definition here `vContactList m_contact_list;`

Projects/CoX/Servers/MapServer/Events/ContactList.h
```diff
@@ -0,0 +1,112 @@
...
+        explicit ContactStatusList() : GameCommandEvent(MapEventTypes::evContactStatusList){}
+        ContactStatusList(Contact contact) : GameCommandEvent(MapEventTypes::evContactStatusList)
+        {
+            m_contact_list.push_back(contact);
```
broxen: You actually want to send the entire array, not just add one contact to it. This section should be changed to accept vContactList as an argument, and elminate the need for push_back.

Projects/CoX/Servers/MapServer/Events/ContactList.h
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        void serializeto(BitStream &bs) const override
+        {
+            bs.StorePackedBits(1, type()-evFirstServerToClient); // 43
```
broxen: nitpick: this is probably a result of my bad habit in other events, but we really should comment these as `// packet 43` so it's clear what 43 means

Projects/CoX/Servers/MapServer/Events/ContactList.h
```diff
@@ -0,0 +1,112 @@
...
+            bs.StorePackedBits(1, m_contact_list.size()); // svr_contacts_total_also_maybe
```
broxen: This is odd. The client side looks something like this though, so maybe you're right and we just send size twice:
```
int svr_contacts_total_also_maybe = my_pakGetPackedBits(pak, 1);
    if ( svr_contacts_total_also_maybe < 0 )
        cl_contacts_total = g_contacts_arr.size();
    else
        cl_contacts_total = svr_contacts_total_also_maybe + 1;

    while(g_contacts_arr.size() > cl_contacts_total)
    {
        ContactStatusDestroy(g_contacts_arr.back());
        g_contacts_arr.pop_back();
    }

    while (g_contacts_arr.size() < cl_contacts_total)
        g_contacts_arr.emplace_back(ContactStatusCreate());

    int svr_contacts_total = my_pakGetPackedBits(pak, 1);
```

Projects/CoX/Servers/MapServer/Events/ContactList.h
```diff
@@ -0,0 +1,112 @@
...
+            bs.StorePackedBits(1, m_contact_list.size()); // svr_contacts_total
+
+            int count = 0;
+            for(const Contact &contact : m_contact_list)
+            {
+                bs.StorePackedBits(1, count);
+                bs.StoreString(contact.m_name);
+                bs.StoreString(contact.m_location_description);
+                bs.StorePackedBits(12, contact.m_npc_id);
+                bs.StorePackedBits(1, contact.m_handle);
```
broxen: this again appears to simply be the same value as your `count` variable. It's just the idx of this contact in the contactlist array.

Projects/CoX/Servers/MapServer/Events/ContactList.h
```diff
@@ -0,0 +1,112 @@
...
+                bs.StorePackedBits(1, contact.m_current_standing);
+                bs.StorePackedBits(1, contact.m_friend_threshold);
+                bs.StorePackedBits(1, contact.m_confidant_threshold);
+                bs.StorePackedBits(1, contact.m_complete_threshold);
+                bs.StoreBits(1, contact.m_has_location);
+
+                bs.StoreFloat(contact.m_location.location.x);
+                bs.StoreFloat(contact.m_location.location.y);
+                bs.StoreFloat(contact.m_location.location.z);
+                bs.StoreString(contact.m_location_map_name);
```
broxen: As I mentioned above, you can make these members of the Destination struct.

Projects/CoX/Servers/MapServer/Events/ContactList.h
```diff
@@ -0,0 +1,112 @@
...
+                bs.StoreString(contact.m_location_name);
+                ++count;
+            }
+            qCDebug(logMapEvents) << "ContactStatusList Event serializeto";
+        }
+
+        void serializefrom(BitStream &bs)
```
broxen: This will never be used, you can actually remove it.

Projects/CoX/Servers/MapServer/Events/ContactList.h
```diff
@@ -0,0 +1,112 @@
...
+
+     // [[ev_def:type]]
+    class ContactSelect : public MapLinkEvent
+    {
+    public:
+        explicit ContactSelect() : MapLinkEvent(MapEventTypes::evContactSelect){}
+
+        void serializeto(BitStream &bs) const override
+        {
+            bs.StorePackedBits(1, type()-evFirstServerToClient); // 44?
```
broxen: This is packet 44, you're correct.

You need to send an idx of the contact in the contactlist array:
```
void  ContactSelectReceive(DbPacket *pak)
{
    int idx  = my_pakGetPackedBits(pak, 1);
    ContactSelect(idx, 0);
}
```

broxen: As far as I can tell `ContactSelect()` does not lead to anything that requires sent packets.

That said, it does rely on there being a TaskList, which we currently do not send. I would test this to see if we get any crashes. If we do, we might need to send all contacts with `m_task_idx` as `0` until we implement Tasks fully.

Projects/CoX/Servers/MapServer/Events/ContactList.h
```diff
@@ -0,0 +1,112 @@
...
+
+        }
+        void    serializefrom(BitStream &bs)
```
broxen: Unneeded, you can remove.

Projects/CoX/Servers/MapServer/Events/ContactList.h
```diff
@@ -0,0 +1,112 @@
...
+
+    public:
+        // [[ev_def:field]]
+        int32_t m_srv_idx = 0;
+        explicit ReceiveContactStatus() : MapLinkEvent(MapEventTypes::evReceiveContactStatus){}
+
+        void serializeto(BitStream &bs) const override
+        {
+            bs.StorePackedBits(1, type()-evFirstServerToClient); // 26
+            bs.StorePackedBits(12, m_srv_idx);
```
broxen: this should never fire, so we can remove this line and add `assert(!"ReceiveContactStatus serializeto");` just in case.

Projects/CoX/Servers/MapServer/Events/MapEventTypes.h
```diff
@@ -62,8 +62,8 @@ enum MapEventTypes
     EVENT_DECL(MapEventTypes, evContactDialogClose         ,140)
     EVENT_DECL(MapEventTypes, evContactDialogOk            ,141)
     EVENT_DECL(MapEventTypes, evContactDialogYesNo         ,142)
-//    EVENT_DECL(MapEventTypes, evContactStatusList        ,143)
-//    EVENT_DECL(MapEventTypes, evContactSelect            ,144)
+    EVENT_DECL(MapEventTypes, evContactStatusList        ,143)
```
broxen: Nit: line up the column for numbers please =P

Projects/CoX/Servers/MapServer/Events/MapEventTypes.h
```diff
@@ -91,6 +91,7 @@ enum MapEventTypes
     EVENT_DECL(MapEventTypes, evBrowser                    ,176)
     EVENT_DECL(MapEventTypes, evTrayAdd                    ,182)
     EVENT_DECL(MapEventTypes, evCombineEnhanceResponse     ,183)
+    EVENT_DECL(MapEventTypes, evReceiveContactStatus       ,184)
```
broxen: Since this event is received from the client (client->server event, aka MapLinkEvent) it has a prefix of `200` instead of 100. So this should be `284` and should move down into the section below.

Projects/CoX/Common/NetStructures/Contact.h
ZopharShinta: These methods are just to set/get the string in Lua. 
We current don't have QString implemented there.
I started on it but ran into issues and walked away from it.
I'll try to look at it after I correct the other things mentioned.

Projects/CoX/Common/NetStructures/Contact.h
ZopharShinta: There are 3 errors when its just those includes

line 16: `implicit instantiation of undefined template 'QHash<QString, int>'`
line 60 & line 104: `calling 'critical' with incomplete return type 'QDebug'`
line 143: `implicit instantiation of undefined template 'QVector<ContactEntry>'`

if I just add `#include "Logging.h"` the errors go away.
Let me know if that is enough or if there's something else I need to do.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
ZopharShinta: That was the thought but I still had issue with multiple clients. 
I'll remove it.

Projects/CoX/Servers/MapServer/Events/ContactList.h
ZopharShinta: This explains the bug I was seeing. Where Flint was replaced with Parks

Projects/CoX/Servers/MapServer/Events/ContactList.h
ZopharShinta: Yeah. The client was crashing or giving the null gull with it missing.

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -224,7 +243,7 @@ void ScriptingEngine::registerTypes()
         "set_default_spawn_point", &MapSceneGraph::set_default_spawn_point
     );
     m_private->m_lua.script("function ErrorHandler(msg) return \"Lua call error:\"..msg end");
-    m_private->m_lua["contactTable"] = m_private->m_lua.create_named_table("contactTable");
+    m_private->m_lua["characters"] = m_private->m_lua.create_named_table("characters");
```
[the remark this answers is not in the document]
ZopharShinta: More testing for multiple client states. Both are not needed. So I'll remove them both.

Projects/CoX/Common/NetStructures/Contact.h
broxen: Got it!

Projects/CoX/Common/NetStructures/Contact.h
```diff
@@ -109,10 +107,9 @@ class Contact
            }
 
        archive(cereal::make_nvp("Name",m_name));
-       archive(cereal::make_nvp("DisplayName",m_display_name));
        archive(cereal::make_nvp("LocationDescription",m_location_description));
        archive(cereal::make_nvp("npcId",m_npc_id));
-       archive(cereal::make_nvp("Handle",m_handle));
+       archive(cereal::make_nvp("Handle",m_contact_idx));
```
broxen: You can change the string "Handle" to better reflect it's purpose.

Projects/CoX/Servers/MapServer/Events/ContactList.h
```diff
@@ -52,36 +52,35 @@ namespace SEGSEvents
...
         EVENT_IMPL(ContactStatusList)
     };
 
      // [[ev_def:type]]
-    class ContactSelect : public MapLinkEvent
+    class ContactSelect : public GameCommandEvent
     {
     public:
-        explicit ContactSelect() : MapLinkEvent(MapEventTypes::evContactSelect){}
+        Contact selected_contact;
```
broxen: Please add `// [[ev_def:field]]` directly above this variable declaration. This informs the unit testing handler of this member variable.

Projects/CoX/Servers/MapServer/Events/GameCommandList.h
```diff
@@ -0,0 +1,39 @@
+/*
+ * SEGS - Super Entity Game Server
+ * http://www.segs.io/
+ * Copyright (c) 2006 - 2018 SEGS Team (see AUTHORS.md)
+ * This software is licensed under the terms of the 3-clause BSD License. See LICENSE.md for details.
+ */
+
+#pragma once
+
```
broxen: This entire file was removed. You can delete it.

Projects/CoX/Common/GameData/chardata_definitions.h
```diff
@@ -0,0 +1,76 @@
+/*
+ * SEGS - Super Entity Game Server
```
broxen: This entire file has been renamed CharacterData.cpp

Please cross reference your changes against that file and delete this one altogether

Projects/CoX/Common/NetStructures/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+${CMAKE_CURRENT_SOURCE_DIR}/Character.cpp
+${CMAKE_CURRENT_SOURCE_DIR}/CharacterHelpers.cpp
+${CMAKE_CURRENT_SOURCE_DIR}/CommonNetStructures.cpp
+${CMAKE_CURRENT_SOURCE_DIR}/Contact.cpp
+${CMAKE_CURRENT_SOURCE_DIR}/Costume.cpp
+${CMAKE_CURRENT_SOURCE_DIR}/Entity.cpp
+${CMAKE_CURRENT_SOURCE_DIR}/Friend.cpp
+${CMAKE_CURRENT_SOURCE_DIR}/LFG.cpp
+${CMAKE_CURRENT_SOURCE_DIR}/Movement.cpp
+${CMAKE_CURRENT_SOURCE_DIR}/Powers.cpp
```
broxen: NetStructures has been merged with GameData, and all of these files moved there.

You can delete this cmake file.

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -516,6 +517,9 @@ void MapInstance::dispatch( Event *ev )
         case evLevelUpResponse:
             on_levelup_response(static_cast<LevelUpResponse *>(ev));
             break;
+    case evReceiveContactStatus:
```
broxen: Nit: check your formatting here to line up indents and make this more readable

Projects/CoX/Common/GameData/Contact.h
```diff
@@ -67,3 +150,5 @@ struct ContactEntryBulk
     QVector<ContactEntry> m_responses; // must be size 11, or cannot exceed 11?
     // size_t num_active_contacts; // we can use size()
 };
+
+struct CharacterData;
```
broxen: Let's put this forward declaration at the top of the file, underneath the includes

Projects/CoX/Common/Messages/Map/ContactList.h
```diff
@@ -0,0 +1,111 @@
...
+                bs.StorePackedBits(12, contact.m_npc_id);
+                bs.StorePackedBits(1, count); // m_contact_idx
+                bs.StorePackedBits(1, contact.m_task_index);
+                bs.StoreBits(1, contact.m_notify_player);
+                bs.StoreBits(1, contact.m_can_use_cell);
+                bs.StorePackedBits(1, contact.m_current_standing);
+                bs.StorePackedBits(1, contact.m_friend_threshold);
+                bs.StorePackedBits(1, contact.m_confidant_threshold);
+                bs.StorePackedBits(1, contact.m_complete_threshold);
+                bs.StoreBits(1, contact.m_has_location);
```
broxen: We need additional logic after this line, because the client only receives location values if location is true.
```
if(m_has_location)
{
    bs.StoreFloat(contact.m_location.location.x);
    bs.StoreFloat(contact.m_location.location.y);
    bs.StoreFloat(contact.m_location.location.z);
    bs.StoreString(contact.m_location.m_location_map_name);
    bs.StoreString(contact.m_location.m_location_name);
}

File: Projects/CoX/Common/Messages/Map/ContactList.h [hunk 33]
```diff
@@ -0,0 +1,111 @@
...
+        explicit ContactStatusList() : GameCommandEvent(MapEventTypes::evContactStatusList){}
+        ContactStatusList(vContactList contact_list) : GameCommandEvent(MapEventTypes::evContactStatusList)
+        {
+            m_contact_list = contact_list;
+        }
+
+        void serializeto(BitStream &bs) const override
+        {
+            bs.StorePackedBits(1, type()-evFirstServerToClient); // packet 43
+            bs.StorePackedBits(1, m_contact_list.size()); // svr_contacts_total_also_maybe
```
broxen: this should be `m_contact_list.size()-1` so that if the array size has changed, the client adjusts on it's end. Hit me up on discord if that doesn't make sense.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -987,7 +1023,10 @@ void addNpc(MapClientSession &cl, const char* name, glm::vec3 *loc, int variatio
     Entity *e = cl.m_current_map->m_entities.CreateNpc(getGameData(),*npc_def,idx, variation);
 
     forcePosition(*e, *loc);
-    e->m_velocity = {0,0,0};
+    e->m_direction.x = ori->x;
```
broxen: this *shouldn't* work because direction is a quaternion derived from a pyr value. I know you said you had no results in testing, but `forceOrientation()` *should* work as it sets the direction quat, the orientation pyr and toggles the flag necessary to force rotation. Perhaps there is another bug happening here?

Projects/CoX/Common/GameData/Character.cpp
```diff
@@ -231,6 +231,9 @@ void Character::GetCharBuildInfo(BitStream &src)
     // This must come after finalize
     addStartingInspirations(starting_insps);      // resurgence and phenomenal_luck
 
+
+    //addStartingContact("Officer Flint");
```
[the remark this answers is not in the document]
ZopharShinta: Error on my part. This was for early testing and should of been removed.
I assume we would create them via a script when a new player logs in.

PR #649, opened 2018-10-22
auth: Don't assert on bad reconnection attempts

### LINKED ISSUE #619: Server crash on incorrect auth procedure
### Expected behavior
The server should tell the connecting client that something went wrong and go on with the day.
### Actual behavior/symptoms
https://github.com/Segs/Segs/blob/915f9096beec4de621a20ad24d6869a9b57a74a2/Projects/CoX/Common/AuthProtocol/AuthEvents.h#L69 is hit and server shuts down.
I'm not totally sure, but I get this assertion several times per day.
### SEGS version
develop HEAD (915f909)

dracc:
Closes #619 .

- Replace `assert(op==3)` with a conditional warning message.

### DIFF
--- a/Projects/CoX/Common/Messages/Auth/AuthEvents.h
+++ b/Projects/CoX/Common/Messages/Auth/AuthEvents.h
@@ -8,6 +8,7 @@
 #pragma once
 #include "LinkLevelEvent.h"
 #include "Buffer.h"
+#include "Logging.h"
 class AuthLink;
 
 namespace SEGSEvents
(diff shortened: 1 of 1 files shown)


broxen:
We need further testing on this issue. I've made this same change locally and it has resolved any crashing (and client still reconnects fine)

But per my conversation with Nem on Discord, there may be an underlying issue here with our order of operations for disconnects. @dracc is running this live on his public server (which is where we've seen this error most) so I'm hoping we can cull some real world testing data to ensure there are no unforeseen consequences here.

broxen:
Any new results from this in testing?

dracc:
I haven't had many connections at all lately as a configuration failure shut down my server without it shutting down and the message was outside of my screen line count. :slight_frown:

broxen:
Moving to v7

broxen:
This has fired twice for me when testing various things locally. While it's probably not a true "solution" it does prevent the hard crash and outputs a nice warning message for server admins.

I'm merging this for now, and we'll try to address the underlying problem in v7.

PR #653, opened 2018-10-24
Fix serialization versioning for client options.

nemerle:
This might be a WIP :)

### DIFF
--- a/Components/BitStream.h
+++ b/Components/BitStream.h
@@ -22,7 +22,7 @@ Description: The BitStream class allows it's user to manipulate data in
 
 #include <cstdint>
 #include <utility>
-
+//#define CUSTOM_CLIENT_CODE
 #ifdef CUSTOM_CLIENT_CODE
 #define PUTDEBUG(x) bs.StoreString(x);
 #else
@@ -57,13 +57,8 @@ explicit        BitStream(size_t size);
         void    StorePackedBits(uint32_t nBits, uint32_t dataBits);
         void    appendBitStream(BitStream &src)
         {
-            if ((src.GetReadPos() & 7) == 0) 
-            {  // source is aligned ?
-                StoreBitArray(src.read_ptr(),src.GetReadableBits());
-                return;
-            }
+            //TODO: optimize this to partial memcopy in special cases ?
             uint32_t bits_to_store =src.GetReadableBits();
-            ByteAlign(false,true);
             while(bits_to_store>32)
             {
                 StoreBits(32,src.uGetBits(32));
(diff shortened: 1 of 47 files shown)


Projects/CoX/Common/Messages/Map/EntitiesResponse.h
```diff
@@ -32,25 +32,7 @@ explicit            EntitiesResponse();
...
-        mutable 
-        // [[ev_def:field]]
-        BitStream entity_removals {16};
-        mutable 
-        // [[ev_def:field]]
-        BitStream client_data {32};
-        mutable 
-        // [[ev_def:field]]
-        BitStream follow_up_commands {16};
+        BitStream blob_of_death {220};
```
broxen: You know better than me, but just checking:

This was broken out into several bitstreams totaling 304 but now equals 220 combined? Is that correct?

Components/BitStream.h
```diff
@@ -22,7 +22,7 @@ Description: The BitStream class allows it's user to manipulate data in
 
 #include <cstdint>
 #include <utility>
-
+//#define CUSTOM_CLIENT_CODE
```
broxen: This is clever. I'll use it on my SGs and Tailor branches

Projects/CoX/Common/CRUDP_Protocol/CRUD_Events.cpp
```diff
@@ -0,0 +1,87 @@
...
+    bs.GetPackedBits(1);
+}
+void DisconnectResponse::serializeto(BitStream &bs) const
+{
+    bs.StorePackedBits(1, 0);
+    bs.StorePackedBits(1, 6);
+}
+void Idle::serializefrom(BitStream &bs)
+{
+    // TODO: check this
```
broxen: How do I check this?

Projects/CoX/Common/GameData/playerdata_definitions.h
```diff
@@ -16,7 +16,7 @@
 // this part of the Entity is for Players only info
 struct PlayerData
 {
-    static const constexpr  uint32_t    class_version   = 2;
+    enum : uint32_t {class_version   = 2};
```
broxen: Is this the new preferred format for all cases or just classed serialize methods?

Projects/CoX/Common/Messages/Map/EntitiesResponse.h
nemerle: This number is just the initial buffer size, so it's mostly a WAG :P, and the bitstream will get expanded if there's a need.

Projects/CoX/Common/CRUDP_Protocol/CRUD_Events.cpp
nemerle: Well, we could be checking idle packet's data, and making sure that it trully is built from two 0s ? :)

Projects/CoX/Common/GameData/playerdata_definitions.h
nemerle: this is a better approach since it does not require also defining the variable in a separate cpp file :)

PR #659, opened 2018-10-26
Fix a map entering process to solve link/server blocking.

nemerle:
The client entering a new map does not want/need constant full world
state updates. So now we only send 2 full world updates before client
'resumes' and starts receiving incremental ones.


broxen:
Zoning is much quicker now. =D

I'll merge this before my stateStorage and doorAnims PRs, juuuuuuuuusssst in case we bump into eachother anywhere (it doesn't look like we will).

PR #664, opened 2018-10-30
Temporary workaround for #555 team-removal and side-kick removal

broxen:
## Summary
This still has issues, to avoid using MapInstance in Team.cpp there is a condition where removing a sidekick can result in one member of the sidekick duo having a "sidekick" state still. This really wont be resolved until #642 is complete.

- The goal of this PR is to prevent the hard crashes when someone removes a team member, or zones while on a team.
- This isn't pretty, but it's supposed to be temporary.


HeraldOfOmega:
It is too, pretty!

PR #666, opened 2018-10-30
Gcc53 abs fix

nemerle:


HeraldOfOmega:
Thanks for fixing this one, guys!

PR #678, opened 2018-11-02
Tasks / LuaBot added to Lua

### LINKED ISSUE #671: Have addNPC return Entity IDX for utilization in Lua
## Summary
Following discussions with @ZopharShinta it may prove useful to have addNPC return the entity IDX to Lua so that it can reference those NPCs in the scripts (and allow users to interact with them, etc)

Also it would help eliminate potentially buggy code like this (if more than 3 users logged in, this code will cause clicking on players to produce simple_dialog)
```lua
elseif id >= 303 then
        MapClientSession.simple_dialog(client, [[This doesn't concern you "hero".]])
```

ZopharShinta:
Tasks added to Lua.
Generic scripts / LuaBot added for testing scripting functions. (Should be in the correct folder for outbreak)
Lua function called when player logs in for player setup steps. (After MOTD is sent)

I believe this closes #671 as well.
Added `npc_added` function to call into Lua after the `addNpc` function finishes with the idx of npc added


Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -294,6 +369,10 @@ std::string ScriptingEngine::callFuncWithClientContext(MapClientSession *client,
     m_private->m_lua["client"] = client;
     m_private->m_lua["vContacts"] = client->m_ent->m_char->m_char_data.m_contacts;
     m_private->m_lua["heroName"] = qPrintable(client->m_name);
+    m_private->m_lua["m_db_id"] = client->m_ent->m_db_id;
+    if(client->m_ent->m_char->m_char_data.m_tasks_entry_list.size() > 0){
```
broxen: Please set brackets on their own line for these if statements. Thanks Zophar!

PR #679, opened 2018-11-03
Less asserts where error messages will work

cyanpill:
## Summary
From a discussion in discord, train is causing asserts due to checking the level higher than you current, which causes expForLevel to assert. I added an error message to train(), and changed the assert in expforlevel into a min0 max49 so we won't have this problem again.

Also, an assert in emails that shouldn't be needed, and 2 returns that are needed to prevent asserting.


broxen:
A quick note. Either way is fine, but In lieu of a constant: `m_experience_and_debt_per_level.m_ExperienceRequired.size()` equates to 49.

PR #681, opened 2018-11-04
Email Handler Part 3 (New branch)

### LINKED ISSUE #618: Email Handler Requirements, Round 3
Self-note for the next step of Email Handler. I hope to get this done before 0.6 launches. Others can chime in to help out as well of course :)

- [ ] Implement timestamps
- [ ] Remove ``FillEmailRecipientIdErrorMessage`` and change it to ``EmailCreateStatusMessage``, which will be sent to the email sender in **both** successful and unsuccessful email sending.
- [ ] In ``MapInstance``'s response to ``EmailCreateStatusMessage``, use ``EmailStatusMessage`` that is a ``GameCommand`` (much like ``EmailHeader`` and ``EmailRead``)
- [ ] Cleanup code and probably structure as well

eos428:
Made a new branch because too lazy to merge possible conflicts. Also because I don't want to clog it up with old commit messages.

Closes #618 .

From the checklist I made in the issue page:

- [ ] Implement timestamps
- [ ] Implement sending emails to multiple recipients
- [x] Remove ``FillEmailRecipientIdErrorMessage`` and change it to ``EmailCreateStatusMessage``, which will be sent to the email sender in **both** successful and unsuccessful email sending.
- [x] In ``MapInstance``'s response to ``EmailCreateStatusMessage``, use ``EmailStatusMessage`` that is a ``GameCommand`` (much like ``EmailHeader`` and ``EmailRead``)
- [x] Cleanup code and probably structure as well (Half-done?)

Comments on the checklist:
1. Timestamps: Found a seemingly valid method, just need to test and implement properly
2. Not started on this yet
3. Didn't actually remove ``FillEmailRecipientIdErrorMessage`` (it will send failing ``EmailCreateStatusMessage`` to sender's map instance instead). Also, a successful ``EmailCreateStatusMessage`` is sent at ``on_email_create_response``
4. Should be ok
5. Need further checks, since Email Handler is quite different to other handlers with multiple events being sent in a single function, and then bringing (cerealized) email data to db_handler back and forth, and so on.


eos428:
Ready for review. You can send emails to multiple recipients using either ``user1; user2`` or ``user1;user2``. Not sure about more spaces than 1, but it shouldn't be allowed.

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2798,10 +2798,22 @@ void MapInstance::on_email_read_by_recipient(EmailWasReadByRecipientMessage *msg
     // route is DataHelpers.onEmailRead() -> EmailHandler -> MapInstance
 }
 
-void MapInstance::on_email_send_error(EmailSendErrorMessage *msg)
+void MapInstance::on_email_create_status(EmailCreateStatusMessage *msg)
 {
     MapClientSession &map_session(m_session_store.session_from_token(msg->session_token()));
-    sendInfoMessage(MessageChannel::DEBUG_INFO, msg->m_data.m_error_message, map_session);
+
+    // for some reason EmailMessageStatus with the boolean set to true will crash the client
```
broxen: This crashes because of the following client-side logic:
```
void receiveEmailMessageStatus(DbPacket *pak)
{
    int success = my_pakGetPackedBits(pak, 1);
    if ( success )
        emailSetNewMessageStatus(success, "");
    else
        emailSetNewMessageStatus(0, my_pakGetString(pak));
}
```

`EmailMessageStatus.h` needs to be adjusted to consider this if/else case:
```
void    serializeto(BitStream &bs) const override 
    {
        bs.StorePackedBits(1, type()-evFirstServerToClient); // packet 67
        bs.StorePackedBits(1, m_status);
        if(!m_status)
            bs.StoreString(m_recipient);
    }
```
It's worth noting that `m_status` can be a boolean, as it's values are *only* ever `true/false`. It's also probably mislabeled. It can probably be named `m_is_successful` or something.

broxen: The purpose of this, is to actually check that the recipient exists, and if they do not, then the recipient list is provided. Client side calls `emailSetNewMessageStatus` which looks something like this:
```
void emailSetNewMessageStatus(int success, const char *recipients)
{
    if ( success )
        window_setMode(WND_EmailComposeTab,WindowDisplayMode::WINDOW_SHRINKING);
    else
    {
        QString err_str("Email not sent, because these recipients do not exist:\n");
        err_str += recipients;
        err_str.replace('%', '@');
        dialogStdQ(1, err_str, nullptr, nullptr, nullptr, nullptr, 1);
    }
}
```

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -1905,54 +1905,55 @@ void cmdHandler_SidekickDecline(const QString &/*cmd*/, MapClientSession &sess)
...
         return;
     }
 
-    sendEmail(sess, result[1], result[3], result[4]);
-
-    QString msg = "Email Sent to recipient: " + result[1];
-    qDebug().noquote() << msg;
-    sendInfoMessage(MessageChannel::SERVER, msg, sess);
+    for (const auto &recipient : recipients)
+        sendEmail(sess, recipient, result[3], result[4]);
```
broxen: As mentioned in my comments in `MapInstance.cpp` we should check that the recipient exists here, and if not, send the status packet (67) with m_success = false, with the recipient list containing the faulty recipients.

It's not clear to me if the email should successfully send to those that do exist, or fail if even one recipient fails. I suppose we can probably decide for ourselves here, and I'd probably go with whatever makes the most sense from the user experience side of things (e.g. was it clear that my email was success for some people? If not, maybe just fail for all recipients)

Projects/CoX/Servers/GameDatabase/GameDBSyncHandler.cpp
```diff
@@ -235,12 +235,8 @@ void GameDBSyncHandler::on_fill_email_recipient_id(FillEmailRecipientIdRequest *
     if (db_ctx.fillEmailRecipientId(ev->m_data, resp))
         HandlerLocator::getEmail_Handler()->putq(new FillEmailRecipientIdResponse(std::move(resp), ev->session_token()));
     else
-    {
-        QString error_msg = QString("Failed to send email because character "
```
broxen: I might have simply missed it, but did this check move somewhere else? We do need to check recipient names against actual players, so that we can return the proper success status in `EmailMessageStatus()` (packet 67)

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1222,4 +1206,19 @@ void showMapMenu(MapClientSession &sess)
     QString msg_body = createMapMenu();
     showMapXferList(sess, has_location, location, msg_body);
 }
+
+uint32_t getCurrentTime()
```
broxen: You can simplify this with Qt, as I did in my SuperGroups PR where I created a TimeHelpers.cpp/h. The new file may not be needed, but you may consider the following as it's might be more readable:
```
QDateTime getPostgreEpoch()
{
    return QDateTime(QDate(2000,1,1));
}
 
int64_t getSecsSince2000Epoch()
{
    QDateTime base_date(QDate(2000,1,1));
    return base_date.secsTo(QDateTime::currentDateTime());
}
```

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2798,10 +2798,22 @@ void MapInstance::on_email_read_by_recipient(EmailWasReadByRecipientMessage *msg
...
+    // so this will be sent only for false cases
+    if (!msg->m_data.m_status)
+    {
+        EmailMessageStatus* msg_status = new EmailMessageStatus(msg->m_data.m_status, msg->m_data.m_recipient_name);
+        map_session.addCommandToSendNextUpdate(std::unique_ptr<EmailMessageStatus>(msg_status));
```
broxen: Let's use the new syntax sugar here: `map_session.addCommand<EmailMessageStatus>(msg_status);`

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -1905,54 +1905,55 @@ void cmdHandler_SidekickDecline(const QString &/*cmd*/, MapClientSession &sess)
 
 void cmdHandler_EmailHeaders(const QString & /*cmd*/, MapClientSession &sess)
 {
-    sendEmailHeaders(sess);
```
broxen: It appears as though the client sends us this `/emailheaders` SlashCommand when the EmailWindow is opened or refreshed.

Seems we do in fact want to respond with `sendEmailHeaders(sess);` here, to refresh the client side data. No?

Projects/CoX/Servers/GameDatabase/GameDBSyncHandler.cpp
eos428: It's done at void EmailHandler::on_fill_email_recipient_id_error which sends EmailCreateStatusMessage with the boolean set to false :)

Projects/CoX/Servers/MapServer/SlashCommand.cpp
broxen: I'd have to test, but my guess is that if you're sending emails back and forth without zoning, this `sendEmailHeaders` is requesting refresh in between zoning?

Am I overlooking something?

Projects/CoX/Servers/GameDatabase/GameDBSyncHandler.cpp
broxen: Perfect.

Projects/CoX/Servers/MapServer/DataHelpers.h
```diff
@@ -48,11 +48,14 @@ QString createMapMenu();
...
  * sendEmail Wrappers for providing access to Email Database
  */
-void sendEmailHeaders(MapClientSession& sess);
+void getEmailHeaders(MapClientSession& sess);
 void readEmailMessage(MapClientSession& sess, const uint32_t email_id);
 void sendEmail(MapClientSession& sess, QString recipient_name, QString subject, QString message);
 void deleteEmailHeaders(MapClientSession& sess, const uint32_t email_id);
 
+// to get the current time since whatever they set as their beginning
+QDateTime getPostgreEpoch();
```
nemerle: It seems `getPostgreEpoch`is not used anywhere?

eos428: No, so I guess we could remove it and avoid extra hassle in merging conflicts if needed.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -251,7 +246,8 @@ void sendEmail(MapClientSession& sess, QString recipient_name, QString subject,
         return;
     }
 
-    uint32_t timestamp = 0;
+    // uint32_t timestamp = 0;
```
broxen: Can remove this comment

PR #688, opened 2018-11-08
Power changes to toggles and buffs

### LINKED ISSUE #613: Powers can be activated while still recharging
### Expected behavior
Powers cannot activate until recharged

### Actual behavior/symptoms
Powers can be activated at any stage of recharge

Create a character with a ranged attack
Spam ranged attacks on Officer Flint.

### SEGS version
0.5.x

### Operating system
Windows 7

### C++ compiler and version
mingw53_32

### Other Notes
Didn't expect this to be possible on the client

### LINKED ISSUE #628: Toggle powers not being handled properly
### Expected behavior
User can activate and deactivate toggleable powers (ie. Sprint)
### Actual behavior/symptoms
Toggleable powers behave as activated powers and are able to be constantly activated, but are never able to be deactivated.
1. Activate Sprint (or other toggleable power)
2. Continue to activate Sprint
### SEGS version
Latest master
### Other relevant information
Toggleable powers will be constantly toggled with no way of untoggling them

cyanpill:
## Summary
Split usePower() into separate functions for better flow of power usage
Activation queue now actually counts down before a power fires
Toggles have their own queue, where the power effects fire automatically and endurance is spent over time
Apply some benefits from buffs to character data, and then check those stats for regeneration, recovery, runspeed, accuracy, toHit, defense 
Fly, teleport and revive effects.
Added hardcoded data for inspirations, replacing inspiration code with a call to usePower.
ValidTarget() to check if a target can be targeted, if it can be autohit, and if it's affected by effects.
ValidTarget() checks a villain flag, for use by AI.
/alignment player command to switches hero/villain.
"none" makes the entity an invalid target for powers, useful for contacts/admins that want to be left alone
"both" makes the entity a valid target for anything, useful for training dummies.
aoe powers check for valid targets in range.
Toggles turn off other toggles that conflict (example: hover and fly)
made each buff contain a vector of name/values/type of stat changes for organization.
Activation rings show up on power icon's when they should.
Powers don't work if you are dead, except for self rez.
Movement is stopped while activating or held by a status effect(requires the server to have movement authority)
Powers can't be activated while held or resting
Toggles turn off when out of power, of the anchor is out of range/dead
Powers that don't have data get either damage or healing, with a value based on activation and recharge times
Damage scales by level and archetype.
Loading of power data from data/bin/powers.json
some usage of damage types

## Issues closed
Closes #628
Closes #613
And a couple lines of #525 

### Further Goals:
Damage/defense types used for hit checks and damage calculations.
Set Debt when a player dies, applying debt protection
Keep buffs when moving between zones/log out, check if it should be deleted on login (store timestamp) 
Auto powers should load when you log in, and when you buy the power (requires access to entities in addpower).
Delayed/ repeat effects (add a queue on targets for effects, calls doeffect directly).
Functions can be moved around, maybe to entityhelper? since I removed some messaging directly to the client.
Advanced effects for sleep/fear/confuse/taunt
Cone limiting targets by an arc.
Knockback/repel effects.
Enhancements will be even more work, I have a list.
Spawn powers require basic AI first.
Animations/FX, movement system
...and that's all powers need!


Projects/CoX/Common/GameData/Character.cpp
```diff
@@ -659,6 +659,12 @@ bool toActualCharacter(const GameAccountResponseCharacterData &src,
         main_costume.setSlotIndex(costume.m_slot_index);
         main_costume.setCharacterId(costume.m_character_id);
     }
+
+    Parse_CharAttrib replace;                   // not the best place to do this, but current attribs need to 0'd every log in
```
broxen: This should probably go in `Character::Character()` instead

broxen:
Regarding `Activation rings aren't showing up on power icon's when they should` what do you mean precisely?

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2284,26 +2284,50 @@ void MapInstance::on_activate_power(ActivatePower *ev)
...
+    }
+    if (powtpl.Target == StoredEntEnum::Teleport && (powtpl.EntsAffected[0] == StoredEntEnum::Caster ))
+        tgt_idx = session.m_ent->m_idx;
+    qWarning() << "Entity: " << session.m_ent->m_idx << "has activated power" << powtpl.m_Name;
+
     sendFaceLocation(session, ev->location);
 
-    qCDebug(logPowers) << "Entity: " << session.m_ent->m_idx << "has activated power"<< ev->pset_idx << ev->pow_idx << ev->target_idx << ev->target_db_id;
+    session.m_ent->m_target_loc = ev->location;
+    QueuedPowers qpowers;
```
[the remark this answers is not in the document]
cyanpill: Only some of the checks in usepower make sense for location based powers. For example there's no toggled location powers. Possibly usepower could be split up further.

cyanpill:
When a power is activating, there should be a ring around the icon,  similar with toggles that are active. It has something to do with sending the clint activation states.

Projects/CoX/Servers/MapServer/MapInstance.cpp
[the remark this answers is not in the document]
broxen: That makes sense. I think we probably should. There are several places where we can refactor usePower() logic into reusable pieces.
- checkRange()
- checkTarget()
- queuePower()
- activatePower()
- rechargePower()

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -112,35 +114,52 @@ void World::checkPowerTimers(Entity *e, uint32_t msec)
...
+    for(auto rpow_idx = e->m_auto_powers.begin(); rpow_idx != e->m_auto_powers.end(); /*rpow_idx updated inside loop*/ )
+    {
+        rpow_idx->m_time_to_activate   -= (float(msec)/1000);
+
+        if(rpow_idx->m_time_to_activate   <= 0)
+        {
+            CharacterPower * ppower = nullptr;
+            ppower = getOwnedPowerByVecIdx(*e, rpow_idx->m_pow_idxs.m_pset_vec_idx, rpow_idx->m_pow_idxs.m_pow_vec_idx);
+            const Power_Data powtpl = ppower->getPowerTemplate();
+            doPower(*e, *rpow_idx.base());
```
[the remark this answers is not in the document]
cyanpill: I think qcreator suggested that, I'll try without.

broxen:
Those activation rings were psuedo-working before (they appeared, but never left). Are they gone altogether now?

If so, it must be the way we handle `m_queued_powers` as we pull from there to send those activation states.

Projects/CoX/Common/GameData/EntityHelpers.cpp
```diff
@@ -166,7 +166,39 @@ void toggleMovementAuthority(Entity &e)
     toggleFullUpdate(e);
     toggleControlId(e);
 }
-
+void modifyAttrib(Entity &e, QString name, float value)
+{
+    if (name == "regeneration")
```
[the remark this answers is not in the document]
cyanpill: It is set to lowercase further up.

Projects/CoX/Common/GameData/Character.cpp
cyanpill: Hm, Character::Character() might not be the best place , it's called about 10 times per entity. I'll find a place that only gets called once or twice...

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1261,22 +1309,22 @@ void showMapMenu(MapClientSession &sess)
 }
 void setAlignment(Entity &e, QString align)
```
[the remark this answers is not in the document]
cyanpill: Hm, that should work, but I never tested 2 players, I just assumed it would work :disappointed:

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -813,20 +823,32 @@ void doEffect(Entity &ent, Entity *target_ent, CharacterPower * ppower){
             else if (lower_name == "revive")
                 revivePlayer(*target_ent, ReviveLevel::AWAKEN);
             else if (lower_name == "teleport")
-               forcePosition(*target_ent, ent.m_target_loc);//
+            {
+               forcePosition(*target_ent, ent.m_target_loc);
```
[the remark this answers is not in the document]
cyanpill: Aha, this is because of the on_activate_power_at_location() not calling usePower(), and thusly not checking the valid targets. I'll fix it with the other.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -813,20 +823,32 @@ void doEffect(Entity &ent, Entity *target_ent, CharacterPower * ppower){
...
                 {
                     Buffs buff;
                     buff.m_name = powtpl.m_Name;
                     buff.m_buff_info = ppower->m_power_info;
                     buff.m_activate_period = powtpl.pAttribMod[i].Duration ;
                     buff.m_value_name = lower_name;
-                    buff.m_value = powtpl.pAttribMod[i].Magnitude;
-                    addBuff(ent, buff);
+                              buff.m_value = powtpl.pAttribMod[i].Magnitude;
+                    addBuff(*target_ent, buff);
```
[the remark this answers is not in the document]
cyanpill: Hover and rest add more than one buff at a time, because they buff more than one thing, and I send an icon for each atrrib buffed. I could change the buffs to hold a vector of attrib changes(which would cause problems with some powers), or, I noticed in the issue 23 powers.bin there is a bool of show_Icon. I could add that into the power data, but that would be a deviation from issue 1. I think would be a better option.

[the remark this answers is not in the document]
cyanpill: For example in some dark powers, they reduce tohit for one length of time, and stun or whatever for a different length:
http://web.archive.org/web/20140821075747/http://tomax.cohtitan.com/data/powers/power.php?id=Tanker_Melee.Dark_Melee.Touch_of_Fear

I think those powers had one buff icon for each effect duration. Or maybe there was a separation for CC effects and stat debuffs?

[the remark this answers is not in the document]
broxen: Oh interesting. I don't know. I only ever remember one icon for buffs/debuffs.

It's interesting because the client only accepts the pow_tpl indexes for buffs. So I assume we're sending the same power more than once and calculating the different effects server side?

Maybe your right that this behavior changed later and later clients expected more information for buffs/debuffs?

I'll throw up the question in Discord and see if anyone has any video to demonstrate the behavior. I'm sure there's something out there.

At the same time, I'm certain Rest and Hover only had one buff icon. And I know that using more than one luck inspiration would simply reset the timer on that buff (not stack). So it sounds like we may need to accommodate both scenarios.

broxen:
Awesome work cyanpill! I'll review this week before traveling for Thanksgiving.

One thing I would add to your Todo for powers is auto-firing of Default Powers. You can set this by right clicking Clickable powers. It should result in a green circle around the power icon.

Again, thank you for all of this hard work!

cyanpill:
I think I have default powers working as intended- you have to click the power again, and it keeps firing until there is no target. I looked to see if there would be an easy way to activate the power as soon as it's set to default, but no target is sent with that event.

cyanpill:
Just found this guide that explains a bit more about how powers work: https://paragonwiki.com/wiki/The_Players%27_Guide_to_the_Cities/Fighting_and_Other_Actions
"Most click or toggle powers can be queued. Trying to reactivate a power before it has finished recharging will queue the power. Trying to activate a power that can only be executed under the right conditions will frequently queue a power. Queued powers will activate as soon as the power is recharged and all special conditions are met. For example, targeting an enemy and attempting to activate a charged melee attack from halfway across a room will queue the melee attack. The attack will then automatically execute if you close into melee range with the target. Queued toggle powers will automatically activate when they finish recharging. Queued power icons will be marked by a red border in your power tray. Only one power may be queued at a time."
That not how we have the queue, but that can be fixed in another issue :smiley:

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -769,113 +837,183 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, int32_t tgt_idx,
...
+    else if (lower_name == "endurancemod")  // Change endurance, can be positive or negative
+    {
+        if (mod.Aspect == AttribMod_Aspect::Current)
+            scale = target_ent->m_char->m_max_attribs.m_Endurance * scale;
+        setEnd(*target_ent->m_char, getEnd(*target_ent->m_char)+scale);
+    }
+    else if (lower_name == "knockback") //   KB isn't a buff like other CC
+        ;
+    else if (lower_name == "revive")
+        revivePlayer(*target_ent, ReviveLevel::AWAKEN);
```
broxen: This seems good for now, but eventually we'll need to handle different levels of Revive (different strength inspirations for example)

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -769,113 +837,183 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, int32_t tgt_idx,
...
+    else if (lower_name == "teleport")
+    {
+       forcePosition(*target_ent, ent.m_target_loc);
+    }
+    else if (lower_name == "spawn")
```
broxen: Are there any powers or inspirations that provide a "spawn" effect?

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -163,8 +200,8 @@ void World::regenHealthEnd(Entity *e, uint32_t msec)
         float hp = getHP(*e->m_char);
         float end = getEnd(*e->m_char);
 
-        float regeneration = hp * (1.0f/20.0f) * float(msec)/1000/12;
-        float recovery = end * (1.0f/15.0f) * float(msec)/1000/12;
+        float regeneration = getMaxHP(*e->m_char) * (e->m_char->m_char_data.m_current_attribs.m_Regeneration/15.0f * float(msec)/1000/12);
```
broxen: Good catch on the mapStat and regen, but I have the calculation for Regen as 5%, instead of 6.66%. Maybe my reference is incorrect?
From here: https://paragonwiki.com/wiki/Health_Regeneration

Endurance Recovery is 6.66% from the same source.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
broxen: Some forms of revive recover end as well.

Something to keep an eye out for

Projects/CoX/Common/GameData/CharacterAttributes.h
```diff
@@ -38,18 +38,18 @@ struct Parse_CharAttrib
     std::array<float,24> m_DefenseTypes;
     float m_Defense             = 0;
     float m_Evade               = 0;
-    float m_SpeedRunning        = 0;
-    float m_SpeedFlying         = 0;
-    float m_SpeedSwimming       = 0;
-    float m_SpeedJumping        = 0;
-    float m_jump_height         = 0;
+    float m_SpeedRunning        = 1;
```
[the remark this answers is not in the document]
cyanpill: Aha, I forgot about the Parse_AllCharClasses has a "AttribBase" for each archetype, we'll want to initialize from that for these values.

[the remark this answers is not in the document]
broxen: That's what I failed to articulate. Thanks Cyanpill

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -91,18 +91,47 @@ void World::checkPowerTimers(Entity *e, uint32_t msec)
...
         {
-            e->m_queued_powers.dequeue(); // remove first from queue
-            e->m_char->m_char_data.m_has_updated_powers = true;
+            if (checkPowerRecharge(*e, qpow.m_pow_idxs.m_pset_vec_idx, qpow.m_pow_idxs.m_pow_vec_idx)
+                    && checkPowerRange(*e, qpow.m_tgt_idx, qpow.m_pow_idxs.m_pset_vec_idx, qpow.m_pow_idxs.m_pow_vec_idx))
+                   e->m_is_activating = true;       //only pass this check once the power is recharged
+            else if(e->m_queued_powers.size() > 1)
+            {
+                e->m_queued_powers.append(qpow);
+                e->m_queued_powers.replace(e->m_queued_powers.size()-1, qpow);     //push recharging power to the back
```
broxen: Doesn't this have the same affect as the append above it? What's your intended goal here?

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -91,18 +91,47 @@ void World::checkPowerTimers(Entity *e, uint32_t msec)
...
+            }
         }
+        else
+        {
+            if(qpow.m_time_to_activate <= -0.3f)        // wait a little longer to make sure the activation state has been sent
```
broxen: Where did we derive `0.3f`? Can we declare and define this magic number as a static variable?

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -91,18 +91,47 @@ void World::checkPowerTimers(Entity *e, uint32_t msec)
...
+                qpow.m_activation_state = false;
+
+            }
+            qpow.m_time_to_activate -= (float(msec)/1000);
+
+        }
+    }
+    else if(e->m_char->m_char_data.m_trays.m_has_default_power) // only activate the default power if there is no other power in the activation queue
+    {
+        if (checkPowerRecharge(*e, e->m_char->m_char_data.m_trays.m_default_pset_idx, e->m_char->m_char_data.m_trays.m_default_pow_idx))
```
broxen: We're using `e->m_char->m_char_data.m_trays` so often that perhaps for readability's sake we declare a reference above and use that throughout this section.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -583,98 +583,250 @@ void sendDoorAnimExit(MapClientSession &sess, bool force_move)
...
 {
     QString from_msg, to_msg;
     CharacterPower * ppower = nullptr;
+
     ppower = getOwnedPowerByVecIdx(ent, pset_idx, pow_idx);
     const Power_Data powtpl = ppower->getPowerTemplate();
-
     if(ppower == nullptr || powtpl.m_Name.isEmpty())
         return;
+    if(/*ent.m_char->getHealth() == 0.0f*/ ent.m_state_mode == ClientStates::DEAD && powtpl.CastableAfterDeath == 0)   //Allows self rez
```
broxen: This inline comment? Maybe we should check for that? Maybe use a logical OR here?

Also we use this exact same check 3 times in this PR. Maybe best to refactor out into some kind of `checkIfDead()`

Projects/CoX/Common/Runtime/HandleBasedStorage.h
```diff
@@ -10,6 +10,7 @@
 #include "Handle.h"
 #include <cassert>
 #include <vector>
+#include "stddef.h"
```
broxen: I don't see where this is needed. Can we remove?

Projects/CoX/Common/GameData/power_serializers.cpp
```diff
@@ -242,36 +242,42 @@ bool loadFrom(BinStore *s, AllPowerCategories &target)
...
+        archive(cereal::make_nvp("ContinuingFX",src.ContinuingFX));
+        archive(cereal::make_nvp("ConditionalBits",src.ConditionalBits));
+        archive(cereal::make_nvp("ConditionalFX",src.ConditionalFX));
+        archive(cereal::make_nvp("EntityDef",src.EntityDef));
+        archive(cereal::make_nvp("PriorityListDefense",src.PriorityListDefense));
+        archive(cereal::make_nvp("PriorityListOffense",src.PriorityListOffense));
+        archive(cereal::make_nvp("PriorityListPassive",src.PriorityListPassive));
+    }
+    catch(cereal::Exception& e)
+    {
```
broxen: Does cereal already spit out an error message here, or should we display one?

Perhaps add a comment inside the block so it's clear why we aren't handling the catch.

Projects/CoX/Common/GameData/Powers.h
```diff
@@ -103,9 +104,13 @@ enum : uint32_t { class_version = 1 };
 
 struct Buffs
 {
+    QString         m_name                  = "unknown";
     PowerPool_Info  m_buff_info;
     float           m_time_to_activate      = 0.0f;
-    float           m_activate_period       = 0.0f;
+    float           m_duration              = 0.0f;
+    std::vector<float>           m_value;
```
broxen: To clarify, is this for buffs with more than one "effect"? Maybe add a small comment so it's clear for future contributors.

Projects/CoX/Common/GameData/Powers.h
```diff
@@ -85,6 +85,7 @@ struct QueuedPowers
     bool            m_active_state_change   = false;
     bool            m_activation_state      = false;
     bool            m_timer_updated         = false;
+    int32_t         m_tgt_idx               = -1;
```
broxen: Do we need to have target here again? Can you queue a power on one target and switch to another? This seems right, but I honestly can't remember how it worked.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -773,113 +916,211 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, int32_t tgt_idx,
...
-            // Send message to another player target
-            if(target_ent->m_type == EntType::PLAYER && tgt_idx != ent.m_idx)
-                sendInfoMessage(MessageChannel::DAMAGE, to_msg, *target_ent->m_client);
+    if (attrib.StackType == AttribStackType::Replace)           // most are replace
+    {
+        for (Buffs & temp : ent.m_buffs)
+        {
+            if ((temp.m_name == powtpl.m_Name) && (temp.source_ent_idx == entidx))
+            {
+                for (int i =0; i<temp.m_value_name.size();i++)
```
broxen: Let's wrap this for loops body in brackets to make it clear what it loops over. Also, you could probably use a ranged based for loop here.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -773,113 +916,211 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, int32_t tgt_idx,
...
+    {
+        if (buff_idx->m_name == buff.m_name)
+        {
+            for (int i =0; i<buff.m_value_name.size();i++)
+                modifyAttrib(ent, buff.m_value_name[i], -buff.m_value[i]);
+
+            ent.m_buffs.erase(buff_idx);
+            return;                                         // without this it removes all buffs of same name, if we need that?
+        }
+        else buff_idx++;
```
broxen: For code consistency, let's pre-increment this iterator.

Otherwise you might be able to replace this entire for loop with find_if()?

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -773,113 +916,211 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, int32_t tgt_idx,
...
+                    {
+                        setStateMode(*target_ent, ClientStates::RESURRECT);
+                        revivePlayer(*target_ent, ReviveLevel::AWAKEN);
+                    }
+                    else if (lower_name == "teleport")
+                    {
+                       forcePosition(*target_ent, ent.m_target_loc);
+                    }
+                    else if (lower_name == "summon")
+                       ; //lots of work needed for this one
```
broxen: For code clarity, can we wrap these empty statements in brackets and spit out a debug message to `logPowers` or something?

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -773,113 +916,211 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, int32_t tgt_idx,
...
+                               sendFloatingNumbers(*target_ent->m_client, target_ent->m_idx, -int(scale));   // this sends damage without needing the FloatingInfo_Damage
+                               sendInfoMessage(MessageChannel::DAMAGE, QString("%1 power heals you for %2!") .arg(ent.name()) .arg(scale) , *target_ent->m_client);
+                        }
+                    }
+                    else if (lower_name == "endurancemod")  // Change endurance, can be positive or negative
+                    {
+                        setEnd(*target_ent->m_char, getEnd(*target_ent->m_char)+(target_ent->m_char->m_max_attribs.m_Endurance*scale));
+                    }
+                    else if (lower_name == "knockback") //   KB isn't a debuff like other crowd control effects
+                        ;
```
broxen: For code clarity, can we wrap these empty statements in brackets and spit out a debug message to `logPowers` or something?

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -753,16 +856,56 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, int32_t tgt_idx,
...
+    }
+    QString from_msg;
+    const Power_Data powtpl = ppower->getPowerTemplate();
+    if (!validTargets(*target_ent, ent, powtpl.EntsAutoHit))//
+       {
         //roll to hit
         int roll = rand()%100;
-        int chance = int(powtpl.Accuracy * 75);
+                                    //todo: find defense values for attack types, find highest if there is more than one
+        int chance = int(std::min(95.0f,(std::max(5.0f, (powtpl.Accuracy /* * enhancement accuracy */    //min 5%, max 95%
```
broxen: I know this algo is beefy already, but it may be best to replace these magic numbers with static variables. Especially since we may eventually expose them to server admins for customization.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -583,98 +583,250 @@ void sendDoorAnimExit(MapClientSession &sess, bool force_move)
...
+    int tgt_idx = powerinput.m_tgt_idx;
+    if(tgt_idx == -1)
+        tgt_idx = getIdx(ent);
 
-        // Face towards your target
-        sendFaceEntity(*ent.m_client, tgt_idx);
+    Entity *target_ent = getEntity(ent.m_client, tgt_idx);
+    if(target_ent == nullptr)
+    {
+
```
broxen: nit: excess line return here

Projects/CoX/Common/GameData/GameDataStore.cpp
```diff
@@ -479,28 +479,131 @@ bool GameDataStore::read_settings(const QString &/*directory_path*/)
...
+    else if(read_data_to<AllPowerCategories, powers_i0_requiredCrc>(directory_path,
+                                                                    "bin/powers.bin",m_all_powers))
+    {
+        qDebug() << "Loaded power data from powers.bin!";
+    }
+    else
         return false;
 
-    // Hardcoding of stats to test powers
+    // Hardcoding of stats to test powers, these will move into the json soon
```
broxen: Please add a TODO here `// TODO: Hardcoding...` so that it's easy to look up TODO items in the code base. Thanks!

Projects/CoX/Common/GameData/EntityHelpers.cpp
```diff
@@ -168,6 +168,107 @@ void toggleMovementAuthority(Entity &e)
...
+        else
+            e.m_motion_state.m_is_flying = false;
+    }
+    else if (name == "flight_speed")
+    {
+        e.m_char->m_char_data.m_current_attribs.m_SpeedFlying += value;
+        setSpeed(e, e.m_char->m_char_data.m_current_attribs.m_SpeedRunning, e.m_char->m_char_data.m_current_attribs.m_SpeedJumping,
+                 e.m_char->m_char_data.m_current_attribs.m_SpeedFlying);
+    }
+    else if (name == "immobolized")//copy for hold, sleep,stun
```
broxen: Should be `immobilized` I think.

Projects/CoX/Common/GameData/CharacterHelpers.cpp
```diff
@@ -141,6 +141,53 @@ void setBattleCry(Character &c, QString val)
     c.m_char_data.m_battle_cry = val;
 }
 
+void initializeCharacter(Character &c)
+{
+    GameDataStore &data(getGameData());
+
+    uint entclass = getEntityClassIndex(data, true, c.m_char_data.m_class_name);
+    c.m_char_data.m_current_attribs.m_ToHit             = data.m_player_classes[entclass].m_AttribBase[0].m_ToHit;
```
broxen: Can we simply: `c.m_char_data.m_current_attribs = data.m_player_classes[entclass].m_AttribBase[0]`?

If not, maybe we should refactor this out to `setCurrentAttribsToBase()` or something as we may reuse this for CRITTERS and NPCS down the road.

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -1381,14 +1384,19 @@ void cmdHandler_AddNPC(const QString &cmd, MapClientSession &sess)
     if(parts.size()<2)
     {
         qCDebug(logSlashCommand) << "Bad invocation:"<<cmd;
```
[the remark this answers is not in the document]
cyanpill: I could have sworn I did this at one point, might have been lost in my updating messup, or maybe I didn't actually do it.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
broxen: That makes sense, though we might still want to expose those boundaries to server admins in the future, and I think we should set the static variables for them. `s_min_tohit_chance` and `s_max_tohit_chance`. Eventually accessible in settings.cfg

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -112,34 +141,49 @@ void World::checkPowerTimers(Entity *e, uint32_t msec)
...
+    for (uint i =0; i<e->m_buffs.size();)
     {
-        buff_idx->m_activate_period -= (float(msec)/1000); // activate period is in minutes
+        e->m_buffs[i].m_duration -= (float(msec)/1000); // activate period is in minutes
 
-        if(buff_idx->m_activate_period <= 0)
-            buff_idx = e->m_buffs.erase(buff_idx);
+        if(e->m_buffs[i].m_duration <= 0)
+        {
+            removeBuff(*e, e->m_buffs[i]);
```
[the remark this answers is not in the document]
cyanpill: When the buff is removed, the iterator points to the next value, so it doesn't need to change the iterator, right?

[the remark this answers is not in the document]
broxen: According to the spec (c++11), every iterator and reference after the point of erase is invalidated. So you have to redefine the iterator. I'm referencing here http://www.cplusplus.com/reference/vector/vector/erase/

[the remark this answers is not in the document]
broxen: Makes sense

Projects/CoX/Common/GameData/CharacterHelpers.cpp
nemerle: It almost looks like we could replace `Character &c` argument with Attrib and class name ?

Projects/CoX/Common/GameData/CharacterHelpers.cpp
```diff
@@ -141,6 +141,53 @@ void setBattleCry(Character &c, QString val)
...
+    c.m_char_data.m_current_attribs.m_Repel             = data.m_player_classes[entclass].m_AttribBase[0].m_Repel;
+    c.m_char_data.m_current_attribs.m_Accuracy          = data.m_player_classes[entclass].m_AttribBase[0].m_Accuracy;
+    c.m_char_data.m_current_attribs.m_Radius            = data.m_player_classes[entclass].m_AttribBase[0].m_Radius;
+    c.m_char_data.m_current_attribs.m_Arc               = data.m_player_classes[entclass].m_AttribBase[0].m_Arc;
+    c.m_char_data.m_current_attribs.m_Range             = data.m_player_classes[entclass].m_AttribBase[0].m_Range;
+    c.m_char_data.m_current_attribs.m_TimeToActivate    = data.m_player_classes[entclass].m_AttribBase[0].m_TimeToActivate;
+    c.m_char_data.m_current_attribs.m_RechargeTime      = data.m_player_classes[entclass].m_AttribBase[0].m_RechargeTime;
+    c.m_char_data.m_current_attribs.m_InterruptTime     = data.m_player_classes[entclass].m_AttribBase[0].m_InterruptTime;
+    c.m_char_data.m_current_attribs.m_EnduranceDiscount = data.m_player_classes[entclass].m_AttribBase[0].m_EnduranceDiscount;
+}
```
nemerle: And another note, this almost looks like we might simply write:
```cpp
c.m_char_data.m_current_attribs             = data.m_player_classes[entclass].m_AttribBase[0];
```

Projects/CoX/Common/GameData/EntityHelpers.cpp
```diff
@@ -168,6 +168,107 @@ void toggleMovementAuthority(Entity &e)
...
+        case StoredEntEnum::DeadPlayer:
+            return (ent.m_is_hero ? target_ent.m_is_hero : target_ent.m_is_villian);       // both have to be heroes, or both villians
+    }
+}
+
+// checks a vector of possible targets
+bool validTargets(Entity &target_ent, Entity &ent, std::vector<StoredEntEnum> targets)
+{
+    if (targets.size() == 0)
+        return false;
```
nemerle: the targets are passed-by-value, was it intentional, or can we just use const & there ?
```cpp
if(targets.empty())
    return false;
```
looks a bit better here

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -583,98 +583,250 @@ void sendDoorAnimExit(MapClientSession &sess, bool force_move)
...
+/*
+ * checkPowerTarget checks yourself, the current target, the target of target, and assist target and returns as soon as one is valid.
+ */
+bool checkPowerTarget(Entity &ent, Entity *target_ent, int32_t tgt_idx, Power_Data powtpl)
+{
+    if(validTarget(ent, ent, powtpl.Target))                    //check self targetting first
     {
         target_ent = &ent;
         tgt_idx = ent.m_idx;
     }
```
nemerle: adding `return true` here, and flipping the condition below will make the code a bit simpler

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -583,98 +583,250 @@ void sendDoorAnimExit(MapClientSession &sess, bool force_move)
...
+    }
+    return true;
+}
+/*
+ * checkPowerRecharge returns false if the power is found in the recharging power vector
+ */
+bool checkPowerRecharge(Entity &ent, uint32_t pset_idx, uint32_t pow_idx)
+{
+    if(!ent.m_recharging_powers.empty())
+    {
```
nemerle: the if seems redundant here, since the loop will 'loop' 0 times on empty containers ?

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1273,6 +1517,36 @@ void showMapMenu(MapClientSession &sess)
     showMapXferList(sess, has_location, location, msg_body);
 }
 
+void setAlignment(Entity &e, QString align)
+{
+    if (align.compare("hero", Qt::CaseInsensitive) == 0)
```
nemerle: To make this a bit simpler you can do
```cpp
QString lowcase_align = align.toLower();
```
then all comparisons below can be done with simple '=='.

nemerle: Also, the `align` argument is passed by value here, it should be `const &`

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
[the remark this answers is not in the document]
nemerle: Lists are bad and only used after a lot of thought and testing.
Usualy when walking the container the template to use looks something like.
```cpp
for(auto iter = container.begin(); iter!=container.end(); /* iter updated in loop body*/ )
{
   if(should_delete(iter)
     iter = container.erase(iter);
   else
     ++iter;
}
```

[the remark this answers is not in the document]
broxen: The lack of direct access for forward_lists seems like it would disqualify their use anyway for this case.

We're using the coding paradigm Nem outlined above many places in the code, we might as well stick with it.

Projects/CoX/Common/GameData/power_serializers.cpp
cyanpill: I'm still testing what proper format is needed in the json, without a catch here it errors and doesn't let the server start. I was hoping to separate out name, scale, and maybe a few others as essential, so it would only ignore errors for data that can be ignored. But that didn't work so it will need a better check of validity.

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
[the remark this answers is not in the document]
cyanpill: Ok, back to vector it goes! I was thinking forward list woul be efficient because we should be iterating through the list, never directly accessing, but that could happen in the future. Plus, forward list doesn't have size() which we need for sending the bufflist to the client.

Projects/CoX/Common/Runtime/HandleBasedStorage.h
cyanpill: without that include I get this error:
 ``` /home/cyanpill/Segs/Projects/CoX/Common/Runtime/HandleBasedStorage.h:49:5: error: ‘size_t’ does not name a type```
 ```    size_t liveCount() const { return m_nodes.size(); }```

I have no idea what I'm doing differently, might be a leftover merge issue.

broxen: That is weird. Can you please add a comment about there being a build error on your platform as the reason for it's inclusion. I want to ensure we don't accidentally remove it later, introducing a regression.

Projects/CoX/Common/GameData/power_serializers.cpp
cyanpill: Turns out, I  was getting an error from trying to assign a enum from text, when  it expexts a number. From googling, you can read text in a json into an enum if you do something like:
Qbytearray stacktype;
archive(cereal::make_nvp("StackType",stacktype));
If (stacktype == "replace")
Src.StackType = StackType::Replace;
Which seems a bit akward, but we could check for more variables this way, which makes the issue24 data easier to use. I think I'll have hold this back for a PR devoted to loading data.

Projects/CoX/Common/GameData/Powers.h
cyanpill: I'm mostly sure when you queued a power it stuck with that target, but queuedpower is also used for activated powers that are locked on to a target.

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
cyanpill: This was an attempt to keep powers in the queue to make sure the icons are updated, but I figured out a simpler way.

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
cyanpill: I think this got mangled in my merge snafu, but the idea was to cycle queued  powers in case one could fire off. But there is only supposed to be one queued power anyways, multiple powers being queued does work as well as I thought.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
cyanpill: ClientStates::DEAD is set when hp reaches 0, so this should be simpler than checking hp each time a power is fired. I'm wondering if critters can use the same states? Or maybe there could be an entitystate.

broxen: As far as I can tell, the client doesn't have any sort of entityState. That doesn't mean we can't have `is_dead` server side. I'd put it in chardata though, as there will be plenty of entity types that will never be dead such as cars, Glowies, pedestrians, etc.

I don't remember fully, but I feel like bodies disappear pretty quickly. Could you even target them?

PsiLupan: Bodies could definitely be targeted as Warshade abilities required it, such as Stygian Circle and Dark Extraction. It specifically couldn't have been too quick for that reason.

cyanpill: Also critters could be revived, or self revive, like the freaksow bosses. We will have to find out how long corpses hung around, it was something around 10 seconds.

PsiLupan: https://youtu.be/aJDTIlxi6KA

I counted a little quickly with this video, I think, but it seemed to be about 15 seconds.

They also had the client-side tech back by then. It's possible they could've cleaned up dead critters faster for those without powers that influenced them, or that may be a good way to handle it.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -584,7 +584,7 @@ void sendWaypoint(MapClientSession &sess, int point_idx, glm::vec3 &location)
 
 void sendStance(MapClientSession &sess, PowerStance &stance)
 {
-    qCDebug(logSlashCommand) << "Sending new PowerStance";
+    //qCDebug(logSlashCommand) << "Sending new PowerStance";
```
broxen: As a rule we should avoid commenting out debug output messages. Let's either put it in a logging category where it can be toggled, or if it's truly superfluous, remove it altogether.

Projects/CoX/Common/GameData/power_serializers.cpp
```diff
@@ -242,36 +242,35 @@ bool loadFrom(BinStore *s, AllPowerCategories &target)
 template<class Archive>
 static void serialize(Archive & archive, StoredAttribMod & src)
 {
+        archive(cereal::make_nvp("Name",src.name));
```
broxen: Why the additional indentation for this code block? Indents should be 4 spaces.

Projects/CoX/Common/GameData/EntityHelpers.cpp
```diff
@@ -355,8 +480,8 @@ void unmarkEntityForDbStore(Entity *e, DbStoreFlags f)
 
 void revivePlayer(Entity &e, ReviveLevel lvl)
 {
-    float cur_hp = getHP(*e.m_char);
-    if(e.m_type != EntType::PLAYER && cur_hp != 0)
+   // float cur_hp = getHP(*e.m_char);
```
broxen: if we don't' need this, let's remove it altogether.

Projects/CoX/Common/GameData/EntityHelpers.cpp
```diff
@@ -170,6 +175,126 @@ void toggleMovementAuthority(Entity &e)
 }
 
 
+bool validTarget(Entity &target_ent, Entity &ent, StoredEntEnum const &target)
+{
+    // first check if the target needs to be dead, and isn't dead yet
+    if (target == StoredEntEnum::DeadPlayer || target == StoredEntEnum::DeadTeammate || target == StoredEntEnum::DeadVillain)
+        if (target_ent.m_char->getHealth() != 0.0f)             // target.dead() maybe?
```
broxen: There is no `getHealth()` method. I think you intend to use `getHP()` here.

Projects/CoX/Common/GameData/Entity.h
```diff
@@ -170,14 +171,15 @@ class Entity
         glm::quat           m_direction;
         int32_t             m_target_idx            = -1;
         int32_t             m_assist_target_idx     = -1;
+        glm::vec3           m_target_loc;
 
         std::vector<Buffs>          m_buffs;
         QQueue<QueuedPowers>        m_queued_powers;
+        std::vector<QueuedPowers>   m_auto_powers;
```
broxen: I think the client looks at queuedPowers and AutoPowers the same way. This would explain why the client only accepts one array for both.

Perhaps it makes sense for us to separate server side, but it may explain some behaviors that you're seeing elsewhere with the client.

Projects/CoX/Common/GameData/power_serializers.cpp
cyanpill: The changes to this file will be reverted.

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2151,6 +2151,8 @@ void MapInstance::on_client_resumed(ClientResumedRendering *ev)
         map_server->session_xfer_complete(session.link()->session_token());
     }
 
+    initializeCharacter(*session.m_ent->m_char);
```
broxen: I think this is probably the wrong place for this. Players shouldn't always spawn with full health and endurance. By executing this here, it will apply even when a player transfers maps.

Projects/CoX/Servers/MapServer/DataHelpers.h
```diff
@@ -109,10 +111,24 @@ void sendMissionObjectiveTimer(MapClientSession &sess, QString &message, float t
...
+void applyInspirationEffect(Entity &ent, uint32_t col, uint32_t row);
+bool useInspiration(Entity &ent, uint32_t col, uint32_t row);
 void increaseLevel(Entity &ent);
-
-
+bool checkPowerTarget(Entity &ent, Entity *target_ent, int32_t tgt_idx, Power_Data powtpl);
+bool checkPowerRecharge(Entity &ent, uint32_t pset_idx, uint32_t pow_idx);
+bool checkPowerRange(Entity &ent, Entity &target_ent, uint32_t range);
+bool checkPowerRange(Entity &ent, int32_t tgt_idx, uint32_t pset_idx, uint32_t pow_idx);
+uint toHitLimit(uint value);
```
broxen: For code consistency, please use `uint32_t`

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1570,7 +1811,25 @@ void sendLocation(MapClientSession &cl, VisitLocation location)
...
+{
+    if (!e.m_char->m_is_dead)
+    {
+        e.m_char->m_char_data.m_current_attribs.m_HitPoints = std::min(val,e.m_char->m_max_attribs.m_HitPoints);
+        if(e.m_char->m_char_data.m_current_attribs.m_HitPoints < 0.1f)
+        {
+            e.m_char->m_is_dead = true;
+            e.m_char->m_char_data.m_current_attribs.m_HitPoints = 0.0f;
 
+            if(e.m_type == EntType::PLAYER)
```
broxen: We can probably include CRITTERS here.

Projects/CoX/Common/GameData/EntityHelpers.cpp
```diff
@@ -170,6 +175,126 @@ void toggleMovementAuthority(Entity &e)
...
+    }
+    else if (name == "immobilized") //TODO: copy for hold, sleep,stun
+    {
+        e.m_char->m_char_data.m_current_attribs.m_Immobilized += value;
+        checkMovement(e);
+    }
+    else if (name == "onlyaffectsself")
+    {
+        e.m_char->m_char_data.m_current_attribs.m_OnlyAffectsSelf += value;
+        checkMovement(e);
```
broxen: Seems like we can probably just `checkMovement(e)` and `resetSpeed(e)` once at the end of `modifyAttrib()`

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -800,133 +902,240 @@ void usePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, int32_t tgt_idx,
...
-    // If not a player, skip the rest
-    if(target_ent->m_type != EntType::PLAYER)
-        return;
+    if(insp == nullptr)
+        return true;
 
-    // Send message to target
-    assert(target_ent->m_client);
+    QStringList revive_names = {"Awaken","Bounce_Back","Restoration",};
+    if (( ent.m_state_mode == ClientStates::DEAD) xor revive_names.contains(insp->m_name, Qt::CaseInsensitive))
```
broxen: Appveyor doesn't seem to like this. Maybe use `^` instead

Projects/CoX/Servers/MapServer/DataHelpers.cpp
cyanpill: Do non-players have state mode and motion states? It didn't seem to cause problems without the player check, but I wanted to be safe.

broxen: I don't see anything that would cause an issue with CRITTERS specifically.

Projects/CoX/Servers/MapServer/MessageHelpers.cpp
```diff
@@ -536,21 +536,41 @@ void storePowerInfoUpdate(BitStream &bs,Entity *e)
     storePowerRanges(*cd, bs); // sending state of all current powers.
 
     qCDebug(logPowers) << "NumQueuedPowers:" << e->m_queued_powers.size();
-    bs.StorePackedBits(4, e->m_queued_powers.size()); // Count all active powers
-    for(const QueuedPowers &qpow : e->m_queued_powers)
+    bs.StorePackedBits(4, uint( uint(e->m_queued_powers.size()) + e->m_auto_powers.size())); // Count all active powers
```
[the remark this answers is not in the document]
cyanpill: apparently QQueue.size is an int, don't know why it's allowed to be negative size, but it gives a warning unless I cast that specifically to uint.

Projects/CoX/Common/GameData/EntityHelpers.cpp
cyanpill: Huh, that has been there a while. I meant to update this to check isdead anyways.

Projects/CoX/Servers/MapServer/MapInstance.cpp
cyanpill: There have been a few changes to character loading, I put this in when we had the bug that set you to 0hp every time you zoned. It seems to work properly now, so I'll remove it.

ZopharShinta:
I tested this real quick and the server would crash when you defeated an enemy. 
Editing line 43 in WorldSimulation.cpp seems to of corrected it. 

`    for(Entity * e : ref_ent_mager.m_live_entlist)
    {
        updateEntity(e,delta);
        if(e->m_destroyed)
            break;
    }`

Projects/CoX/Common/GameData/Entity.h
cyanpill: Aha, the problem was I was sending auto powers to the client as activated, which confused the client because they should never be deactivated. I think I have a good way to fix that, which will also make all of the queuing much simpler...

Projects/CoX/Common/GameData/Entity.h
```diff
@@ -169,12 +169,12 @@ class Entity
         uint32_t            m_db_id                 = {0};
         EntType             m_type                  = {EntType::Invalid};
         glm::quat           m_direction;
-        int32_t             m_target_idx            = -1;
-        int32_t             m_assist_target_idx     = -1;
+        uint32_t            m_target_idx            = 0;
```
nemerle: This is pretty interesting change, until now, we were using -1 as an 'INVALID' marker, and index 0 seems to be a possibly valid one ?

cyanpill: As far as I can tell, 0 was also used for invalid targets/no target/target self, everything seems to work the same with this change.

nemerle: Good enough for me :)

Projects/CoX/Common/GameData/GameDataStore.cpp
```diff
@@ -489,28 +489,136 @@ bool GameDataStore::read_settings(const QString &/*directory_path*/)
...
+        qDebug() << "Failed to load power data from powers.bin!";
+    }
 
-    // Hardcoding of stats to test powers
     StoredAttribMod temp;
     Power_Data *temppower = nullptr;
 
-    temp.name = "Damage";
-    temp.Magnitude = 5;
+    temp.Scale = 5;
```
nemerle: We could really use an array of StoredAttribMod here + an array of int tuples
```cpp
static const StoredAttribMod s_per_pow_mods[] = {
  { 5,1.0f,"regeneration},
};
static const PowerPool_Info s_per_pow_infos[] = {
{26,0,7},
};
```
and then
```cpp
for(int i=0; i<TEMP_POWER_COUNT; ++i)
{
  temppower = editable_power_tpl(s_per_pow_infos[i]);
  temppower->pAttribMod.push_back(s_per_pow_mods[i]);
}```
It would make moving to a 'load things from file/script' much easier.
The first step would be to simply load the powers to those two s_per_pow from a crl file ;)

cyanpill: I'm working on loading from a file, when that's done all of this can be removed. Unless we want to keep some data for when it can't find the power file?

broxen:
This just needs to be pulled down and tested prior to merge.

broxen:
I'm experiencing some issues with this PR when using certain powers.
- Using `Prestige Power Surge` (a sprint power) causes user to lose 10 hp per tick
- Using `Rest` causes user to lose 11.49 hp per tick
- Using `Hover` causes user to lose 13.96 hp per tick

Powers that affect `Practice Drone`s seem to work fine. It's just every power that affects `self`

PR #697, opened 2018-11-17
Hash all spawners and use them to spawn and mapxfer

### LINKED ISSUE #469: Server is looking for case-sensitive CHUNKS bin.
Server is looking for case-sensitive CHUNKS bin. It probably has no issue on Windows since windows ignores case in filenames.

```
Critical: Node "swr_trialroom_main_1461&" 
can't find member "object_library/Villain_Lairs/sewers/elements/CHUNKS/swr_chunks_01"
```

broxen:
## Summary
Correct an issue with typecase when loading bin files, and load all spawners into a MultiHash that we can reference for spawning.

### Issues closed
Closes #469 
Resolves falling under the map

- Ignore typecase when loading bin files
- Load all spawners on a map into a MultiHash
- Use MultiHash to spawn new characters and properly zone transfer
- Couple minor code cleanups

### DIFF
--- a/Projects/CoX/Common/GameData/scenegraph_serializers.cpp
+++ b/Projects/CoX/Common/GameData/scenegraph_serializers.cpp
@@ -374,25 +374,53 @@ bool loadFrom(const QString &filepath, SceneGraph_Data &target)
 }
 bool LoadSceneData(const QString &fname, SceneGraph_Data &scenegraph)
 {
+    // Windows is far too lax about case sensitivity. Consequently
+    // filenames aren't consistent. This should derive the filename
+    // based upon a case-insensitive comparison, and use the actual
+    // formatted filepath when loading scene data.
+    QString fixed_path = fname;
+
+    // check file exists, if not, check filecase
+    if(!QFile(fixed_path).exists())
+    {
+    // get base from path
+    QStringList orig_path = fixed_path.split(QDir::separator());
+
+    if(orig_path.size()>1)
+        orig_path.pop_back(); // remove file name
+
+    QDir dir(orig_path.join('/'));
+    if (!dir.exists())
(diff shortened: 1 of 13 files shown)


Projects/CoX/Common/GameData/scenegraph_serializers.cpp
```diff
@@ -374,25 +374,53 @@ bool loadFrom(const QString &filepath, SceneGraph_Data &target)
...
+    // based upon a case-insensitive comparison, and use the actual
+    // formatted filepath when loading scene data.
+    QString fixed_path = fname;
+
+    // check file exists, if not, check filecase
+    if(!QFile(fixed_path).exists())
+    {
+    // get base from path
+    QStringList orig_path = fixed_path.split(QDir::separator());
+
```
nemerle:
```cpp
QString base_path = QFileInfo(fixed_path).path();
QDir dir(base_path);
```
should make this much simpler :)

Projects/CoX/Common/GameData/scenegraph_serializers.cpp
```diff
@@ -374,25 +374,53 @@ bool loadFrom(const QString &filepath, SceneGraph_Data &target)
...
+        QStringList files = dir.entryList(QDir::Files | QDir::NoSymLinks);
+        for(QString &f : files)
+        {
+            //qWarning() << "Comparing" << f << fixed_path;
+            if(fixed_path.endsWith(f, Qt::CaseInsensitive))
+                fixed_path = orig_path.join('/') + "/" + f;
+        }
+    }
+
```
nemerle: Another note, the `locate_file_case_insensitive()` probably should be extracted as a more standalone function, since we might want to use it elsewhere ? + it would make the code a bit cleaner.

PR #698, opened 2018-11-19
mysql: Unremove `costume` table removal

dracc:
Updating a MySQL-server from pre-tailor PR to current develop HEAD crashes because of foreign key constraints no longer being taken care of in the correct order.

Having references to old tables which shouldn't exist anymore might feel silly, but it makes the admin life easier.


broxen:
Thank you for doing this. Is there a way to comment the mysql file with this explanation?

`-- comment here` I think?

PR #700, opened 2018-11-20
Force Client Logout / Early test

ZopharShinta:
Added the code to LuaBot to kick the player that selected it.
Will need to add the option to enter player name or show listed of players, like the mapmenu?


Projects/CoX/Common/Messages/Map/ForceClientLogout.h
```diff
@@ -0,0 +1,38 @@
...
+public:
+    // [[ev_def:field]]
+    QString reason;
+    ForceLogout() :GameCommandEvent(MapEventTypes::evForceLogout) {}
+    ForceLogout(const QString &_reason) :GameCommandEvent(MapEventTypes::evForceLogout),reason(_reason)
+    {}
+    void serializeto(BitStream &bs) const override
+    {
+        bs.StorePackedBits(1, type()-evFirstServerToClient); //packet 12
+        //bs.StorePackedBits(1,12); // opcode
```
broxen: You can remove this. We wont need it.

Projects/CoX/Common/Messages/Map/ForceClientLogout.h
```diff
@@ -0,0 +1,38 @@
...
+namespace SEGSEvents
+{
+// [[ev_def:type]]
+class ForceLogout final : public GameCommandEvent
+{
+public:
+    // [[ev_def:field]]
+    QString reason;
+    ForceLogout() :GameCommandEvent(MapEventTypes::evForceLogout) {}
+    ForceLogout(const QString &_reason) :GameCommandEvent(MapEventTypes::evForceLogout),reason(_reason)
```
broxen: Since you're in namespace SEGSEvents, you don't actually need `MapEventTypes::` before `evForceLogout`

Projects/CoX/Common/Messages/Map/ForceClientLogout.h
```diff
@@ -0,0 +1,38 @@
...
+    {}
+    void serializeto(BitStream &bs) const override
+    {
+        bs.StorePackedBits(1, type()-evFirstServerToClient); //packet 12
+        //bs.StorePackedBits(1,12); // opcode
+        bs.StoreString(reason); // opcode
+    }
+    void serializefrom(BitStream &bs) override
+    {
+        bs.GetString(reason);
```
broxen: Since we should never receive this, you can remove this line and add a qWarning:
`qWarning() << "Server does not receive ForceLogout packet from Client.";`

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1273,4 +1274,10 @@ void addSouvenir(MapClientSession &cl, Souvenir souvenir)
     cl.addCommand<SouvenirListHeaders>(souvenir_list);
 }
 
+void sendForceLogout(MapClientSession &cl, const char* message)
+{
+    QString logout_message = QString::fromUtf8(message);
+    qCDebug(logScripts) << "SendForceLogout. Mesage: " << logout_message;
```
broxen: Typo in `SendForceLogout. Message`

broxen:
We should probably add a slash command to `/force_logout $target`

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -2150,6 +2153,21 @@ void cmdHandler_ReloadScripts(const QString &/*cmd*/, MapClientSession &sess)
...
+    parts = cmd.splitRef(' ');
+
+    if(parts.size() < 2)
+    {
+        qCDebug(logSlashCommand) << "Bad invocation: " << cmd;
+        sendInfoMessage(MessageChannel::USER_ERROR, "Bad invocation:" + cmd, sess);
+        return;
+    }
+
+    sendForceLogout(sess, parts[1].toUtf8(), parts[2].toUtf8());
```
broxen: This will cut off longer messages like `/forceLogout $target "I'm kicking you with this longer message, but it will get cut off, even if I wrap it in quotes"` as parts[2] will equal `"I'm`

One solution would be to follow the format of `cmdHandler_InfoMessage()` to capture anything after the 2nd space in the string.

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -2150,6 +2153,21 @@ void cmdHandler_ReloadScripts(const QString &/*cmd*/, MapClientSession &sess)
...
 }
 
+void cmdHandler_ForceLogout(const QString &cmd, MapClientSession &sess)
+{
+    QVector<QStringRef> parts;
+    parts = cmd.splitRef(' ');
+
+    if(parts.size() < 2)
+    {
+        qCDebug(logSlashCommand) << "Bad invocation: " << cmd;
```
broxen: I apologize, because I set the example with this "bad invocation" output, but the truth is that we should be more verbose with our error feedback.

I would describe the format this slash command expects so that users know what they *should* type

PR #703, opened 2018-11-20
lua: Set scripting engine default enabled

### LINKED ISSUE #701: ENABLE_SCRIPTING_ENGINE should be ON by default
## Summary
Now that we're shipping with some default scripts (luabot!) we should probably just enable scripting engine support by default.

Reported by Leandro on Discord

dracc:
Closes #701.


broxen:
Thanks @dracc, you rock brotha!

PR #710, opened 2018-11-23
Lua / Contacts layout changes / LuaBot fixes

### LINKED ISSUE #685: Lua bot menu glitch
the lua bot seems to have problems with keeping track of what page it's on when accessing all of the options in the menu, such as somtimes when going into player stats and leaving the player stats page it still acts as if it were on the player stats page.

 when clicking temp power somtimes the game opens it up as mapmenu so there seems to be a lot of problems on what option it is actually on compared to what it thinks it's on.

ZopharShinta:
## Summary
Adds Contacts.Lua to manage Lua scripts for contacts
Added respawn/revive options to Lua. 
Updates to LuaBot:
Works in Outbreak and Atlas Park.
Bit test mission will send you to Atlas Park or Outbreak depending on where you started it. (two scripts for this, one for each zone.
Can now be used by multiple players at same time.

### Issues closed
#685
Should fix the current issues with LuaBot behavior.

We need doors linked so you can leave the hospital after reviving.


Projects/CoX/Data/scripts/City_01_01/luaBot.lua
```diff
@@ -0,0 +1,566 @@
...
+            if (id == 1) then
+                Player.GiveRandomInsp()
+            elseif (id == 2) then
+                Player.GiveRandomEnhancement(1)
+            elseif (id == 3) then
+                contactStatus.currentStanding = 0
+            elseif (id == 25) then
+                --Since LevelUp closes all dialogs, we need to reset current status
+                contactStatus.currentStanding = 0
+                Player.SetTitle("You're a hero!")
```
broxen: Can this call the set title window instead?

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1273,4 +1297,133 @@ void addSouvenir(MapClientSession &cl, Souvenir souvenir)
     cl.addCommand<SouvenirListHeaders>(souvenir_list);
 }
 
+void setEndurance(MapClientSession &sess, float end)
```
broxen: In lieu of redefining these helper methods here, can we simply correct the way we register the type in ScriptingEngine to pass the character object from the session token? By doing so we'd eliminate this "middle man" method.

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -314,66 +308,32 @@ void ScriptingEngine::registerTypes()
...
-                qCDebug(logScripts) << "giveTempPower Name: " << pset.m_Powers[i].m_Name;
-                if(pset.m_Powers[i].m_Name == power)
-                {
-                    qCDebug(logScripts) << "giveTempPower PowerFound";
-                    ppool.m_pow_idx = uint32_t(i);
-                    addPower(cd, ppool);
-                    break;
-                }
-            }
+            QString title_name = QString::fromUtf8(title);
```
broxen: This is how we should handle QString from Lua moving forward. It will allow us to eliminate a lot of duplicated code in DataHelpers and reuse that code for ScriptingEngine.cpp and SlashCommands.cpp

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -314,66 +308,32 @@ void ScriptingEngine::registerTypes()
                 listToSend.push_back(kvp);
             }
             addListOfTasks(cl, listToSend);
-        },
+        },*/
         "addTask", sendUpdateTaskStatusList,
         "removeTask", removeTask,
         "selectTask", selectTask,
         "setWaypoint", sendWaypoint,
-        "setHp", [](MapClientSession *cl, float hp)
```
broxen: This was a better way to go, no? It eliminates the code duplicity in DataHelpers?

Projects/CoX/Data/scripts/City_00_01/luaBot.lua
```diff
@@ -377,13 +279,289 @@ LuaBot.callback = function(id)
...
+            button02 = {"Increase standing","CONTACTLINK_MAIN"},
+            button03 = {"Lower standing","CONTACTLINK_MISSIONS"}, 
+            button04 = {"Add Call button","CONTACTLINK_LONGMISSION"} ,
+            button05 = {"Remove Call button","CONTACTLINK_SHORTMISSION"} ,
+            button06 = {"Remove Contact","CONTACTLINK_ACCEPTLONG"} ,
+            button07 = {"Back","CONTACTLINK_INTRODUCE"}
+        }
+    } 
+    
+   --[[ LuaBot.contactDialogs[5] = {
```
broxen: Do we need this comment?

Projects/CoX/Data/scripts/City_00_01/contacts.lua
```diff
@@ -0,0 +1,125 @@
+printDebug('Loading contacts.lua...')
```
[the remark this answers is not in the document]
ZopharShinta: The functions are all helpers that could go into the global.lua but the contacts object, I expect to be limited to the zone.
Or do we want each contact(s) script available for each zone?

Projects/CoX/Data/scripts/City_00_01/locations.lua
```diff
@@ -5,3 +5,19 @@ location_visited = function(name, location)
...
+enter_door = function(name, location)
+    printDebug("enter_door name: " .. name .. " Loc X: " .. round(location.x, 4) .. " Y: " .. round(location.y, 4) .. " Z: " .. round(location.z, 4))
+
+    local x = round(location.x, 0)
+    local y = round(location.y, 0)
+    local z = round(location.z, 0)
+
+    printDebug("Rounded Location:  X: " .. tostring(x) .. " Y: " .. tostring(y) .. " Z: " .. tostring(z))
+
+    if(x == 36) then
```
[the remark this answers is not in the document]
ZopharShinta: Oh this was the beginning of my old if statements for the hospital doors.
I'll remove it as it isn't needed.

Projects/CoX/Data/scripts/City_00_01/luaBot.lua
```diff
@@ -1,215 +1,136 @@
...
-    if (LuaBot.id ~= nil) then
-        if(LuaBot.settingTitle) then -- Title selection buttons trigger dialog_button
+    if (contactStatus ~= false) then
+        if(contacts.LuaBot.settingTitle) then -- Title selection buttons trigger dialog_button
             LuaBot.settingTitle = false
-            LuaBot.Mode = nil
+            contactStatus.currentStanding = 0
             if(id == 1) then
-                LuaBot.Mode = 21 -- "Choose later" on set title will show previous screen.
+                contactStatus.currentStanding = 2 -- "Choose later" on set title will show previous screen.
```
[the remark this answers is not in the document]
ZopharShinta: It is the players standing with the contact. That was the easiest variable to use to help navigate the contact screens.
We could create another variable for this. We just have to update the contact object in the server so it is stored in the database.

Projects/CoX/Data/scripts/City_00_01/contacts.lua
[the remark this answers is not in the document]
broxen: Seems like the same methods in the contacts.lua for Atlas Park. I'm not as familiar with Lua, but seems like we could pass scope per zone?

Projects/CoX/Data/scripts/City_01_01/luaBot.lua
ZopharShinta: That is exactly what it does. The setTitle method probably need to be adjusted to clear up what is actually being done.

```
void setTitle(MapClientSession &sess, QString &title)
{
    bool select_origin = false;
    if(!title.isEmpty())
        select_origin = true;

    sendContactDialogClose(sess);
    sendChangeTitle(sess, select_origin);
}
```

Projects/CoX/Servers/MapServer/DataHelpers.cpp
ZopharShinta: Originally we had lambdas for this however we ran into the issue of too many characters in declaring the usertype. So we had to remove some of them. I tried to reference the `sess.m_ent->m_char` in lua by `client.m_ent.m_char` the `m_char` is not defined. I could pass m_ent but then I still need a helper method.
I'm open to suggestions on this one.

Projects/CoX/Data/scripts/City_00_01/luaBot.lua
ZopharShinta: No. I held on to it just in case I needed it. I'll remove it.

Projects/CoX/Data/scripts/City_00_01/contacts.lua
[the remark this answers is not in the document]
ZopharShinta: I believe the scope is defined when we load the scripts for the first time in that zone.
If you leave and return to that zone the old scope for that zone should be returned.

I think the simple answer would be to move the functions that are "global" to the global.lua file under a helper namespace or something but leave the contacts.lua file there with just this.

```
printDebug('Loading contacts.lua...')

contacts = {}

-- Begin Contact Scripts
include_lua('luaBot.lua')
-- End Contact scripts

printDebug('Finished Loading contacts.lua')
```
That way we can use `include_lua('CustomContactScript.lua')` to add/run any contact scripts for that zone.
The thought is each contact script would create a local object with everything that contact needs then add it into the `contacts` object for that zone.
The concern I have with moving all of it into global is I'm expecting custom contact scripts to include where the NPC should spawn. So if you're in the wrong zone, that NPC might spawn in the incorrect spot.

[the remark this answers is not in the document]
broxen: That makes perfect sense. I agree.

Your proposed solution sounds great. Move the generic methods to global. Zone specific code should remain.

Thanks Zophar

Projects/CoX/Data/scripts/City_00_01/luaBot.lua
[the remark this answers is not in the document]
broxen: Got it. Seems like it works for Luabot. Do you foresee any issues with other custom NPCs down the line?

Projects/CoX/Data/scripts/City_01_01/luaBot.lua
broxen: Agreed. We can modify setTitle later to make it more clear.

Thanks again Zophar!

Projects/CoX/Servers/MapServer/DataHelpers.cpp
broxen: Ahh I understand.

Let me do a bit of research. I wonder if we can add those usertypes individually, instead of inline. Otherwise, perhaps we can expose Character object to Lua and call it that way.

It's not the end of the world to have the wrapper methods, just seems silly (and possibly confusing for future developers) to have two methods doing the same thing (setHP for example)

Projects/CoX/Common/GameData/Contact.cpp
```diff
@@ -39,6 +39,7 @@ void Contact::serialize(Archive &archive, uint32_t const version)
     archive(cereal::make_nvp("CanUseCell",m_can_use_cell));
     archive(cereal::make_nvp("HasLocation",m_has_location));
     archive(cereal::make_nvp("Location",m_location));
+    archive(cereal::make_nvp("CurrentDialogScreen",m_dlg_screen));
```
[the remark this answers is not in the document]
ZopharShinta: You're right. If the scripter wants to certain dialog after a logoff/zone change they should probably use the `currentStanding` flag on the contact.
I'll remove it.

Projects/CoX/Data/scripts/City_01_01/entities.lua
```diff
@@ -1,111 +1,66 @@
 --- ATLAS PARK
-
-LuaBot2 = {}
-LuaBot2.spawned = false;
-LuaBot2.contactDialogs = {}
-
+include_lua('luaBot.lua')
```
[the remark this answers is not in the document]
ZopharShinta: Nope. Should only be in contacts.lua. Copy pasta error.

Projects/CoX/Data/scripts/City_01_01/luaBot.lua
```diff
@@ -0,0 +1,691 @@
...
+
+LuaBot.callback = function(id)
+    printDebug("LuaBot Callback: " .. tostring(id));
+    local contactStatus = Contacts.FindContactByName("LuaBot");
+    if (contactStatus ~= false) then
+        local button = Contacts.ParseContactButton(id);
+        printDebug("LuaBot.Callback Button String: " .. tostring(button));
+
+        if(button ~= false and button ~= "CONTACTLINK_BYE") then
+            if(contactStatus.settingTitle ~= true) then
```
broxen: If we're setting the value of the contact member variable, doesn't it set for all players? Should this be in Character instead of Contact?

ZopharShinta: The `contactStatus` is a contact object, for a contact in that player's list.
`local contactStatus = Contacts.FindContactByName("LuaBot");`
Finds the contact named `LuaBot` in the players contact list.

I added `settingTitle` to the contact object because if its stored in the Lua zone context then it could make the dialogs out of sync for other players.

broxen: Ahh that makes sense then.

I'm gong to merge this. Great work Zophar! Huge leap forward

PR #711, opened 2018-11-24
Make some doors work.

ghost:
## Summary
Doors with nearby GotoSpawn properties will attempt to send the player to that SpawnLocation, but only in the current map.

- This is terrible, since it looks for the properties every time the door is clicked. Not a terrible deal since doors aren't being clicked constantly, but doors should look for the nearest properties once and store them for quick retrieval.
- Doors that link to other zones use the property GotoSpawn which uses the text name of the destination zone to link them together. After the map transfer, the player should spawn in the SpawnLocation in the destination zone.


leandrotlz:
This breaks mapmenu! I hadn't realized that it uses a door trigger. Leaving the pull request up, but don't merge until I fix that.

broxen (approved):
I'm ok merging this as is. We'll need to patch this to load those gotospawns once, like we do Spawners, and hash it for lookup later.

PR #719, opened 2018-12-03
Send VisitLocations

ZopharShinta:
Sends packet 63.
Tested and it doesn't seem to affect contacts or tasks.
Setting a location on the map and walking over it does not trigger the `on_location_visited` call.
So the purpose is still unknown.

Broxen thinks it might be tied to glowies.
Once we have mission maps and glowies we should test it there.


Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1490,7 +1491,21 @@ void sendForceLogout(MapClientSession &cl, QString &player_name, QString &logout
...
+}
+
+void sendLocation(MapClientSession &cl, VisitLocation location)
+{
+    qWarning("SendLocation, unknown affect on client");
+    qCDebug(logSlashCommand) << "sendLocation: " << location.m_location_name;
+    vLocationList location_list;
+    location_list.push_back(location);
+
+    VisitLocation l2;
```
broxen: I know this is really only temporary, but I'd prefer a more verbose name for this variable. `l2` is too vague

PR #725, opened 2018-12-22
Code cleanups and attempted build fix

broxen:
## Summary
Cleaning up code and attempting to fix build errors.


broxen:
Looks like the build error relates to GLM + gcc 7

Nothing specific changed on our end, so Travis must have updated the toolchain somehow? Might need to use the latest GLM, or maybe use a newer gcc version.

PR #734, opened 2019-01-05
I24 powers

Baaleos:
## Summary
Contains json data for powers taken from I24
We will probably need to use these and work backwards towards issue 1-esq powers.
Figured it would be better to have these than not have them

### Issues closed
None

I've separated and grouped the power jsons via powerset
Note that the powerset itself gets a json as well.
This data was acquired from mids hero designer database files with issue 24 data.
There is another folder included which has datadumps from Red Tomax's City of Data


broxen:
Looks like it's currently one power per file? Might be easier to load if all the powers for each powerset are in one json?

So: empathy.json, energy-blast.json and so on.

Baaleos:
I toyed with that idea, but given that a single power is around 1000 lines long, merging an entire powerset could make the json's unreadable and increase likelihood of human error. (Huge JSON files and malformed etc)
It might be more manageable if they stay granular - and then maybe a build process can then merge them at a later time?
Just thinking out loud

cyanpill:
The code to import the powers from JSON hasn't been written yet, so it could go through a powers folder, that shouldn't add much overhead. But users won't want to edit the JSON directly they will want a GUI interface.
The names are slightly different from what we are using, such as "probability" instead of "chance". The import could deal with that.

Baaleos:
Perhaps a web-interface would be possible?
Eg: Select from dropdown the powerset, then from a further dropdown the individual power you want to edit.
Some sort of node based web editor?

dracc:
@Baaleos; I'd rather see a Qt based interface as that's a dependency we're dealing with already. But if you feel like creating one in node.js - please feel free to do so!

nemerle:
I do agree with @broxen, the place for the power data itself should be in external repo. 
As for the multiple-vs-single file - from the experience I'd say that loading a single large is much faster.
I'd envision a small power pre-processor that knows how to load the provided per-power jsons and produces a binary file ( cereal binary format ) that is loaded by the server at runtime.

broxen:
I'd rather see a QT based power editor too, but certainly a web-based solution would be welcome in addition.

The tool should be able to spit out the powers.bin that Nemerle described.

@nemerle, can you clarify if you want to see all of this JSON data in the segs-community repo, or just the RedTomax source data and datadumps?

I personally don't see any issue merging here, as this info was publicly available I12+, but want to make sure we're on the same page.

nemerle:
@broxen : whatever the format will be, the original client data or any of it's simple transformations, should not be put in the official repository. As for the actual format of stored data, I'd say the format could be anything that's computer-readable, as long as we can easily write a simple translation tool to SEGS-power-description-format ( whatever it might end up looking as ).

broxen:
Following this guidance, let's close this PR and move all the json files to the segs-community repo.

Moving forward appears to be up for discussion:
In my opinion, supporting the I23+ powers.bin format makes the most sense. Most users will be able to simply drop the powers.bin file into their server data folder to realize a shutdown-era power system. 

Customization can come in the form of json scripts that can be converted with a tool back into the bin format or an editor that can modify the bin directly.

I agree that fewer files is better, or even just  one file, as I don't think that users will be modifying each power manually with notepad, but perhaps I'm mistaken.

I'm not sure what best (or common) practice is here, so opinions welcome.

cyanpill:
Can we support issue23+ bins? At the same time as issue1? I guess we will have at some point. And the issue 24 client is easy to access.

I've successfuly loaded the json conversion of the issue1 powers. I didn't time it, but it seemed less than a second. That was for a 10meg file. So I think loading times for the human readable text should be fine.

The loading of "pattribmods" will take need tweaks, especially if we want to support the later issues. I am assuming there is a way to load data with multiple possible names.

nemerle:
Oh, I don't doubt we could at least partially support I23/I24 bins, but the amount of work could be substantial. The main problem as always lays with the client and the assumptions it forces on us.

broxen:
I figure we'd want to eventually with a custom client anyway. The problem will be that the client won't have powers at the appropriate idx when it looks up the idxs we send it.

The immediate challenge is providing a file or set of files to server admins that contains the necessary powers data to accomplish combat. Right now we're hard coding values into the server which isn't ideal.

Baaleos:
Closing - 
Is there any tools out there that can serialise / deserialise the powers.bin format?
I can work on some means of reading / writing that format out for power customisation by server owners.

PR #744, opened 2019-01-28
rpc: Make jcon-cpp a submodule

dracc:
## Summary
jcon-cpp is a living project, albeit slowly moving.
We should keep up with its development by submoduling it.


dracc:
I'm not sure, but I'm guessing Travis CI fails because it doesn't initialize the submodules?

broxen:
Hi dracc! Can you resolve the merge conflict and we'll try to get to the bottom of why Travis is failing.

dracc:
The fail is due to changes in our local versions cmake file. Thought I had given up and closed this one already.
But I guess we could supply our own CMakeLists somehow?

broxen:
The objective seems desirable (make jcon-cpp a submodule). What's the reason for abandoning it?

dracc:
Only reason for me giving up on this is that my CMake skills are lacking at best.

broxen:
Saw a discussion with Nem about maybe modifying out local copy of this library. Is they still the plan? If so, let's close this PR

dracc:
Yes, that is correct @broxen. Closed.

Issue #742, opened 2019-01-23
Add [AdminRPC] section to SEGSAdmin and Settings_template.cfg


miles200:
- Add option to SEGSAdmin to allow users to 'Publish' their server online status to SEGSLauncher (if included in the community json. if yes = add [AdminRPC] section with location_addr
- Add [AdminRPC] and location_addr to settings_template.cfg, commented out with instructions on what it does.

broxen:
This was complted by #745. Great work team!

Issue #692, opened 2018-11-15
Enable configuration for WebSockets RPC address and/or port and SSL support


whitegryphon:
## Feature Summary
When installed in a multi-server environment, for example when the game server, database server, and web server all reside on different servers, the web UI will not be able to communicate with the RPC server since it binds to 127.0.0.1 by default. SSL should also be enabled to allow secure WebSocket communication (WSS). Besides being a Good Idea, WSS is required to allow web pages served over HTTPS to access the WebSocket resource. I believe this is the case whether the web server is local to the game server or external.

### Is your feature request related to a problem? 
Related to #633

### Proposed Implementation
Add section and configuration settings to settings.cfg to control the interface address used to listen on for the RPC server. Also add configuration in AdminRPC to enable (and enforce) SSL communication.

Issue #633, opened 2018-10-14
Create an RPC call for exposing server stats for web admin UI


miles200:
Having a play around with this and creating for me to track

valcryst:
I would like to suggest some AdminRPC functions for the purpose of monitoring.
It could be used to create statistics/graphs to visualize utilization or debug the servers
behaiviour under different circumstances.

It could be something like this:

| Command                             | Description                                                                            |
|-------------------------------------|----------------------------------------------------------------------------------------|
| AdminRPC::get_users_online_count    | Current online users count as number                                                   |
| AdminRPC::get_users_login_count     | User logins since server started as number                                             |
| AdminRPC::get_users_registered      | How many users are registered on this server as number                                 |
| AdminRPC::get_users_characters      | How many characters are created on this server as number                               |
| AdminRPC::get_users_connection_rtt  | A list of the rtt of all connected users (output: user_id rtt)                         |
| AdminRPC::get_users_connection_time | A list of the connection time of all connected users (output: user_id time_in_seconds) |

broxen:
@Magentah and here again, per our discord

mobbyg:
Update: currently the calls available right now are:

- Server Status : Show if online or not.
- Server Uptime
- Server Version and Version Name

There is a thread in the discord for more suggested calls to be added to AdminRPC in the development-forum threads.

PR #745, opened 2019-01-30
Extend AdminRPC server and update settings_template.cfg

whitegryphon:
## Summary
Update settings_template.cfg with AdminRPC section, with default values and instructions for use. Also added ability for user to specify RPC server type (tcp or websockets) that will be created by AuthServer. 

### Issues closed
Does not fully close any issue.

These features are referenced in issues #633, #692, and #742.

- Added:
Config section in settings_template.cfg for AdminRPC settings. 
- Modified:
`Projects/CoX/Servers/AuthServer/AdminRPC.cpp` and `Projects/CoX/Servers/AuthServer/AdminRPC.h` to support selection of RPC server type via config file.

### DIFF
--- a/Projects/CoX/Data/settings_template.cfg
+++ b/Projects/CoX/Data/settings_template.cfg
@@ -12,13 +12,14 @@
 #      for example: 10.0.0.2
 #
 #    Default ports are listed below:
-#      AccountDatabase db_port:		5432
-#      CharacterDatabase db_port:	5432
-#      AuthServer location_addr:	2106
-#      GameServer listen_addr:		7002
-#      GameServer location_addr:	7002
-#      MapServer listen_addr:		7003
-#      MapServer location_addr:		7003
+#      AccountDatabase      db_port:            5432
+#      CharacterDatabase    db_port:            5432
+#      AuthServer           location_addr:      2106
+#      GameServer           listen_addr:        7002
+#      GameServer           location_addr:      7002
+#      MapServer            listen_addr:        7003
+#      MapServer            location_addr:      7003
+#      AdminRPC             location_addr:      6001
 #
 ##############################################################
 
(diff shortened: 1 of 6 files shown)


Projects/CoX/Servers/AuthServer/AdminRPC.cpp
```diff
@@ -33,62 +39,81 @@ bool AdminRPC::ReadConfig()
 {
     ACE_Guard<ACE_Thread_Mutex> guard(m_mutex);
 
-    qInfo() << "Loading AdminRPC settings...";
+    qDebug() << "[ADMINRPC]: Start loading settings...";
```
dracc: Why not `qInfo()`?

Projects/CoX/Servers/AuthServer/AdminRPC.cpp
```diff
@@ -33,62 +39,81 @@ bool AdminRPC::ReadConfig()
...
     QString location_addr = config.value(QStringLiteral("location_addr"),"127.0.0.1:6001").toString();
+    QString server_type = config.value(QStringLiteral("server_type"),"tcp").toString();
+    if(server_type == "tcp")
+    {
+        m_socket_type     = SocketType::tcp;
+    }
+    else
+    {
+        m_socket_type     = SocketType::websocket;
+    }
```
dracc: nit: I'd personally prefer if these were swapped around so TCP is the default if a misspelling has occurred.

Projects/CoX/Data/settings_template.cfg
```diff
@@ -39,6 +40,27 @@ CharacterDatabase\db_pass   = segs123
...
+#
+#    Uncommenting the following section and setting the
+#    parameters will allow the RPC server to listen on an
+#    alternative interface IP address and port. The server type
+#    can be changed as well. 
+# 
+#  location_addr  Can be any valid IP and port combination
+#                 for the server.
+#
+#  server_type    Can be either 'websockets' or 'tcp'.
```
dracc: Called `websocket` everywhere else.

Projects/CoX/Servers/AuthServer/AdminRPC.h
```diff
@@ -15,13 +15,19 @@
 #include <QVariant>
 #include <QtCore/QDateTime>
 
+enum class SocketType {
```
broxen: For code consistency please put brackets on their own line

Projects/CoX/Servers/AuthServer/AdminRPC.cpp
broxen: If we do elect to use qInfo here, we can revert it to the original text to match the format of other Info messages we output to the server.

Projects/CoX/Servers/AuthServer/AdminRPC.cpp
```diff
@@ -33,62 +37,82 @@ bool AdminRPC::ReadConfig()
...
+        m_socket_type     = SocketType::websocket;
+    else
+        m_socket_type     = SocketType::tcp;
+
     config.endGroup(); // AdminRPC
 
     if(!parseAddress(location_addr,m_location))
     {
-        qCritical() << "Badly formed IP address: " << location_addr;
+        qCCritical(logRPC) << "Badly formed IP address: " << location_addr;
```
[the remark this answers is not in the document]
whitegryphon: Since using `qCritical()` makes sense for the reason you point out, should critical messages havce some indication as to what called it? In this specific use about the badly formed IP, this exact wording is used elsewhere, so tracking down in where it failed could be more difficult without the category. That was the original intention of prepending `[AdminRPC]` to the messages. Or should I change it to something like `qCritical() << "Badly formed IP address: " << location_addr << " in AdminRPC";`?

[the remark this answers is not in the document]
broxen: Makes sense. Purely for consistency, let's do it the second way.

Issue #706, opened 2018-11-22
[Meta] How Can I Help? aka Future Server Features, and Roadmap to v0.7


broxen:
## Summary
We need to implement these items one by one. I'm marking this `good-first-issue` as although these aren't all simple tasks, this can become a great resource for new members of all skill levels looking for how to help in meaningful ways.

**This also provides a _loose_ Roadmap for v0.7, but note that some features may be pushed back depending upon time and level of completion.**

When you take on one of the tasks (or groups) below, please [create a new issue](https://github.com/Segs/Segs/issues/new) to track your progress. Thank you!

> See FloatingDamage for reference on how to implement server->client pkt communication: [Segs/Projects/CoX/Servers/MapServer/Events/FloatingDamage.h](https://github.com/Segs/Segs/blob/develop/Projects/CoX/Common/Messages/Map/FloatingDamage.h)

## Check the Issue Queue
- [Issues labeled `bug`](https://github.com/Segs/Segs/labels/bug)
- [Issues labeled `good first issue`](https://github.com/Segs/Segs/labels/good%20first%20issue)
- [Issues labeled `enhancement`](https://github.com/Segs/Segs/labels/enhancement)
- [Issues labeled `wishlist`](https://github.com/Segs/Segs/labels/wishlist)

### Low Difficulty
- [ ] Force Client Logout (boot client) (packet 12) -- In progress #700 
- [ ] Client Console Output (packet 17)
- [ ] Client Console PrintF (packet 19)
- [ ] Send VisitLocations (not sure of purpose, maybe early badges?) (pkt 63)
- [ ] Correlate Map Indexes with Original Servers -- In Review: see #584
- [ ] Update "Current Features" list on Wiki -- see #554
- [ ] Replace `std::unique_ptr<T>(newT())` with `std::make_unique()` calls -- see #532 
- [ ] Normalize basic type arguments -- see #357 
- [ ] Update server architecture on wiki -- see #333 
- [ ] Update doxygen comments throughout the code base -- see #360 #699 
- [ ] Implement `getEntitiesInRadius()` and `getEntitiesInCone()` helper methods -- see #686 

#### Misc: 
- [ ] Map Cells Fog of War (packet 22) - Partially implemented #371 
- [ ] Control Player (pkt 38)
- [ ] Bug Report (pkt 80) -- Track in: #287 #35 
- [ ] Initiate Respec - using `/start_menu 13` somehow? Otherwise spoof with slash command and training?
[...]

PR #750, opened 2019-02-02
Lua Hide and Seek mission

### LINKED ISSUE #726: Scripted Mission: Hide-n-Seek
## Request
Can we script a game of hide n seek? Basic gameplay would follow:
- Script spawns NPC (the hider) in predefined location (under Atlas for example)
- Talking to NPC would provide dialog to initiate game
- Once game has been started, a countdown from 5 to 1 would begin, with the NPC using the broadcast channel to announce the game beginning.
- at `0` or `start` or whatever, the NPC would randomly teleport to a place on the map.
- Maybe NPC `/b`roadcasts hints periodically? Neighborhood or rough direction (east/west/etc) -- implemented by referencing map beacons, or maybe using location to extrapolate position relative to NPC starting location?
- First player to interact with the NPC receives a point and resets the game.
- Extra cool if we can use the `/kiosks` to track player scores for that day.

### Challenges
Challenges will be spawning in a safe location and tracking score. These might be overcome by referencing movement waypoints for spawn points, and by leveraging the souvenir system for tracking score

ZopharShinta:
## Summary
Adds a small hide and seek mission.
There are 10 hiding places (more can be added easily) in Atlas park.
Any player in Atlas Park can start the mission.
NPC will broadcast a clue once at the hiding spot.
Every minute the NPC will say (Did I hear someone?) in "local" chat to help a player nearby.
After 15 minutes the NPC will broadcast the clue again.
After 30 minutes the NPC returns to the starting point near the steps of Atlas Plaza.
The first player to find the NPC and talk to her will get a point.
Points can be checked at Kiosks or by entering /kiosk in the chat

### Issues closed
#726 
Starts the Kiosk in #706


Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2674,13 +2687,13 @@ void MapInstance::on_awaiting_dead_no_gurney(AwaitingDeadNoGurney *ev)
 
     auto val = m_scripting_interface->callFuncWithClientContext(&session,"revive_ok", session.m_ent->m_idx);
 
-    if (val.empty())
+    /*if (val.empty())
```
broxen: Doesn't removing this check result in the script and the following cpp running every time a player resurrects? Why remove it?

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -1337,7 +1341,21 @@ void cmdHandler_SendConsolePrint(const QString &cmd, MapClientSession &sess)
     sendClientConsoleOutput(sess, msg);
 }
 
+void cmdHandler_ClearTarget(const QString &cmd, MapClientSession &sess)
```
broxen: Did we get this working? I know we had some questions about it.

Projects/CoX/Servers/MapServer/MapInstance.h
```diff
@@ -27,6 +27,16 @@ class SEGSTimer;
 class World;
 class GameDataStore;
 class MapSceneGraph;
+
+struct LuaTimer {
```
broxen: Please put brackets on their own line.

Projects/CoX/Servers/MapServer/SlashCommand.cpp
ZopharShinta: No it doesn't work. The client overwrites the change before the server sends the update to the client.

Projects/CoX/Servers/MapServer/MapInstance.cpp
ZopharShinta: This was an error on my part.
I missed removing the comment when I restored the return values.

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -625,14 +833,14 @@ std::string ScriptingEngine::callFunc(const char *name, int arg1)
     if(!funcwrap.valid())
     {
         qCritical() << "Failed to retrieve script func:" << name;
-        return "";
+         return "";
```
broxen: Nit: there are several of these extra spaces in this file

Projects/CoX/Data/scripts/global.lua
```diff
@@ -1,5 +1,7 @@
 printDebug('Loading global.lua...')
 
+
+
```
broxen: We don't need all these extra line returns

Components/TimeHelpers.cpp
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+#include "TimeHelpers.h"
+#include <QDateTime>
+
+QDateTime getPostgreEpoch()
+{
+    return QDateTime(QDate(2000,1,1));
+}
+
```
nemerle: We need to check if `getPostgreEpoch` is used anywhere, from what I've gathered, it's not :)

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -3077,5 +3117,49 @@ void MapInstance::on_store_buy_item(StoreBuyItem* ev)
...
+            break;
+
+        ++count;
+    }
+    this->m_lua_timers[count].m_is_enabled = false;
+}
+
+void MapInstance::clearTimer(uint32_t entity_idx)
+{
+    int count = 0;
```
nemerle: This function will fail strangely if entity_idx is incorrect.

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -9,6 +9,7 @@
  * @addtogroup MapServer Projects/CoX/Servers/MapServer
  * @{
  */
+#pragma once
 #include "ScriptingEngine.h"
```
nemerle: No need for `pragma once` in a cpp file.

Projects/CoX/Servers/MapServer/ScriptingEngine.cpp
```diff
@@ -221,6 +232,88 @@ void ScriptingEngine::registerTypes()
...
+    };
+
+    m_private->m_lua["MapInstance"]["ClearTimer"] = [this](int entityIdx)
+    {
+        MapInstance *mi = m_private->m_lua["map"];
+        Entity* e = getEntity(mi, entityIdx);
+        if(e != nullptr)
+        {
+          mi->clearTimer(entityIdx);
+        }
```
nemerle: Indentation is 4-space :)

Components/TimeHelpers.cpp
broxen: This is my bad precedent. I added this to TimeHelpers thinking it would have more utility and both eoS and Zophar followed my lead there.

Turns out pretty much every time calculation factors the msecs since this time, which we do in the method below, so this method is useless.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1514,4 +1576,66 @@ void sendClientConsoleOutput(MapClientSession &cl, QString &message)
     cl.addCommand<ConsolePrint>(message);
 }
 
+void npcSendMessage(MapClientSession &cl, QString& channel, int entityIdx, QString& message)
```
Magentah: As mentioned in discord - I don't think this should be creating a MapClientSession as it's entirely for NPCs, and MapClientSession is part of the client authentication flow.

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -3077,5 +3117,63 @@ void MapInstance::on_store_buy_item(StoreBuyItem* ev)
         qCDebug(logStores) << "Error processing buyItem";
 }
 
+void MapInstance::add_chat_message(MapClientSession *sender,QString &msg_text)
```
Magentah: This shouldn't use the MapClientSession after 'npcSendMessage()` is changed to not use MapClientSessions.

Projects/CoX/Servers/MapServer/MessageHelpers.cpp
```diff
@@ -35,6 +35,23 @@ void sendChatMessage(MessageChannel t, const QString &msg, MapClientSession *src
 
     tgt.addCommandToSendNextUpdate(std::move(res));
 }
+
+
+void sendNpcChatMessage(MessageChannel t, const QString &msg, int npc_idx, MapClientSession &tgt)
```
Magentah: This is functionally the same as `sendChatMessage()`, so can combine them.

Projects/CoX/Servers/MapServer/SlashCommand.cpp
broxen: For now I don't see any harm in leaving this framework in place then, but I would add a comment to this method describing that the behavior isn't functioning as expected based upon some odd behavior from the client, and I'd also spit out a qWarning() to the same effect, so that if it's used accidentally the server op knows about it.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
broxen: I agree NPCs should not have a MapClientSession, but with the current structure the only way to send a message to the clients that need to see it is through the use of the Session object, for example `tgt` here `tgt.addCommandToSendNextUpdate(std::unique_ptr<ChatMessage>(res));`

crylico's ChatService should fix this issue, as it will handle that routing and use the recipients own Session to send the packet.

Until then, I'm not sure what our options are?

Magentah: Sorry, I meant just the Source shouldn't rely on a MapClientSession. The target using MapClientSession is fine in this case I think, and will work until the ChatService is implemented.

broxen: Ah, makes perfect sense.

PR #759, opened 2019-02-07
AdminRPC: change getVersion output and add getCodename

valcryst:
Removed the versions name from AdminRPC::getVersion and added  
the command AdminRPC::getCodename wich returns the versions name

## Summary
From a data processing point of view it is better if only one data type  
per field is used in the result output. (str, int, ..)  

inspired by:
```
 ⌁ root  ~   lsb_release -r
Release:        9.6
 ⌁ root  ~   lsb_release -c
Codename:       stretch
```

### DIFF
--- a/Projects/CoX/Servers/AuthServer/AdminRPC.cpp
+++ b/Projects/CoX/Servers/AuthServer/AdminRPC.cpp
@@ -87,11 +87,11 @@ QString AdminRPC::getVersion()
     return version;
 }
 
-QString AdminRPC::getCodename()
+QString AdminRPC::getVersionName()
 {
     qCDebug(logRPC) << "Someone requested the versions name...";
-    QString codename = VersionInfo::getVersionName();
-    return codename;
+    QString versionname = VersionInfo::getVersionName();
+    return versionname;
 }
 
 QString AdminRPC::ping()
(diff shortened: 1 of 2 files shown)


Projects/CoX/Servers/AuthServer/AdminRPC.cpp
```diff
@@ -83,10 +83,17 @@ QString AdminRPC::helloServer()
 QString AdminRPC::getVersion()
 {
     qCDebug(logRPC) << "Someone requested the version...";
-    QString version = VersionInfo::getAuthVersionNumber() + QString(" ") + VersionInfo::getVersionName();
+    QString version = VersionInfo::getAuthVersionNumber();
     return version;
 }
 
+QString AdminRPC::getCodename()
```
broxen: To follow the nomenclature used on the server and keep consistency, I'd prefer we call this `getVersionName()` instead.

Projects/CoX/Servers/AuthServer/AdminRPC.cpp
```diff
@@ -83,10 +83,17 @@ QString AdminRPC::helloServer()
 QString AdminRPC::getVersion()
 {
     qCDebug(logRPC) << "Someone requested the version...";
-    QString version = VersionInfo::getAuthVersionNumber() + QString(" ") + VersionInfo::getVersionName();
+    QString version = VersionInfo::getAuthVersionNumber();
```
dracc: No need to assign the variable, just do `return VersionInfo::getAuthVersionNumber()` instead.

PR #763, opened 2019-02-08
Email regression fix

### LINKED ISSUE #762: EmailHandler regression causing crash!
### Summary
dracc reported a crash in EmailHandler that appears to be a regression from #637. 
```
Segs/Projects/CoX/Servers/GameServer/EmailService/EmailHandler.cpp:52: virtual void EmailHandler::dispatch(SEGSEvents::Event*): Assertion `false' failed.
```
It looks as though removing emailHeaderRequest has caused the dispatch switch to fall through to the default condition.
```
- case EmailEventTypes::evEmailHeaderRequest: 
-     on_email_header(static_cast<EmailHeaderRequest *>(ev)); break;
```

### Proposed Solution
Looks like we need to do two things. First it seems like the intention was to replace EmailHeaderRequest with another event, and that isn't working, so we need to get to the bottom of that. Secondly, we really should provide better error handling there by changing the `assert(false);` to a more verbose error message. qFatal() if failing to dispatch is truly server halting, but more likely qWarning(), with a message about why it failed and what the handler was trying to do (which `type`).

I'm assigning eoS, as he's most familiar with the code base, and the bug appears to have originated in his PR #637

eos428:
Fixes #762


broxen:
Tagging @Segs/reviewers for Code Review. Can someone please build and test to ensure this resolves the issue #762? There are instructions here https://help.github.com/articles/checking-out-pull-requests-locally/

Magentah:
This segfaults here.

`
void EmailHandler::on_email_header(EmailHeaderRequest *msg)
{
    std::vector<EmailHeaderData> email_headers;
    int unread_emails_count = 0;
    fill_email_headers(email_headers, msg->m_data.m_user_id, unread_emails_count);

   msg->src()->putq(new EmailHeaderResponse({email_headers}, msg->session_token())); // This line segfaults.
}
`

eos428:
> 
> 
> This segfaults here.
> 
> ```c++
> void EmailHandler::on_email_header(EmailHeaderRequest *msg)
> {
>     std::vector<EmailHeaderData> email_headers;
>     int unread_emails_count = 0;
>     fill_email_headers(email_headers, msg->m_data.m_user_id, unread_emails_count);
> 
>    msg->src()->putq(new EmailHeaderResponse({email_headers}, msg->session_token())); // This line segfaults.
> }
> ```

Seems like this happens because ``src()`` would return null because I didn't set the MapInstance as ``ev_src`` param when sending the event. Oh well, this probably should be fixed by just finding the relevant MapInstance from EmailHandler's stored client datas instead.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1271,7 +1273,9 @@ void showMapMenu(MapClientSession &sess)
 int64_t getSecsSince2000Epoch()
 {
     QDateTime base_date(QDate(2000,1,1));
-    return base_date.secsTo(QDateTime::currentDateTime());
+
+    // 25200 -> seven hours in secs
+    return base_date.secsTo(QDateTime::currentDateTimeUtc()) - 25200;
```
Magentah: Is there any particular reason we're removing 7 hours from utc..?

broxen: I think I saw a discussion about this in Discord, but we should resolve the underlying issue here, instead of masking it with magic numbers.

Projects/CoX/Servers/GameServer/EmailService/EmailHandler.cpp
```diff
@@ -111,7 +116,12 @@ void EmailHandler::on_email_header(EmailHeaderRequest *msg)
     int unread_emails_count = 0;
     fill_email_headers(email_headers, msg->m_data.m_user_id, unread_emails_count);
 
-    msg->src()->putq(new EmailHeaderResponse({email_headers}, msg->session_token()));
+    const ClientSessionData &session (m_state.m_stored_client_datas[msg->m_data.m_user_id]);
+    EventProcessor *session_map_instance = HandlerLocator::getMapInstance_Handler(
+                session.m_server_id,
+                session.m_instance_id);
+
+    session_map_instance->putq(new EmailHeaderResponse({email_headers}, msg->session_token()));
```
broxen: Per Discord, to avoid duplicates, let's ensure that we're setting `m_fullupdate` whenever we want to reset the Emails array client side.

broxen:
Thank you eoS. If any @Segs/testers cam pull this down to test, I'd appreciate the help.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
broxen: We should be leveraging `Components/TimeHelpers.h` here, and keeping this as `currentDateTime()`.

Why the switch to UTC? Especially given that you then need to artificially modify that time value.

Components/TimeHelpers.cpp
```diff
@@ -15,7 +15,7 @@
 
 int64_t getSecsSince2000Epoch()
 {
-    QDateTime base_date(QDate(2000,1,1));
+    QDateTime base_date(QDate(2000,1,1),QTime(0,0),Qt::UTC);
     return base_date.secsTo(QDateTime::currentDateTime());
```
nemerle: hmm, shouldn't that be QDateTime::currentDateTimeUTC() ? :smile:

Magentah:
Looks like we're still sending `m_fullupdate` as `true` every time, is that correct?

eos428:
If sent as a vector, yes.
If sent individually, no.

PR #764, opened 2019-02-08
Refactoring MapInstance into multiple Services

### LINKED ISSUE #755: Refactor MapInstance for code hygiene and less merge conflicts
We have talked this before in Discord, and since EmailHandler PR is finally merge I'd like to start doing this.

The idea is to split MapInstance into multiple CPP files categorized as Services (so, like GameDbSyncService I guess), each one dealing with their own specific functions. For example, an EmailService would take all the email-related functions in MapInstance, and etc.

Why?
- Reduce the big file size and line count in MapInstance
- More resistant towards merge conflicts because now people will be editing and/or adding their own files, as opposed to one singular big file

What to do:
- Don't need to do anything with MapInstance.h
- Split up MapInstance.cpp into multiple Services, all of which would include MapInstance.h
- Create a new folder to store all the Services (oh, dealing with all the Cmake stuff again...)
- Maybe create a ``Service`` class that derives from EventProcessor, but we'll see if there's some kind of function that all these Services must implement that is not already in EventProcessor

eos428:
Closes #755 

Putting up EmailService first because it's the most familiar one to me and I should ask for opinions on whether this is sound enough structurally or not.

And I think we can probably do with an entire folder containing only the Services. It currently compiles, but of course we know just compiling is never enough :)


broxen:
@crylico I'd love to get your input here on the Services, and how they may interact with your work over on TeamService and UserRouter.

crylico:
I think this goes back to a convo we had in discord. I'd love to set up MapInstance to have 'sub-files' for each of the functionalities. I'm happy to see what @eos428 comes up with, then rename some of my stuff to fit :)

crylico:
oh wow, didn't even look at the commit. I like this a lot. that way the map instance "owns" the service

crylico:
If we could merge this into develop, I'll be able to merge it into my branch so that when it's ready for review, I can build off this

Magentah:
Small thing, but were we not trying to move away from using `new Class()` to using smart pointers?

crylico:
@Magentah I'm not sure what a smart pointer is...willing to learn for future changes :D

Magentah:
`std::make_unique<T>(T)`, `std::make_shared<T>(T)`, etc

Magentah:
You can take a look at PR #749 to see how it changes things

Magentah:
Why are MapInstance:: functions now in EmailService.cpp? I don't think just moving the functions from MapInstance to different files is the intention, and makes things more confusing later.

eos428:
The intent is to emulate partial classes in C#. So to be fair, I am also not sure if I should name the new classes as Services. I prefer writing them as MapInstance_Emails.cpp and MapInstance_Settings.cpp etc

Magentah:
I don't think partial classes are the correct solution to this personally. I think just having service classes for each distinct area makes the most sense. They don't necessarily need to be event processors, but EmailService, SupergroupService, etc seems to be make the most sense

eos428:
Tonight (or soon TM) I will probably make this closer to the initial implementation, minus the EventProcessor thing, unless they are sending events of their own. At the very least for emails, MapInstance didn't have any email events that would send events, since those are done via SlashCommand functions.

eos428:
Latest commit reworks the implementation to something like the first commit (EmailService with its own functions) but minus the EventProcessor for EmailService. As before, this is a test commit for others to see and if everyone agrees with this approach, I'll replicate this implementation for other services as well.

There's a bit of gaffe due to reverting commit so I hope the merged conflicts are alright.

Some talking points (I discussed with Nem about this some time ago, but I think it's still worth mulling over the options again):
- If in the end MapInstance is still in charge of sending stuff to the client sessions, that means there would be less corners cut from it. And I think in the end it wouldn't make much difference to reducing the line count in MapInstance in that way.
- What exactly should be stored in the 'states' of the services that would be reused in a later time? Granted, I'm just doing emails atm so I don't see the use for a state in EmailService, but others might want it I suppose.

In the end, there would be a lot of made of Services made (some of them may need to be EventProcessors), and we'd need to review it on a case-by-case basis. I'll try my best to get this and SlashCommands refactor out of the way for v0.7.

Magentah:
Separating out things from MapInstance to services isn't just to reduce the line count, but also to separate things out for better organisation. At the moment, MapInstance has a lot of things that it does, when ideally it should only really be doing a couple of things at most (receiving and sending client events on the map, maybe some other things).  The line count should decrease a fair bit anyway if everything ends up in services, as all the functions would end up with just a few lines, or sometimes just a single line.

Service states would be on a per-service basis, with the ideal being stateless I imagine. We probably don't want to store the state of things if we can help it, but if it makes things a lot easier / faster to store things, then we should.

That's just my take on it though.

Projects/CoX/Servers/GameServer/EmailService/EmailService.cpp
```diff
@@ -0,0 +1,88 @@
...
+    std::vector<EmailHeaders::EmailHeader> email_headers;
+    for (const auto &data : ev->m_data.m_email_headers)
+    {
+        email_headers.push_back(EmailHeaders::EmailHeader{data.m_email_id,
+                                                          data.m_sender_name,
+                                                          data.m_subject,
+                                                          data.m_timestamp});
+    }
+
+    map_session.addCommandToSendNextUpdate(std::make_unique<EmailHeaders>(email_headers));
```
Magentah: I thought that the idea was to return the data to the MapInstance, who would than handle sending events. So EmailService would just return email_headers here, and MapInstance would change to

```cpp
// Something like this..
// In EmailService
EmailHeaders EmailService::on_email_header_response(EmailHeaderResponseData* data)
{
    auto email_headers;
    // populate from data
    return email_headers;
}

// In MapInstance
auto& response = email_service->on_email_header_response(static_cast<EmailHeaderResponse *>(ev->m_data));
map_session.addCommandToSendNextUpdate(response);
```
The idea being that the services don't care about events or anything.

broxen: This makes a lot of sense. Let's just keep in mind that there are edge cases where it might need to return multiple commands in a response. An example would be in `on_email_headers_to_client()`

Magentah: Can just return the list and iterate through, or add a new function to map session that takes a list/array of commands to add.

Magentah: I can see how this may look a bit redundant for simple events like there are in the EmailService, where it's literally just taking data from one event and sending it to another event but it's just adding additional steps. It makes more sense for more involved functions that do a lot before sending a response event. It does however handle the whole separation of concern thing that I think is a main goal of the refactor.

nemerle: I like a hybrid approach here actually :)
Since EmailService is a thing that is a service but not an 'EventProcessor' one ( stand-alone ), we should likely treat those things like this:
Mapping things from EventProcessor approach.

| Activated EP | Non-EP |
| --- | --- |
| Receive an event| handler function directly called |
| send a 0-n responses| return a std::vector<std::unique_ptr<Event>> |

note: In cases where we want to send messages to different EP's we will need the 'holder' EP of the Non-EP service be available ( likely passed as a Non-EP service's constructor parameter ). 
Actually, this might be a good place to introduce NonActiveService interface class?
```cpp
class EventHandler {
    EventProcessor *m_holder;
    EventHandler(EventProcessor *ep) : m_holder(ep) {}
    virtual std::vector<std::unique_ptr<SEGSEvent>> dispatch_sync(SEGSEvent *) = 0;
};
```

The issue we might encounter with this is sending messages to different sessions in the target EP, but that could maybe solved with a bit of re-architecting ?

```cpp
class EventProcessor {
////////////////
virtual void sendToSession(int id,std::unique_ptr<SEGSEvent>) { /* by default we do nothing*/}
////////////////
}
```

eos428: A major bottleneck with Magenta's solution - SEGSEvent::Event (without being casted) cannot give the session token and cannot give m_data unless casted to another event.

So, the Service would need to extract the token (easy enough) and then the content of the event. And well, the more tricky part is having to make the classes to contain both of the things returned.

I already started with SettingsService, which contains only keybind stuffs for now. They use session_from_event, so it's a bit different to EmailService. The big difference is that those keybind functions are located inside the entities.

Magentah: I'm not sure how that is a problem? The service doesn't need a token or a session (unless it's an eventprocessor maybe, which this one isn't). Anything related to tokens or sessions should probably be handled in MapInstance or relevant eventprocessor. This service just needs the m_data from the event and you're already casting the event when you pass it to the service in your current solution.

eos428: The service does not need the token, but MapInstance certainly does to find the needed session.

So, we'd probably need to adjust things a bit - cast the event before we'd pass it to the Service's function because MapInstance needs the token from the casted event.

Magentah: I would assume the MapInstance would continue to handle the sessions/tokens the same way it does now. It needs them to be able to send responses to the client. Just do something like:

```cpp
// kinda pseudo, probably won't compile.
case evEmailHeaderResponse:
    {
        auto ev = std::make_unique<EmailHeaderResponse>(ev);
        MapClientSession &map_session(m_session_store.session_from_token(ev->session_token()));
        auto &response = m_email_service->on_email_header_response(ev->data);
        map_session.addCommandToSendNextUpdate(response);
        break;
    }
    // etc
```

eos428: Yeah, probably something like this.

AddCommandToSendNextUpdate probably would need a new pointer again afaik since there's some rvalue-lvalue issue.

eos428: Hmm... so far it is okay for email_header_response and similar, where I'd just have to return the EmailHeaders (or a GameCommandEvent, if you want to be more direct) but as broxen also raised, I'm a bit lost on stuff when I'd have to do more than one thing, like on_email_headers_to_client.

This is probably where I have to create a Service class that contains the EmailHeaders and then the message to send to the client? But this might confuse people even more since I'd have to make something like `EmailHeadersToClientServiceData` in which the message is also obtainable from `EmailHeaderResponseData` in the first place, although I'd need to add a little bit more to the message inside it.

On the WIP SettingsService (haven't pushed yet), something like `on_select_keybind_profile` literally just modifies the entity with something directly inside the event, and would simply cause redundancy.

So overall, if an event is responsible for let's say sending a debug message to the client, I don't feel we have to use a Service to relay that thing back to MapInstance, since at times it's literally there inside the event itself. Redundancy would just add unnecessary steps and confuse people further. What we should do for such cases then?

Magentah: This seems to mostly be a decision between redundancy and separation of concerns. My preference is to generally lean towards separation, but probably best to see what @nemerle and @broxen think.

Returning multiple things, such as for `on_email_headers_to_client` should be fairly simple - just return a vector, and add a new function to `MapSession` that can take a vector parameter for commands. 
```cpp
MapSession::addCommandtoSendNextUpdate(std::vector<T> commands)
{
    for (auto &command : commands)
    {
        addCommandToSendNextUpdate(command);
    }
}`

Comment (eos428) [reply Projects/CoX/Servers/GameServer/EmailService/EmailService.cpp]: It appears `EmailHeadersToClient` in EmailService hasn't used the new implementation done in `EmailHeaderResponse` (one command containing a vector of emails instead of multiple commands each containing one email. which is important because m_fullupdate is true with the new solution). My bad for not updating it.

My issue is with the message string since you can extract it immediately from the event data and then you just have to add a little bit to it. Note that sometimes the entire message is inside the event already, and sometimes you just have to do QString.arg or such, but nothing complicated.

This is where the redundancy issue kicks in - making a Service function that relays exactly the same message in the event data we already cast in MapInstance is a waste of time imo.

File: Projects/CoX/Servers/GameServer/CMakeLists.txt [hunk 2]
```diff
@@ -27,8 +27,9 @@ target_link_libraries(GameServer INTERFACE SEGS_Components gameData )
 
 add_subdirectory(FriendshipService)
 add_subdirectory(EmailService)
+add_subdirectory(ClientOptionService)
 
-target_link_libraries(GameServer PUBLIC Qt5::Core FriendshipService EmailService )
+target_link_libraries(GameServer PUBLIC Qt5::Core FriendshipService EmailService)
```
broxen: You need to add ClientOptionService here and it should solve your linking error.

eos428: wtf I had been getting cannot find interface error doing that and today it suddenly worked??? Anyways we finally have the linking error done. I think it's a good time to do preliminary tests and see if what we have done so far is actually working or not.

eos428:
Latest commit splits ServiceToClientData to two based on how the session is supposed to be found. Internal Service will find based on token, while Map Service will find based on entity. The latter should have less margin for error, but I still added the try/catch block for its response function in MapInstance, just in case.

So a major issue encountered is that the new services uses functions in DataHelpers and such. These functions in DataHelpers are also responsible for sending commands to the client. Since we want to make MapInstance the only one responsible for doing that, these functions should just give the commands to send in the Service, which will return it back to the MapInstance.

However, dealing with the functions in DataHelpers may take a while. I'd like to do that in a separate PR, once this one is over with if that's okay with you people.

Projects/CoX/Servers/GameServer/EmailService/EmailService.cpp
```diff
@@ -16,69 +16,77 @@
...
-InternalServiceToClientData* EmailService::on_email_header_response(Event* ev)
+ServiceToClientData* EmailService::on_email_header_response(Event* ev)
 {
-    EmailHeaderResponse* resp = static_cast<EmailHeaderResponse *>(ev);
+    EmailHeaderResponse* casted_ev = static_cast<EmailHeaderResponse *>(ev);
 
-    std::vector<EmailHeaders::EmailHeader> email_headers;
-    for (const auto &email_header : resp->m_data.m_email_headers)
+    std::vector<EmailHeaders::EmailHeader> emailHeadersVector;
+    for (const auto &email_header : casted_ev->m_data.m_email_headers)
```
nemerle: since `casted_ev->m_data.m_email_headers` has known size, you can `reserve` the emailHeadersVector to have correct size ( in case of many headers it will reduce the re-allocations caused by growing the emailHeadersVector vector)

eos428:
Expanding on the commit message:

The implementation for TransactionService and ZoneTransferService are crude at the moment (especially the latter). For the former I guess we can just let it have a pointer to the owner MapInstance, but for ZoneTransferService, I'd need to:
- Refactor the Helper classes too so that they'd return the GameCommands instead of sending those to the sessions by themselves - which I'd hold on doing until this PR is complete.
- Maybe do link->putq() in `on_service_to_client_response` instead?

Before the new services are added this PR works fine, at least for email stuff. The most dangerous change is probably ZoneTransferService, because it does a lot of things at once.

Projects/CoX/Common/Servers/InternalEvents.h
```diff
@@ -243,4 +247,96 @@ struct MapSwapCollisionData
...
+    {
+        m_ent = ent;
+        m_commands = std::move(commands);
+        m_message = msg;
+        m_message_channel = messageChannel;
+    }
+
+    // use this if you are sending GameCommands and find your session from token (eg EmailService)
+    ServiceToClientData(uint64_t token, GameCommandVector commands, QString msg, MessageChannel messageChannel = MessageChannel::DEBUG_INFO)
+    {
```
nemerle: std::move from an argument passed as a value does not make much sense ;)
Either pass commands by universal reference `&&`, or just assign it as is

Projects/CoX/Servers/GameServer/CMakeLists.txt
```diff
@@ -27,8 +27,15 @@ target_link_libraries(GameServer INTERFACE SEGS_Components gameData )
...
-target_link_libraries(GameServer PUBLIC Qt5::Core FriendshipService EmailService )
+add_subdirectory(ClientOptionService)
+add_subdirectory(LocationService)
+add_subdirectory(CharacterService)
+add_subdirectory(TransactionService)
+add_subdirectory(ZoneTransferService)
+add_subdirectory(InteractionService)
+
+target_link_libraries(GameServer PUBLIC Qt5::Core FriendshipService EmailService ClientOptionService LocationService)
+target_link_libraries(GameServer PUBLIC Qt5::Core CharacterService TransactionService ZoneTransferService InteractionService)
```
nemerle: You're adding Qt5::Core twice :)

Projects/CoX/Servers/GameServer/CharacterService/PowerService/PowerService.h
```diff
@@ -0,0 +1,31 @@
...
+ * This software is licensed under the terms of the 3-clause BSD License. See LICENSE.md for details.
+ */
+
+#pragma once
+
+#include "Common/Servers/InternalEvents.h"
+#include "GameData/Entity.h"
+
+#include <QVector>
+
```
nemerle: Not a change request: we should provide a writeup for every `Service` that describes the service's responsibilities.

Projects/CoX/Servers/GameServer/ClientOptionService/ClientOptionService.cpp
```diff
@@ -0,0 +1,95 @@
...
+    qCDebug(logMapEvents) << "Saving viewpoint mode to ClientOptions" << casted_ev->new_viewpoint_is_firstperson;
+}
+
+void ClientOptionService::on_window_state(Entity* ent, Event* ev)
+{
+    WindowState* casted_ev = static_cast<WindowState* >(ev);
+    // Save GUISettings to character entity and entry in the database.
+
+    ent->m_player->m_gui.m_wnds.at(casted_ev->wnd.m_idx) = casted_ev->wnd;
+
```
nemerle: `at(casted_ev->wnd.m_idx)` will throw an exception in a case where m_idx is out of bounds, it's likely not something we want. Consider rewriting this to check `idx>0 && idx<wnds.size()` and log/return.

Projects/CoX/Servers/GameServer/EmailService/EmailService.cpp
```diff
@@ -0,0 +1,101 @@
...
+    return new ServiceToClientData(casted_ev->session_token(), std::move(commands), messageToClient);
+}
+
+ServiceToClientData* EmailService::on_email_read_response(Event* ev)
+{
+    EmailReadResponse* casted_ev = static_cast<EmailReadResponse *>(ev);
+    std::unique_ptr<EmailRead> emailRead = std::make_unique<EmailRead>(
+                casted_ev->m_data.m_email_id, casted_ev->m_data.m_message, casted_ev->m_data.m_sender_name);
+    GameCommandVector commands;
+    commands.push_back(std::move(emailRead));
```
nemerle: using `emplace_back` might be useful here?
```
commands.emplace_back(std::make_unique<EmailRead> (.....) ); 
```
consider it ?

Projects/CoX/Servers/GameServer/EmailService/EmailService.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+    return new ServiceToClientData(casted_ev->session_token(), std::move(commands), QString());
+}
+
+ServiceToClientData* EmailService::on_email_read_by_recipient(Event* ev)
+{
+    EmailWasReadByRecipientMessage* casted_ev = static_cast<EmailWasReadByRecipientMessage *>(ev);
+    return new ServiceToClientData(casted_ev->session_token(), {}, casted_ev->m_data.m_message);
```
nemerle: Hmm, we might seriously consider returning unique_ptr<ServiceToClientData> in all cases, to make sure the memory is properly released/managed?

Projects/CoX/Servers/GameServer/InteractionService/InteractionService.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    return new ServiceToClientData(ent, {scriptData}, QString());
+}
+
+ServiceToClientData* InteractionService::on_receive_task_detail_request(Entity* ent, Event *ev)
+{
+    ReceiveTaskDetailRequest* casted_ev = static_cast<ReceiveTaskDetailRequest *>(ev);
+
+    qCDebug(logMapEvents) << "ReceiveTaskDetailRequest Entity: " << ent->m_idx << "wants detail for task " << casted_ev->m_task_idx;
+    QString detail = "Testind Task Detail Request";
```
nemerle: small typo above ( "Testind" -> "Testing" )

Projects/CoX/Servers/GameServer/TransactionService/TransactionService.h
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (c) 2006 - 2019 SEGS Team (see AUTHORS.md)
+ * This software is licensed under the terms of the 3-clause BSD License. See LICENSE.md for details.
+ */
+
+#pragma once
+
+#include "Common/Servers/InternalEvents.h"
+#include "Servers/MapServer/ScriptingEngine.h"
+
+class TransactionService
```
nemerle: I'd rename this to TradingService, since Transaction has a more specific meaning.

Projects/CoX/Servers/MapServer/MapInstance.h
```diff
@@ -142,12 +92,23 @@ class MapInstance final : public EventProcessor
         std::unique_ptr<SEGSTimer> m_afk_update_timer;
         std::unique_ptr<SEGSTimer> m_lua_timer;
         World *                 m_world;
-        GameDBSyncService*      m_sync_service;
         uint32_t                m_owner_id;
         uint32_t                m_instance_id;
         uint32_t                m_index = 1; // what does client expect this to store, and where do we send it?
         uint8_t                 m_game_server_id=255; // 255 is `invalid` id
 
+        std::unique_ptr<GameDBSyncService>      m_sync_service;
```
nemerle: Since all of the service classes are included above, we can simply store them by value here?

Projects/CoX/Servers/GameServer/EmailService/EmailService.cpp
eos428: There was actually an anonymous person that gave the same advice, and emplace_back is used in InteractionService. I forgot to change the old ones haha

Projects/CoX/Servers/MapServer/MapInstance.h
eos428: I think GameDBSyncService should still be a unique_ptr (it's the one different to all of the others), but the rest should be alright as a value

Projects/CoX/Servers/GameServer/InteractionService/InteractionService.cpp
```diff
@@ -0,0 +1,195 @@
...
+    SouvenirDetailRequest* casted_ev = static_cast<SouvenirDetailRequest* >(ev);
+    vSouvenirList sl = ent->m_player->m_souvenirs;
+
+    Souvenir souvenir_detail;
+    bool found = false;
+    for(const Souvenir &s: sl)
+    {
+        if(s.m_idx != (uint32_t)casted_ev->m_souvenir_idx)
+            continue;
+\
```
broxen: stray character

Projects/CoX/Servers/GameServer/InteractionService/InteractionService.cpp
```diff
@@ -0,0 +1,195 @@
...
+    commands.emplace_back(std::make_unique<EntityInfoResponse>(description));
+
+    qCDebug(logDescription) << "Entity info requested" << casted_ev->entity_idx << description;
+    return std::make_unique<ServiceToClientData>(ent, std::move(commands), QString());
+}
+
+std::unique_ptr<ServiceToClientData> InteractionService::on_interact_with(Entity* ent, Event *ev)
+{
+    InteractWithEntity* casted_ev = static_cast<InteractWithEntity *>(ev);
+    // Entity *entity = getEntity(ent->m_client, casted_ev->m_srv_idx);
```
broxen: We can remove this commented code, as we're not using it anywhere and it provides no useful context to this code block.

Projects/CoX/Servers/GameServer/ClientOptionService/ClientOptionService.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+    ChatDividerMoved* casted_ev = static_cast<ChatDividerMoved* >(ev);
+
+    ent->m_player->m_gui.m_chat_divider_pos = casted_ev->m_position;
+    qCDebug(logMapEvents) << "Chat divider moved to " << casted_ev->m_position << " for player" << ent->name();
+}
+
+void ClientOptionService::on_browser_close(Entity* ent, Event */*ev*/)
+{
+    //BrowserClose* casted_ev = static_cast<BrowserClose *>(ev);
```
broxen: Why comment this out? We'll need to handle this event for closing the Browser Window.

Projects/CoX/Servers/GameServer/ClientOptionService/ClientOptionService.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    ent->m_player->m_keybinds.resetKeybinds(default_profiles);
+    //qCDebug(logMapEvents) << "Resetting Keybinds to defaults.";
+}
+
+void ClientOptionService::on_select_keybind_profile(Entity* ent, Event *ev)
+{
+    SelectKeybindProfile* casted_ev = static_cast<SelectKeybindProfile *>(ev);
+    ent->m_player->m_keybinds.setKeybindProfile(casted_ev->profile);
+    //qCDebug(logMapEvents) << "Saving currently selected Keybind Profile. Profile name: " << ev->profile;
```
broxen: Instead of commenting these out, we should simply change their logging category to logKeybinds and toggle that category off.

Projects/CoX/Servers/GameServer/ChatService/ChatService.cpp
```diff
@@ -0,0 +1,815 @@
...
+#include "GameData/EntityHelpers.h"
+#include "Messages/Map/MapEvents.h"
+#include "Servers/MapServer/DataHelpers.h"
+
+#include <QtCore/QDebug>
+#include <QRegularExpression>
+
+using namespace SEGSEvents;
+
+std::unique_ptr<ChatServiceToClientData> ChatService::on_emote_command(Entity *ent, const QString &command)
```
broxen: Seems like emotes could be their own thing, eh?

Best for a follow up PR.

Projects/CoX/Servers/GameServer/CharacterService/PowerService/PowerService.cpp
```diff
@@ -0,0 +1,131 @@
...
+    ptray->m_default_pset_idx = casted_ev->powerset_idx;
+    ptray->m_default_pow_idx = casted_ev->power_idx;
+
+    qCDebug(logMapEvents) << "Set Default Power:" << casted_ev->powerset_idx << casted_ev->power_idx;
+}
+
+void PowerService::on_unset_default_power(Entity* ent, Event* /*ev*/)
+{
+    // unused event
+    // UnsetDefaultPower* casted_ev = static_cast<UnsetDefaultPower *>(ev);
```
broxen: We should uncomment this?

Projects/CoX/Servers/GameServer/CharacterService/PowerService/PowerService.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "Messages/Map/MapEvents.h"
+#include "Servers/MapServer/DataHelpers.h"
+#include <QtCore/QDebug>
+
+using namespace SEGSEvents;
+
+void PowerService::on_abort_queued_power(Entity* ent, Event* /*ev*/)
+{
+    // unused event
+    // AbortQueuedPower* casted_ev = static_cast<AbortQueuedPower*>(ev);
```
broxen: I'm not understanding why these are commented out. Is this in preparation for Cyanpill's PR or something?

Projects/CoX/Servers/GameServer/LocationService/LocationService.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <QtCore/QDebug>
+
+using namespace SEGSEvents;
+
+std::unique_ptr<ServiceToClientData> LocationService::on_location_visited(Entity* ent, Event *ev)
+{
+    LocationVisited* casted_ev = static_cast<LocationVisited *>(ev);
+    qCDebug(logMapEvents()) << "Attempting a call to script location_visited with:"<<casted_ev->m_name<<qHash(casted_ev->m_name);
+
+    //auto val = m_scripting_interface->callFuncWithClientContext(&session,"location_visited", qPrintable(casted_ev->m_name), casted_ev->m_pos);
```
broxen: we can remove this old code.

Projects/CoX/Servers/GameServer/LocationService/LocationService.cpp
```diff
@@ -0,0 +1,74 @@
...
+    ScriptVector scripts {scriptData};
+    return std::make_unique<ServiceToClientData>(ent, scripts, casted_ev->m_name);
+}
+
+std::unique_ptr<ServiceToClientData> LocationService::on_plaque_visited(Entity* ent, Event* ev)
+{
+    PlaqueVisited* casted_ev = static_cast<PlaqueVisited *>(ev);
+    qCDebug(logMapEvents) << "Attempting a call to script plaque_visited with:"<<casted_ev->m_name<<qHash(casted_ev->m_name);
+
+    //auto val = m_scripting_interface->callFuncWithClientContext(&session,"plaque_visited", qPrintable(casted_ev->m_name), casted_ev->m_pos);
```
broxen: We can remove this old code.

Projects/CoX/Servers/GameServer/TradingService/TradingService.cpp
```diff
@@ -0,0 +1,169 @@
...
+    else
+        qCDebug(logStores) << "Error processing buyItem";
+
+    return nullptr;
+}
+
+std::unique_ptr<ServiceToClientData> TradingService::on_trade_cancelled(Entity* ent, Event* /*ev*/)
+{
+    // unused event
+    //TradeWasCancelledMessage* casted_ev = static_cast<TradeWasCancelledMessage *>(ev);
```
broxen: If these are for future use, let's add a TODO comment with an explanation for why they are here.

Projects/CoX/Servers/GameServer/ZoneTransferService/ZoneTransferService.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    if (!m_owner_instance->get_map_transfers().contains(casted_ev->m_data.m_node_name))
+    {
+        qCDebug(logMapXfers) << QString("Map swap collision triggered on node_name %1, but that node_name doesn't exist in the list of map_transfers.");
+        return nullptr;
+    }
+
+    MapXferData map_transfer_data = m_owner_instance->get_map_transfers()[casted_ev->m_data.m_node_name];
+    //doesn't seem like we are using it atm
+    //Entity *e = getEntityByDBID(ent->m_client->m_current_map, casted_ev->m_data.m_ent_db_id);
```
broxen: Let's remove this old code.

Projects/CoX/Servers/GameServer/ZoneTransferService/ZoneTransferService.cpp
```diff
@@ -0,0 +1,213 @@
...
+            QString door_msg = "You cannot enter.";
+            sendDoorMessage(*ent->m_client, 2, door_msg);
+        }
+        else
+        {
+            // Attempt to send the player to that SpawnLocation in the current map.
+            QString anim_name = "RUNIN";
+            glm::vec3 offset = casted_ev->location + glm::vec3 {0,0,2};
+            sendDoorAnimStart(*ent->m_client, casted_ev->location, offset, true, anim_name);
+            // ent->m_client->m_current_map->setSpawnLocation(*session.m_ent, gotoSpawn);
```
broxen: Why comment this out? Was it commented before? Let's leave a TODO and note of explanation.

eos428: Oops, looks like this wasn't actually commented before. I forgot why I commented this out, but maybe it was due to a linking error or something

Projects/CoX/Servers/GameServer/ClientOptionService/ClientOptionService.cpp
eos428: These were commented out before I moved them into ClientOptionService, but let's un-comment them out

Projects/CoX/Servers/GameServer/ClientOptionService/ClientOptionService.cpp
eos428: It doesn't seem like we are doing anything with this event? CMIIW

Projects/CoX/Servers/GameServer/CharacterService/PowerService/PowerService.cpp
eos428: As we have discussed through PM these events are currently only used to find the session, and then the entity from the session- which we have done in MapInstance already.

So here we have an event in which the params are not used, and thus commented to reduce compiler warnings.

PR #766, opened 2019-02-09
Character creation bug fix and "EMPTY" string tidy

### LINKED ISSUE #753: Character Creation Bug
### BUG DESCRIPTION
A deleted and then remade character _with the same name_ does not start out as a new character and instead retains some information/settings of the previously deleted character.

### To Reproduce
1. Create a character named "Shazam"
2. Enter the world. (Outbreak)
3. Use /mapmenu to transition to Boomtown (or any other map)
4. Delete that character.
5. Make a new character, also named "Shazam"
6. Enter the world.

### Expected behavior
The character expects to enter the Outbreak map.

### Actual behavior
The character unexpectedly starts in Boomtown. It is possible to repeat this behavior of deleting and remaking a character to log into the previous character's last map.

Note that deleting the character, then shutting down and restarting the server before remaking him, results in the expected behavior. (i.e., The character spawns in Outbreak.)

### SEGS Version
0.6.1

miles200:
- Fix new character not spawning in default map (Outbreak) in certain scenarios
- Add const QString EMPTY_CHAR_NAME = "EMPTY" to replace all uses of "EMPTY" to provide clearer use. Added const in GameDBSyncEvents.h as this file is common across all places "EMPTY" is used.

### Issues closed
Closes #753

### DIFF
--- a/Projects/CoX/Common/GameData/Character.cpp
+++ b/Projects/CoX/Common/GameData/Character.cpp
@@ -56,11 +56,11 @@ Character::Character()
 
 void Character::reset()
 {
-    m_name                                  = "EMPTY";
+    m_name                                  = EMPTY_CHAR_NAME;
     m_char_data.m_level                     = 0;
     m_char_data.m_combat_level              = m_char_data.m_level;
-    m_char_data.m_class_name                = "EMPTY";
-    m_char_data.m_origin_name               = "EMPTY";
+    m_char_data.m_class_name                = EMPTY_CHAR_NAME;
+    m_char_data.m_origin_name               = EMPTY_CHAR_NAME;
     m_char_data.m_has_sg_costume            = false;
     m_char_data.m_current_costume_idx       = 0;
     m_char_data.m_using_sg_costume          = false;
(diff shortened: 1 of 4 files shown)


miles200:
Will squash commits shortly

miles200:
@HeraldOfOmega @JamesDW1 be good to get some testing on this if possible, I've done a first round and it looks to be working okay

JamesDW1:
Sure, I can test real quick

Projects/CoX/Common/GameData/Character.cpp
```diff
@@ -56,11 +56,11 @@ Character::Character()
 
 void Character::reset()
 {
-    m_name                                  = "EMPTY";
+    m_name                                  = EMPTY_CHAR_NAME;
     m_char_data.m_level                     = 0;
     m_char_data.m_combat_level              = m_char_data.m_level;
-    m_char_data.m_class_name                = "EMPTY";
-    m_char_data.m_origin_name               = "EMPTY";
+    m_char_data.m_class_name                = EMPTY_CHAR_NAME;
```
broxen: ahh, I see what Magenta was referring to here. Since this is used for more than just "NAME" I would rename the variable `EMPTY_STRING`

Projects/CoX/Common/Messages/GameDatabase/GameDBSyncEvents.h
```diff
@@ -14,6 +14,8 @@
 namespace SEGSEvents
 {
 
+const QString EMPTY_CHAR_NAME = "EMPTY"; // "EMPTY" is used as CharSel screen in client expects this in some cases
```
broxen: I'd modify this comment slightly to be clear there are several places where the client expects EMPTY_STRING.

miles200:
Thanks @broxen changes fixed and pushed in latest commit

PR #769, opened 2019-02-10
[WIP] Mission Maps

### LINKED ISSUE #768: Mission Maps
This is for the implementation of transferring to mission maps.

### To Implement

- Loading scene graphs of mission maps
- Storing map data for mission maps

A lot of the functionality to transfer to mission maps already exists with the map transfers as they seem to work identically for solely transferring. There is additional packets related to missions, however I believe they're related to mission objectives, timers, etc. This can probably all be implemented at a later date, when an actual mission system is implemented.

This isn't particularly difficult as such, it's more that there needs to be a bit of a restructure of how map paths are stored and parsed, due to mission maps having a variety of possible files.

Each mission map type has subtypes based on what I think are level bounds - 15, 30, 45 and 60. Not all map types have all level bounds. These are separate folders.

Sewers/Sewers_15
Sewers/Sewers_30
...
etc

In addition to that, each level-bounded folder has a potential number of layouts and possibly groups of layouts.

Sewers/Sewers_15/Sewers_15_Layout_01_01
Sewers/Sewers_15/Sewers_15_Layout_01_02
Sewers/Sewers_15/Sewers_15_Layout_02_01
Sewers/Sewers_15/Sewers_15_Layout_03_01
..
etc

The biggest difficulty is making a system to handle all these variations in a way that makes sense.

I've currently got a single map loading (although there's a lot of hackiness to get it to load currently), so I can confirm so far at least that no additional packets are required just to transfer, and it's simply a case of providing the mission paths/data to the existing map transfer system.

Magentah:
## Summary
Add functionality to map transfer to mission maps.

### Issues closed
Closes #768 

- Move map data to a file, and potentially update the format to allow for variations (specifically for missions, which have many subsets for maps.
- Update scenegraph to handle parsing of mission file paths when loading mission map scene graphs
- Update MapTemplate to handle mission maps
- Update MapManager to handle mission maps
- Update MapServer to handle mission maps
- Update MapInstance::on_scene_request to handle mission maps.
- Add data for mission maps.
- Update map transfer slash commands and / or mapmenu to handle mission maps (randomise layouts / level ranges  maybe)


Projects/CoX/Common/Runtime/SceneGraph.cpp
```diff
@@ -617,7 +617,9 @@ namespace SEGS {
 
 bool loadSceneGraph(const QString &path,LoadingContext &ctx,PrefabStore &prefabs)
 {
+    qInfo() << "SceneGraph Path: " << path;
```
broxen: We might want to add these to a logging category, to avoid spamming the main console output

Projects/CoX/Servers/MapServer/MapManager.cpp
```diff
@@ -55,6 +55,17 @@ bool MapManager::load_templates(const QString &template_directory, uint8_t game_
 
             qInfo() << "Found map:" << map_dir_visitor.fileInfo().filePath();
         }
+
+        qInfo() << "Directory: " << dirname;
+        if (dirname.contains("Mission_", Qt::CaseInsensitive)) 
+        {
+            qInfo() << "Found Mission Map: " << map_dir_visitor.fileInfo().filePath();
```
broxen: With the addition of these mission maps the output for loading each map file should be categorized. Maybe we provide a generic "Loading maps" type message, and then spit out a warning if one fails to load, otherwise hide each individual map loading message in a category.

Projects/CoX/Servers/MapServer/MapTemplate.cpp
```diff
@@ -32,7 +32,15 @@ MapInstance * MapTemplate::get_instance()
         m_instances.push_back(new MapInstance(m_map_filename,ListenAndLocationAddresses(m_base_loc,port_inc)));
         m_instances.back()->activate(THR_NEW_LWP|THR_JOINABLE|THR_INHERIT_SCHED,1);
         m_instances.back()->spin_up_for(m_game_server_id,m_map_server_id,s_template_id++);
-        m_instances.back()->start(client_filename());
+        if (is_mission_map)
+        {
+            qDebug() << "Is mission map. Mission Level: " << mission_level_cap << " -- Layout: " << mission_layout;
```
broxen: I'd add this debug to a new category as well. Perhaps `logMissions`

broxen (commented):
I know this is WIP, but I left a few comments.

Magentah:
Ya, I was planning on changing all the debug things later. The changes in this so far were just the bare minimum to get it working, and `qInfo()` was quicker, and probably all of it will change before the PR is ready, once I know how I want to structure things.

Projects/CoX/Common/GameData/map_definitions.cpp
```diff
@@ -2,50 +2,177 @@
...
+    //{55000, "Sewers_15_Layout_01_01", "maps/Missions/Sewers/Sewers_15/Sewers_15_Layout_01_01.txt", "Sewers 15 Layout 1"}
 };
 
-const std::vector<MapData> &getAllMapData()
+uint8_t getMissionMapLevelRange(uint8_t char_level)
+{
+    uint8_t current = level_ranges[0];
+    for (auto &level_range : level_ranges)
+    {
+        if (abs(char_level - level_range) < abs(char_level - current))
```
Magentah: I realize this won’t work correctly and needs to change.

Projects/CoX/Common/GameData/map_definitions.cpp
```diff
@@ -2,50 +2,177 @@
...
+        }
+    }
+}
+
+std::vector<MapData> &getAllMapData()
 {
     return g_defined_map_datas;
 }
 
-const MapData &getMapData(const QString &map_name)
```
Magentah: I removed a lot of const due to some changes I was making having issues with it, I’ll eventually add all the consts back where able

Projects/CoX/Common/GameData/map_definitions.h
```diff
@@ -30,12 +39,23 @@ struct Map_Data
     QByteArray Icon;
 };
 
+
+struct MissionMapData
+{
+    uint8_t m_level;        // Sewers_15
+    uint8_t m_layout;       // Sewers_15_Layout_01
```
Magentah: I’m thinking that I’ll change this so that there’s only a single mission data for each level, which has all possible layouts in it.

Also need to think about what to do with the mission maps that don’t have levels/latouts..

Projects/CoX/Servers/MapServer/MapServer.cpp
```diff
@@ -196,7 +201,13 @@ void MapServer::on_expect_client(ExpectMapClientRequest *ev)
         ev->src()->putq(new ExpectMapClientResponse({1, 0, m_base_location}, ev->session_token()));
         return;
     }
-    EventProcessor *instance = tpl->get_instance();
+    // Get character data so we can get the character level for mission map level bounds.
+    GameAccountResponseCharacterData char_data;
+    serializeFromQString(char_data,request_data.char_from_db_data);
+    CharacterData cd;
+    serializeFromQString(cd, char_data.m_serialized_chardata);
```
Magentah: It seems awkward to have to pass character data to the instance (to get their level to give the correct mission map).. I feel like there should be a better option here

Projects/CoX/Servers/MapServer/MapTemplate.cpp
```diff
@@ -52,8 +63,23 @@ void MapTemplate::shut_down_all()
...
+    else
+    {
+        // number suffix is likely the level limit (sewers_30 is levels 16-30)
+        // each mission has a number of different layouts, each with a number of sub bins?
+        // layout_01_01 - layout_01_04 for sewers_30, layout_02_01 - layout_02_04, etc.
+        // "maps/missions/sewers/sewers_15/sewers_15_layout_01_01"
+        // "maps/missions/sewers/sewers_30/sewers_30_layout_02_04"
+        // some folders have audio.
+        QString map_desc_from_path = mission_base_name();   // Sewers, Caves, etc.
+        return QString("maps/missions/%1").arg(map_desc_from_path);
```
Magentah: I’m not sure if I like this either, storing mission maps as just their root, and cities as their whole path. Seems awkward

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -354,10 +354,10 @@ void cmdHandler_MoveZone(const QString &cmd, MapClientSession &sess)
 {
     uint32_t map_idx = cmd.midRef(cmd.indexOf(' ') + 1).toInt();
     if(map_idx == getMapIndex(sess.m_current_map->name()))
-        map_idx = (map_idx + 1) % 23;   // To prevent crashing if trying to access the map you're on.
+        map_idx = (map_idx + 1) % 39;   // To prevent crashing if trying to access the map you're on.
```
Magentah: Should be a constant at least, or just the total number of maps defined..

Projects/CoX/Servers/MapServer/MapServer.cpp
broxen: Just pass their level only. In fact, it probably makes more sense to pass the level only, as when sidekicking or in a Team we may want to pass only one player's level for the mission.

Projects/CoX/Common/GameData/map_definitions.h
broxen: I like the idea of keeping that mission data together and calling upon the level/layout as needed. For those without, those values would be zero? Or could be -1?

Projects/CoX/Servers/MapServer/SlashCommand.cpp
broxen: Well... this was sort of a hotfix to provide a random zone option. With the changes you're making (and have made to zoning) I'd say we can probably remove the random map change altogether.

Besides, on live, if you "zone" into a map you were already on (`/stuck` for instance) you would simply reload the zone or teleport to the new location. This is the desired behavior here.

Projects/CoX/Servers/MapServer/MapTemplate.cpp
broxen: Are there instances where maps require their entire name? Or simply sections of it?

We could store the parts in a QStringList and join with `/` when needed?

Magentah: Maps only ever use their full name for anything as far as I remember, the shortened name is only used when requesting a new instance of a map

Magentah:
There's a couple of issues that I've found playing around with this a bit..

1. The COT mission zone seems to load in very briefly, then disappears and becomes entirely black - the same as if the map didn't load.
2. Outdoor_Industrial and Outdoor_Forest both spawn the player out of bounds, and /stuck doesn't help. The other 2 Outdoor zones seem to spawn you kinda correctly, although you're above the floor still.

I haven't implemented 2 of the mission map types yet (unique and Outdoor_Missions) due to their structure being a bit different to everything else, and I want to think about how best to handle them. 

Outdoor_Missions has sub folders for different types of Outdoor mission - Ruined, Forest, etc - then has folders for each layout. Those folders then have several files each, that I'm entirely unsure how to handle at the moment.

e.g.
```
maps/Missions/Outdoor_Missions/Outdoor_City/Outdoor_City_01/Outdoor_City_01a.bin
maps/Missions/Outdoor_Missions/Outdoor_City/Outdoor_City_01/Outdoor_City_01b.bin
maps/Missions/Outdoor_Missions/Outdoor_City/Outdoor_City_01/Outdoor_City_01c.bin

maps/Missions/Outdoor_Missions/Outdoor_Unique/Outdoor_Unique_Forest_01/Outdoor_Unique_Forest_01c.bin
```

Unique missions have a completely different structure all together..

```
maps/Missions/unique/Interdimensional/Interdimensional_01_01.bin
maps/Missions/unique/TrialRooms/Sewers/trial_01_01_room.bin
maps/Missions/unique/TrialRooms/Sewers/trial_01_01_room_spawns.bin
maps/Missions/unique/TrialRooms/Sewers/trial_01_01_room_beacons.bin
maps/Missions/unique/TrialRooms/Sewers/trial_01_01_room_floorbeacons.bin
maps/Missions/unique/jumppuzzles/Jumppuzzles_Layout_01.bin
```

As of right now, you can load in to any of the other mission maps, I just need to work out the best way to handle selecting the layouts as currently it will always select the first layout found.

Magentah:
Just to give some info about how the mission maps are currently structured --

Each mission map lair type has a single map definition.
```cpp
{55000, "5th_Column", "5th_Column", MapType::MISSION },
```
During startup, the server will go through the maps directory for missions and load all the mission data. For the indoor mission types such as caves, tech, sewers, etc, this means navigating through the folders to the bin files, and building a structure like so:

- Sewer
    - Mission Level Range: 15
        - Mission Layouts:
            - Layout 01
            - Layout 02
                - Sub Layout 01
            - Layout 02
                - Sub Layout 02
    - Mission Level Range: 30
        - Mission Layouts
            - Layout 01
            - Layout 02_01
            - Layout 03_01

For outdoor missions, it's a similar process, however the level range is fixed at 15 (so they're always accessible in the current system).

Then from this structure we dynamically create the file paths when requested, based on the level of the character requesting the map and the layout requested (although the layout selection part isn't implemented yet)

Projects/CoX/Common/GameData/map_definitions.h
Magentah: Those without a level currently use the lowest level "range" value of 15 so that they're always returned. I don't like how this has ended up being structured for outdoor missions though, due to them not really having levels / sublayouts. The unique / outdoor_missions folders will cause more issues probably. I may need to try to think of a new data structure that will be able to handle all the map types without weird uses of some properties (like levels for outdoor missions not existing, but being needed anyway due to how the mission maps work at the moment)

YellowAdder:
Map names appear at least synonymous with map size which is more informative to present unless we mean to restrict by level, which I don't think we should do. (What makes a map a "level 45 range map"? Just size?) If someone wants to make a low-level mission inside a large map, barring some technical or mechanical reason, why not? 

Ultimately, mission designers should define whether a map is appropriate for their X-level missions. This is how the AE system does it and, if it existed, AE drops any pretense of map's level association. 

(A selected enemy group appears more strongly related to a mission's level range and difficulty than anything with the chosen map. I could be wrong on that, but if I am, I'd very much be interested in how since it has system implications for missions and spawning.) 

In AE, maps are categorized like:

- Map Type
  - Map Length or Category
    - Map

Map Type corresponds to _sets_ such as the City of Villain and Cargo Ship sets. For example:

- Council Set
  - Small | Medium | Large
    - Map 0X to Map XX
- Office Abandoned - City of Heroes Set
  - Tiny | Small | Medium | Large
    - Map 0X to Map XX
- Office Abandoned - City of Villains Set
  - Tiny | Small | Medium | Large
    - Map 0X to Map XX
- Office Transition to Caves Set
  - Tiny | Small | Medium | Large
    - Map 0X to Map XX

Unlike all other types, Unique Maps uses _category_ instead of map length. Atypical maps, such as the outdoor group, are considered categories under this map type. For example:

- Unique Maps
  - Outdoor Maps
    - Atlas Park 01
    - City Map 01 (Industrial)
    - Forest 01
    - Graveyard 01
  - Office
    - Dark Astoria Office
    - Frostfire
    - Etc.

This seems like a pretty good model to mimic, especially since it makes it easy to accommodate future map sets.

Magentah:
The structure you provided is essentially the same as what I'm currently using, except that I use 15, 30, 45 and 60 instead of tiny, small, medium and large as it made sense at the time to use the sections from the file name. There's no specific reason that it's called "level" other than it's the first name I thought of that'd kinda fit. I can change the name to size and it'd be the same, functionally.

The issue I'm having is how to structure the actual MapData struct, and the sub mission structs, such that it makes sense and fits all map types (cities, indoor missions, outdoor missions, unique missions) without having weird edge cases and if statements for some things when navigating through the folder directory.

Loading maps based on size or level is something that can be changed later once there's ways to actually load them outside of the mapmenu or movezone commands. I don't want to populate the map definitions with each size of each lair just to be able to access them when it shouldn't be that way once the relevant systems are created, so choosing the size based on the characters level was the easiest way to be able test the current system. I may look in to creating a new /missionmap command specifically for navigating through all the mission maps.

YellowAdder:
> There's no specific reason that it's called "level" other than it's the first name I thought of that'd kinda fit.

That was my instinct too. What changed my mind was seeing the maps with the idea that people were only getting tiny ones up to level 15. They're T.I.N.Y. I was thinking, noooo. This can't be?

>The issue I'm having is how to structure the actual MapData struct, and the sub mission structs, such that it makes sense and fits all map types ...

My first question would be, do we have to handle all types right now? For a first wash, using an approach that doesn't concern itself with edge cases doesn't seem unreasonable. Should we immediately provide for what may be indiscernible map variants, for example? We're going to learn more that may change how we think about their organization and handling as it is. 

For it's part, AE doesn't provide for these possible variants unless they are as you guessed: multiple levels for their base map. That's a good guess, but I'm not convinced yet. I think you're instead right about them being base-map variants. That might be less related to layout alterations as to "map details," offering different sets of spawn and objective markers. You can probably discover whether that's true without too much effort?

>I may look in to creating a new /missionmap command specifically for navigating through all the mission maps.

You're undoubtedly the best judge on how sensible it is compared to cramming it all in the main command. I rather like the idea.

Magentah:
> > The issue I'm having is how to structure the actual MapData struct, and the sub mission structs, such that it makes sense and fits all map types ...
> 
> My first question would be, do we have to handle all types right now? For a first wash, using an approach that doesn't concern itself with edge cases doesn't seem unreasonable. Should we immediately provide for what may be indiscernible map variants, for example? We're going to learn more that may change how we think about their organization and handling as it is.

I would prefer to make it as reasonable as can be expected right now, rather than just making it work. Just making it work is the easy part, and this would probably be done by now if I were going with that route. But I think that even if we are to change things in the future, having a sensible base to start from makes things easier. 

After thinking things through a bit more I think I may have come up with a somewhat simplistic way to make it work that'll be enough for now, so we'll see..

CMakeLists.txt
```diff
@@ -155,6 +155,21 @@ set(MAPNAMES
   Trial_03_01
   Trial_04_01 Trial_04_02
   Trial_05_01
+  Mission_5th_Column
```
broxen: These will need to be added to the new map list in Projects\CoX\CMakeList.txt for scripts, otherwise we'll get some errors there.

Or perhaps consolidated there since these folders are only needed for the CoX project.

Projects/CoX/Common/GameData/map_definitions.cpp
```diff
@@ -2,50 +2,247 @@
...
+void getMissionMapLevelData(QFileInfo map_level_folder, MapData &map_data)
+{
+    // maps/Missions/Sewers/Sewers_15/
+    QString level = map_level_folder.fileName().mid(map_level_folder.fileName().lastIndexOf("_") + 1);
+    QDir map_layout_dir(map_level_folder.filePath());
+    map_layout_dir.setFilter(QDir::Dirs | QDir::Files | QDir::NoSymLinks | QDir::NoDot | QDir::NoDotDot);
+
+    MissionMapData mission_data;
+    mission_data.m_mission_name = map_level_folder.baseName();
+    qInfo() << "Mission name: " << mission_data.m_mission_name;
```
broxen: All of 17 of these `qInfo()` messages in this PR should be put into a logging category, because we don't really want to spam the server console with this information.

`qInfo()` should really be short informative text that the server is working properly:
`Loading all maps... done!` with more verbose output reserved for categories that can be toggled for troubleshooting, etc.

Projects/CoX/Common/GameData/map_definitions.cpp
```diff
@@ -2,50 +2,247 @@
...
+            }
+            if (mapDir.exists())
+            {
+                for (auto &map_level : mapDir.entryInfoList())
+                {
+                    if (map_level.isDir())
+                    {
+                        if (map_data.m_map_type == MapType::MISSION)
+                        {
+                            getMissionMapLevelData(map_level, map_data);
```
broxen: Per your discussion with YellowAdder, perhaps we change this to `map_size` or something, as it doesn't have a direct correlation with level (server ops may use any size map for any sort of difficulty or player level)

Projects/CoX/Common/GameData/map_definitions.cpp
broxen: I'd like to see the consts replace where possible prior to merging this PR.

Projects/CoX/Common/GameData/map_definitions.cpp
```diff
@@ -2,50 +2,247 @@
...
+std::vector<MapData> &getAllMapData()
 {
     return g_defined_map_datas;
 }
 
-const MapData &getMapData(const QString &map_name)
+MapData &getMapData(QString &map_name)
 {
-    for (const auto &map_data : g_defined_map_datas)
+    for (auto &map_data : g_defined_map_datas)
```
broxen: here for example, we aren't even changing (nor should we) `g_defined_map_datas`

Projects/CoX/Common/GameData/map_definitions.cpp
```diff
@@ -130,33 +327,45 @@ QString getMapName(uint32_t index)
...
     {
         if(map_name.contains(map_data.m_map_name, Qt::CaseInsensitive))
-            return map_data.m_map_path;
+        {
+            if (!map_data.m_mission_data.empty())
+            {
+                return getMissionPath(map_data.m_map_name, MissionCategory::MEDIUM);
+            }
+            return QString("maps/City_Zones/%1/%1.txt").arg(QString(map_data.m_map_name));
```
broxen: What about Trial Zones etc?

Projects/CoX/Common/GameData/map_definitions.h
```diff
@@ -30,14 +53,67 @@ struct Map_Data
...
     uint32_t m_map_idx;
     QByteArray m_map_name;             // City_00_01, City_01_01, etc...
-    QByteArray m_map_path;             // The ones ending with .txt
     QByteArray m_display_map_name;     // Outbreak, Atlas Park...
+    MapType m_map_type;
+    QByteArray m_map_path;             // The ones ending with .txt
+    std::vector<MissionMapData> m_mission_data;
 };
 
+/*
```
broxen: This may be useful to have here, but let's add a blurb at the top to describe what it is we're looking at and why it's relevant.

Projects/CoX/Servers/MapServer/MapTemplate.cpp
```diff
@@ -52,8 +63,23 @@ void MapTemplate::shut_down_all()
...
-    QString map_desc_from_path = base_name();
-    return QString("maps/city_zones/%1/%1.txt").arg(map_desc_from_path);
+    if (!m_is_mission_map)
+    {
+        QString map_desc_from_path = base_name();
+        return QString("maps/city_zones/%1/%1.txt").arg(map_desc_from_path);
+    }
+    else
+    {
+        // number suffix is likely the level limit (sewers_30 is levels 16-30)
```
broxen: Given your conversation with YellowAdder regarding levels and map sizes, perhaps this comment is dated and could use a rewrite.

broxen (changes requested):
Prior to merge, I'd like to see a lot of the items here addressed. Constness and logging for sure, also some questions about comments and variable names. Overall great PR, just want to tidy it up before it goes in!

Thanks Magenta! I know you're busy right now, so no rush on any of this.

Pushing this out to our next release v0.8

nemerle:
This was merged 'by hand'

PR #770, opened 2019-02-10
Lua Relay Race / Mission Objective Timer

### LINKED ISSUE #735: Lua Pit Stop Race
## Feature Summary
Using Lua, create a milestone race with 5-6 contacts or phone boxes that you must click on before proceeding to the next "pit stop". Your time to reach each pitstop would be recorded, and at the end you'd receive a Souvenir based upon your speed (bronze, silver, gold).

Could even spawn the pit stop NPCs at badge locations, or simply use phone call boxes, or maybe a combination of both. I think 4-6 "stops" is probably enough.

Bonus points if you can make the `/kiosk home home` city info terminals show a leader board

ZopharShinta:
## Summary
Adds Lua Relay Race
Adds Packet 73 - MissionObjectiveTimer

### Issues closed
#735 

Moved Contacts, Clues, Souvenirs and tasks into PlayerData
Added PlayerStatistic object in PlayerData to store HideAndSeek count, RelayRaceResults, Hunt counts and future player statistics.

The database change will require a new database or @broxen 's Database upgrade tool


Projects/CoX/Common/GameData/PlayerStatistics.h
```diff
@@ -0,0 +1,64 @@
...
+
+    uint16_t m_segment;
+    uint32_t m_last_time;
+    uint32_t m_best_time;
+
+    template<class Archive>
+    void serialize(Archive &archive, uint32_t const version);
+};
+
+struct Hunt
```
broxen: Although modern c++ doesn't make much of a distinction between classes and structs, I look at structs as containers for Simple Data Objects only. If we start to add methods and logic, then I prefer to see these objects as classes.

Projects/CoX/Common/GameData/playerdata_serializers.cpp
```diff
@@ -34,6 +34,11 @@ void serialize(Archive &archive, PlayerData &pd, uint32_t const version)
     archive(cereal::make_nvp("Gui", pd.m_gui));
     archive(cereal::make_nvp("KeyBinds", pd.m_keybinds));
     archive(cereal::make_nvp("Options", pd.m_options));
+    archive(cereal::make_nvp("KnownContacts", pd.m_contacts));
```
broxen: Can we add auth data here as well? Or is it stored in the DB elsewhere?

Projects/CoX/Common/GameData/Clue.h
```diff
@@ -68,15 +68,3 @@ class Souvenir
 };
 
 using vSouvenirList = std::vector<Souvenir>;
-
-class ClueSouvenirList
```
broxen: Where did all these classes go? I don't see them anymore. Won't we need these class definitions?

Projects/CoX/Common/GameData/Contact.cpp
```diff
@@ -5,18 +5,11 @@
...
  */
 
-#include "Character.h"
+#include "playerdata_definitions.h"
 #include "Logging.h"
 #include "serialization_common.h"
 #include "serialization_types.h"
 
-
-void addContact(CharacterData &cd, Contact &contact)
```
broxen: Why remove this? How do we add contacts now?

Projects/CoX/Common/GameData/Clue.cpp
```diff
@@ -44,18 +44,3 @@ void Souvenir::serialize(Archive &archive, uint32_t const version)
 
  CEREAL_CLASS_VERSION(Clue, Clue::class_version)   // register Clue class version
  SPECIALIZE_CLASS_VERSIONED_SERIALIZATIONS(Clue)
-
- template<class Archive>
- void ClueSouvenirList::serialize(Archive &archive, uint32_t const version)
```
broxen: I don't see where we serialize clues again. Am I missing it?

Projects/CoX/Common/GameData/playerdata_serializers.cpp
ZopharShinta: It was never stored so I didn't add it. I can though I'm not sure what its for.

Projects/CoX/Common/GameData/Clue.h
ZopharShinta: This is the combined Clue and Souvenir list.
It was just an object to contain a Clue list and a Souvenir list.
Since I pulled those out to be in PlayerData. This object isn't needed.

Projects/CoX/Common/GameData/Contact.cpp
ZopharShinta: This was unused. DataHelpers has the functions that add/update and remove them

Projects/CoX/Common/GameData/Clue.cpp
ZopharShinta: This is the parent object being serialized. Since I pull the children out and placed them in PlayerData, this object isn't needed.

Projects/CoX/Common/GameData/playerdata_serializers.cpp
broxen: Ahh sends client authorizations for preorder unlocks and things. It should probably be at the account level, but I think I stuck it here because it was easily accessible from where we send it to the client (I think per Character)

There's the possibility that the original game had these auths per Character, because we don't yet know what each flag does.

Projects/CoX/Common/GameData/PlayerStatistics.h
ZopharShinta: So all the objects that are being serialized should be classes? I'm ok with that.

Projects/CoX/Common/GameData/Clue.cpp
broxen: Yes of course. I didn't think it through

Projects/CoX/Common/GameData/PlayerStatistics.h
broxen: We tend to leave them as structs where it's only the archive methods, but for Hunt where you actually have getters/setters, then I'd make it a class.

PR #777, opened 2019-02-14
logging: Name logs with ISO-8601 compliant dates

dracc:
### Summary:
Parsing big endian dates is hard. ( 30th of April `300419` should not be less than 31st of March `310319` )
There is a date format ISO standard for a reason.
Let's use it.

And while we're at it, change the file name format from `[logname]_[date].log` to `[date]_[logname].log` as suggested (and agreed upon) in Discord chat.


dracc:
Please note that I do not have access to SEGSAdmin - my changes to `SettingsManager.cpp` have yet to receive further testing than compilability.

broxen:
Will await the verdict of the `date_name.log` vs `name_date.log` discussion before proceeding here.

dracc:
We've had 3 votes in favor and none opposing so far regarding the log file name format change.

miles200:
Looks good thanks @dracc

PR #781, opened 2019-02-17
Server side encounter spawning / WIP

ZopharShinta:
## Summary
Adds server spawning of Enemies/citizens in encounters.
Still early and I expect lots of comments/changes but wanted to get something available for us to start making adjustments.

### Issues:
The rotation of entities is off on some spawns. Other "look" correct.


Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1667,5 +1695,82 @@ RelayRaceResult getRelayRaceResult(MapClientSession &cl, int segment)
...
+    e->m_char->m_max_attribs.m_HitPoints = 100;
+    e->m_char->m_max_attribs.m_Endurance = 100;
+    e->m_char->m_char_data.m_current_attribs.m_HitPoints = 100;
+    e->m_char->m_char_data.m_current_attribs.m_Endurance = 100;
+
+    //Sets target info menu faction. Skull, Hellions, Freakshow, etc
+    e->m_faction_data.m_faction_name = faction_name;
+
+    //Required to send changes to clients
+    e->m_pchar_things = true;
```
broxen: I'd move this to the initializer, where my other comment is.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1478,6 +1478,34 @@ void logSpawnLocations(MapClientSession &cl, const char* spawn_type)
...
+        return;
+    }
+    int count = 0;
+
+    for(auto const &s: spawners)
+    {
+        glm::mat4 mat4 = s;
+        glm::vec3 loc = glm::vec3(mat4[3]);
+        qCDebug(logScripts) << "Spawn: " << spawn_name << " loc x: " << loc.x << " y: " << loc.y << " z: " << loc.z ;
+        if(count / 3 == 0) {
```
broxen: Please put brackets on their own line

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1478,6 +1478,34 @@ void logSpawnLocations(MapClientSession &cl, const char* spawn_type)
...
+        qCDebug(logScripts) << "spawners empty";
+        return;
+    }
+    int count = 0;
+
+    for(auto const &s: spawners)
+    {
+        glm::mat4 mat4 = s;
+        glm::vec3 loc = glm::vec3(mat4[3]);
+        qCDebug(logScripts) << "Spawn: " << spawn_name << " loc x: " << loc.x << " y: " << loc.y << " z: " << loc.z ;
```
broxen: Let's move all of these to logSpawn or some new category instead

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1667,5 +1695,82 @@ RelayRaceResult getRelayRaceResult(MapClientSession &cl, int segment)
...
+
+    //Sets target info menu faction. Skull, Hellions, Freakshow, etc
+    e->m_faction_data.m_faction_name = "Citizen";
+
+    //Required to send changes to clients
+    e->m_pchar_things = true;
+
+    forcePosition(*e, loc);
+    forceOrientation(*e, ori);
+    qCDebug(logScripts()) << QString("Created Victim with ent idx:%1 at location x: %2 y: %3 z: %4").arg(e->m_idx).arg(loc.x).arg(loc.y).arg(loc.z);
```
broxen: Let's move all of these to a logSpawn as well. You don't need the parentheses.

logScripts should really be left for logging scriptingEngine operations. But if we want a new category maybe something like logCritters would make sense.

Projects/CoX/Common/GameData/EntityHelpers.cpp
```diff
@@ -266,6 +266,53 @@ void initializeNewNpcEntity(const GameDataStore &data, Entity &e, const Parse_NP
...
+    {
+        // Get timestamp in ms
+        auto now_ms = std::chrono::steady_clock::now().time_since_epoch().count();
+
+        p.m_position = e.m_entity_data.m_pos;
+        p.m_pyr_angles = e.m_entity_data.m_orientation_pyr;
+        p.m_timestamp = now_ms;
+        addPosUpdate(e, p);
+    }
+    e.m_pchar_things = true;
```
broxen: This is duplicated. Let's remove the second one.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1478,6 +1478,34 @@ void logSpawnLocations(MapClientSession &cl, const char* spawn_type)
...
 }
 
+void logEncounterSpawnLocations(MapClientSession &cl, const char* spawn_type)
+{
+    QString spawn_name = QString::fromUtf8(spawn_type);
+    auto spawners = cl.m_current_map->getEncounterSpawners();
+
+    if(spawners.empty())
+    {
+        qCDebug(logScripts) << "spawners empty";
```
broxen: This one too.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1478,6 +1478,34 @@ void logSpawnLocations(MapClientSession &cl, const char* spawn_type)
...
+    int count = 0;
+
+    for(auto const &s: spawners)
+    {
+        glm::mat4 mat4 = s;
+        glm::vec3 loc = glm::vec3(mat4[3]);
+        qCDebug(logScripts) << "Spawn: " << spawn_name << " loc x: " << loc.x << " y: " << loc.y << " z: " << loc.z ;
+        if(count / 3 == 0) {
+            QString name = "Thug_Contaminated_01";
+            QString test = "test";
```
broxen: What is test? Faction or name? Let's rename this variable to make it clear what it is.

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -3316,4 +3318,83 @@ void MapInstance::clearTimer(uint32_t entity_idx)
         this->m_lua_timers[count].m_remove = true;
 }
 
+void MapInstance::spawn_enemies()
+{
+    uint32_t diff = 3; // control number spawned in each group
```
broxen: I'd rename this something like `spawn_limit`

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1667,5 +1695,82 @@ RelayRaceResult getRelayRaceResult(MapClientSession &cl, int segment)
...
+    e->m_char->m_max_attribs.m_HitPoints = 100;
+    e->m_char->m_max_attribs.m_Endurance = 100;
+    e->m_char->m_char_data.m_current_attribs.m_HitPoints = 100;
+    e->m_char->m_char_data.m_current_attribs.m_Endurance = 100;
+
+    //Sets target info menu faction. Skull, Hellions, Freakshow, etc
+    e->m_faction_data.m_faction_name = "Citizen";
+
+    //Required to send changes to clients
+    e->m_pchar_things = true;
```
broxen: This we should set to false, as victims will never have Attribs or FX

Projects/CoX/Servers/MapServer/DataHelpers.cpp
broxen: Come to think of it, I believe logSpawn is for player spawning, so maybe a new category is best.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
cyanpill: Just as long as they are able to emote, most victims were cowering or doing the bag pull emote.
Also hero should be set to false, so they won't be hit by any powers.

YellowAdder:
The server should refer to an associated Spawndef file (in Lua) for the encounter that would define how NPCs (hostile or otherwise) are generally positioned among the current group of spawn markers, an encounter table for the sort of critters that might spawn, as well as defining what they do and say for the encounter itself.

We get a glimpse of how City of Heroes originally handled this via files that Kai recently provided. For example, a hostage who would be rescued would include something akin to these entries in the Spawndef:

> 		Placement Back
>		Quantity 1
>		InactiveSpeech "Has the cake been found yet?"
>		ActiveSpeech Trouble!
>		CompassDescriptionSingular "Head Baker"
>		PersonInactiveSpeech "Please don't hurt me!"
>		PersonActiveSpeech "Help! They've locked the staff in the cooler!"
>		PersonRescuedSpeech "Thank you $name....is the cake alright?"
>		PersonAnimation HandsOnHead_Kneel
>		PersonRescuedAnimation CHANTING

These were undoubtedly abstracted for players using the AE designer compared to what they used internally. Meaning, the original Spawndefs were more explicit, so instead of "Back" as placement it would've identified Encounter_V_40 for the hostage position, for example. We should give that sort of control to content creators as well. 

By changing even just a few lines they could substantially alter a template, and the encounter that spawns in the world. (I don't think scripters would have any problems understanding or using this more explicit model, and would certainly prefer it.) It will give them quite a bit of control while leaving the handling of when encounters can be generated (via encounter managers or whatever other mechanisms get used) obfuscated within the server. 

There are a few other questions here, however, regarding how much data is exposed to scripters and whether the server actually manages the literal act of spawning (and just references the Spawndef) or whether scriptors will have more exposed that allow them to even tweak positioning and orientation data that would otherwise be kept within the server. Should I maybe make those proposed approaches clearer in a post back in the original conversation? Or would it be more appropriate here since it deals with what information the server might want to expose?

ZopharShinta:
@YellowAdder I would add those to the original conversation. Just because this entire aspect of SEGS is much larger and complex than I expected. So I think there will be multiple PR's to iterate on this and get us to where we want to be for Server and Scripting spawns. We don't want to miss something.

This does add `MapInstance.AddEnemy` to the scripting side. So at the very least scripts can add enemies at custom locations with it.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
ZopharShinta: I'm still unsure as to what entity type citizen's are. I remember them being blue when you selected them. So they would be a critter? But you need to set their health or its 0. Maybe they were NPC's in the early issues then moved to something else later?

Projects/CoX/Servers/MapServer/DataHelpers.cpp
ZopharShinta: logEntitySpawn?

YellowAdder:
Oh yes, I forsee a good bit of iterating on this subject. That's the reason I'd like to see us stick with more scripting and a tighter scope in the early stages as it's faster and easier to prototype. We also won't require much from the server outside of exposing some data (that you're now handling) and a small number of new methods.

Regardless, I'll hop back to the original discussion with some further thoughts. Great work so far, @ZopharShinta ! Very exciting! :+1:

Projects/CoX/Servers/MapServer/DataHelpers.cpp
broxen: Cyan has a good point, but `pchar_things` are only for FX (particles) and triggered moves (animations) are flagged separately by `m_update_anims` (presumably when animations change)

I don't remember actually being able to con victims, so I don't remember them having a con color. But perhaps the community can help us find some source material to narrow it down?

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -145,11 +145,13 @@ void MapInstance::start(const QString &scenegraph_path)
                 scene_graph_loaded = m_map_scenegraph->loadFromFile("./data/geobin/" + scenegraph_path);
                 m_all_spawners = m_map_scenegraph->getSpawnPoints();
                 m_map_transfers = m_map_scenegraph->get_map_transfers();
+                m_spawn_encounters = m_map_scenegraph->get_encounters();
             }, "Loading original scene graph");
 
         TIMED_LOG({
             m_map_scenegraph->spawn_npcs(this);
             m_npc_generators.generate(this);
+            spawn_enemies();
```
broxen: This should belong to NPCGeneratorStore, or maybe it's own CritterGeneratorStore

Projects/CoX/Servers/MapServer/MapSceneGraph.cpp
```diff
@@ -47,9 +48,13 @@ bool MapSceneGraph::loadFromFile(const QString &filename)
         return false;
     for(SceneNode *def : m_scene_graph->all_converted_defs)
     {
+        if(def->m_name.contains("4076", Qt::CaseInsensitive))
+            bool pause = true;
```
broxen: What's the purpose of this? I don't see it used anywhere else.

Also, `pause` should probably be renamed to avoid confusion with `system("pause")` or it's expected behavior.

ZopharShinta: Oh, this was my way of seeing how ids in the Bins appear in the scenegraph.
It's just used as a breakpoint to look at the object. I'll remove it.

PR #786, opened 2019-02-19
rpc: Add WebUI statistics procedure

dracc:
## Summary
Send all data used for the WebUI statistics table in one command.

This will have to be expanded in the future, but number of online characters, accounts and characters in general is not available at this time afaik.

### DIFF
--- a/Projects/CoX/Servers/AuthServer/AdminRPC.cpp
+++ b/Projects/CoX/Servers/AuthServer/AdminRPC.cpp
@@ -105,10 +105,11 @@ QString AdminRPC::getStartTime()
     return m_start_time;
 }
 
-QVariantList AdminRPC::getWebUIData()
+QVariantMap AdminRPC::getWebUIData()
 {
     qCDebug(logRPC) << "Someone requested WebUI information...";
-    return {VersionInfo::getAuthVersionNumber(), m_start_time};
+    return {std::pair<QString,QString>("version",VersionInfo::getAuthVersionNumber()),
+	    std::pair<QString,QString>("starttime", m_start_time)};
 }
 
 void startRPCServer()
--- a/Projects/CoX/Servers/AuthServer/AdminRPC.cpp
+++ b/Projects/CoX/Servers/AuthServer/AdminRPC.cpp
@@ -105,6 +105,12 @@ QString AdminRPC::getStartTime()
     return m_start_time;
 }
 
+QVariantList AdminRPC::getWebUIData()
+{
+    qCDebug(logRPC) << "Someone requested WebUI information...";
+    return {VersionInfo::getAuthVersionNumber(), m_start_time};
+}
+
 void startRPCServer()
 {
     static jcon::JsonRpcServer *m_server;
--- a/Projects/CoX/Servers/AuthServer/AdminRPC.cpp
+++ b/Projects/CoX/Servers/AuthServer/AdminRPC.cpp
@@ -110,12 +110,21 @@ QString AdminRPC::getStartTime()
  * @return Returns a QMap<QString,QVariant> with server information
  */
 
-QVariantMap AdminRPC::getWebUIData()
+QVariantMap AdminRPC::getWebUIData(QString const& version)
 {
     qCDebug(logRPC) << "Someone requested WebUI information...";
     QMap<QString,QVariant> ret;
-    ret.insert("version", VersionInfo::getAuthVersionNumber());
-    ret.insert("starttime", m_start_time);
+    if (version.compare("0.6.1") == 0)
+    {
+      ret.insert("version", VersionInfo::getAuthVersionNumber());
+      ret.insert("starttime", m_start_time);
+    }
+    else
+    {
+      // Default to v0.6.1 format
+      ret.insert("version", VersionInfo::getAuthVersionNumber());
+      ret.insert("starttime", m_start_time);
+    }      
     return ret;
 }
 
--- a/Projects/CoX/Servers/AuthServer/AdminRPC.cpp
+++ b/Projects/CoX/Servers/AuthServer/AdminRPC.cpp
@@ -105,11 +105,18 @@ QString AdminRPC::getStartTime()
     return m_start_time;
 }
 
+/*!
+ * @brief Get a dictionary of information as used by the WebUI
+ * @return Returns a QMap<QString,QVariant> with server information
+ */
+
 QVariantMap AdminRPC::getWebUIData()
 {
     qCDebug(logRPC) << "Someone requested WebUI information...";
-    return {std::pair<QString,QString>("version",VersionInfo::getAuthVersionNumber()),
-	    std::pair<QString,QString>("starttime", m_start_time)};
+    QMap<QString,QVariant> ret;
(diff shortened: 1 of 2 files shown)


Projects/CoX/Servers/AuthServer/AdminRPC.cpp
```diff
@@ -105,6 +105,12 @@ QString AdminRPC::getStartTime()
     return m_start_time;
 }
 
+QVariantList AdminRPC::getWebUIData()
```
nemerle: Maybe we should return json-encoded struct here?

dracc: I guess we could but I don't see the point as jcon-cpp has support for `QVariantList` and converts it into an JSON array as expected.

nemerle: True, but `getWebUIData` suggests a data-blob :)
And since it's a blob, we might as well return dictionary of stuff

dracc: Guess I should rename the function to `getWebUIArray` then? :+1:

nemerle: How do You envision webUI using this?
`getWebUIArray` -> [ stuff1, stuff2 ]
What are we going to put in that array, what will the index mean ?

dracc: You sure make a valid point. I'll have to put a bit more thought into this!

nemerle: Since this will be part of the project's external interface, we will end up with a pretty stringent requirement of breaking this only on major versions. Best case would be to have something extensible here, so when we add new data here, the older webUI interfaces do not break.
We could provide a versioned dictionary:
```json
{
 "Data_v0.7.0": {},
 "Data_v0.7.1": {},
}
```
where every version has data format that is 'set in stone' when the version is released.
And then we deprecate the older versions on major release ( 1.0/2.0 )

Or we provide a much more granular interface:
```cpp
QVariantMap getData(const QString &path, const QString &version);
```
Where we allow some kind of `path` expression to refer to an exposed piece of server data and we take into account the requested version.

Projects/CoX/Servers/AuthServer/AdminRPC.cpp
```diff
@@ -105,6 +105,13 @@ QString AdminRPC::getStartTime()
     return m_start_time;
 }
 
+QVariantMap AdminRPC::getWebUIData()
+{
+    qCDebug(logRPC) << "Someone requested WebUI information...";
+    return {std::pair<QString,QString>("version",VersionInfo::getAuthVersionNumber()),
+	    std::pair<QString,QString>("starttime", m_start_time)};
```
miles200: Don't think anything wrong with this, just personal preference

I'd just do
```
`QVariantMap<QString, QString> result;
result.insert("version", VersionInfo::getAuthVersionNumber());
result.insert("starttime", m_start_time);

return result;
```

broxen: There are actually performance benefits to using QMap (or QVariantMap) vs std::pair here. Especially if we ever need to find something by it's "key" like `version`

broxen: Also the format of return looks off to me, I think this would return `QVariantMap< std::pair<...>, std::pair<...> >` which I dont think is the intention =P

dracc: `QVariantMap` is just a typedef for `QMap<QString, QVariant>` and one of the `QMap` constructors take one or more `std::pair` in an initializer list as argument.
Thus the `QMap.key` is still available. :+1:

I can see that the readability might be bad (as proven by your comments @broxen) - I should probably do it the way @miles200 suggested.

dracc:
@miles200 I believe I would want to add those bits but I'm still not sure how.

Projects/CoX/Servers/AuthServer/AdminRPC.cpp
```diff
@@ -105,6 +105,20 @@ QString AdminRPC::getStartTime()
...
+ * @brief Get a dictionary of information as used by the WebUI
+ * @return Returns a QMap<QString,QVariant> with server information
+ */
+
+QVariantMap AdminRPC::getWebUIData()
+{
+    qCDebug(logRPC) << "Someone requested WebUI information...";
+    QMap<QString,QVariant> ret;
+    ret.insert("version", VersionInfo::getAuthVersionNumber());
+    ret.insert("starttime", m_start_time);
```
broxen: So reading Nem's comments regarding versioning this blob...

I think he's proposing that we nest the UIData in json objects keyed by version so that users who haven't updated WebUI will still properly receive this blob. Looking at web-ui, it seems the only way this becomes an issue is if a key name changes. We might want to consider this possibility.

To achieve this, you'd simply pass a QMap to your QMap:
```cpp
    // current data set
    QVariantMap cur_blob;
    cur_blob.insert("version", VersionInfo::getAuthVersionNumber());
    cur_blob.insert("starttime", m_start_time);

    // previous data set
    QVariantMap v6_blob;
    // whatever we used to send in v6
    v6_blob.insert("old_start_time", m_start_time); // key name is different!

    QVariantMap ret;
    ret.insert("Data_v0.7.0", cur_blob);
    ret.insert("Data_v0.6.0", v6_blob);
    // so on.

    return ret;
```
this would return a json blob like
```json
{
    "Data_v0.7.0": {
        "version" : "v0.7.0",
        "starttime" : "03:00:00 UTC"
    },
    "Data_v0.6.0": {
        "old_start_time" : "03:00:00 UTC"
    },
}
```

Magentah: I would argue that specifying the version at call time rather than in the resulting data is a better idea. I'm not entirely sure how this would be best with rpc, but nem suggested passing version numbers as a parameter.

Something like

```cpp
QMap<QString, QVariant> getData(QString &path, QString &version)
{
    switch (version)
    {
        case "v0.7.0":
            return getData_v070(path);

        case "v0.7.1":
            return getData_v071(path);
    }
}
```

It looks pretty ugly though...

dracc: What has me confused is mainly the `path` argument. Adding a `version` switch case ought to be easy enough though.

Magentah: `path` is just an example parameter

broxen: That certainly works. I'm not as familiar with the rpc call, but I assume you can pass a parameter to it (version) to receive the correct response.

I agree with Magentah, I don't think path is needed. I think Nems suggestion was that it would be an expression identifying what information the server should send, but I can't foresee a scenario where "all" isn't the answer, or where the overhead in sending "all" is high enough to warrant selective parsing.

For the Data_v# methods, I'd put them in their own header so it doesn't pollute the rpc files

PR #788, opened 2019-02-23
Character constructor issue

### LINKED ISSUE #689: Character constructor fires 10+ times per Entity
## Summary
Character constructor is firing multiple times per Entity.

### Steps to Reproduce
Set debug message in Character constructor, and login. Will fire more than 8 times on a new account with no Characters

### Expected behavior
We should be generating Character once and passing that object around until it needs to be destroyed.

ZopharShinta:
## Summary
Creates Character object for each character slot, passes references where needed rather than creating new objects.

### Issues closed
Closes #689


nemerle:
I've talked about this issue with @broxen yesterday. The underlying problem is not the fact that we construct temporary character objects in here:
```cpp
for(int i=0; i<m_data.m_max_slots; i++)	    for(int i=0; i<m_data.m_max_slots; i++)
    {	    {
        Character converted;	
        EntityData entity_data;	        EntityData entity_data;
        PlayerData player_data;	        PlayerData player_data;
        toActualCharacter(m_data.m_characters[i], converted, player_data, entity_data);	        toActualCharacter(m_data.m_characters[i], m_characters->at(i), player_data, entity_data);
        converted.serializetoCharsel(tgt, getDisplayMapName(entity_data.m_map_idx));	        m_characters->at(i).serializetoCharsel(tgt, getDisplayMapName(entity_data.m_map_idx));
    }
```
The construction itself could be simplified a bit ( reinitializing a temporary object instead )
but the **only** way this code would significantly slow things down would be if it was IO bound ( logs/db access )
So perhaps somwhere in this loop we spam the logs ?

ZopharShinta:
I only noticed two places where the constructor was called in the character selection screen.
`CharacterResponse::serializeto` and `CharacterSlots::serializeto` that @nemerle listed.
Both are for the `toActualCharacter` call.

After the change I made it calls its 8 times, once for each character slot. I suppose we could replace the vector was a single object then replace that with the data from `toActualCharacter`?

Projects/CoX/Common/Messages/Game/GameEvents.h
```diff
@@ -173,12 +174,19 @@ class CharacterResponse : public GameLinkEvent
...
+        m_data=gad;
+        m_character = character;
+    }
     void serializeto(BitStream &bs) const override;
     void serializefrom(BitStream &bs) override;
     // [[ev_def:field]]
     uint8_t m_index;
     // [[ev_def:field]]
     GameAccountResponseData m_data;
+    Character *m_character;
```
broxen: After discussing with Nem: we don't want to have raw pointers in our events, because events should all be serializeable. Raw pointers aren't.

ZopharShinta:
I moved the character object out of the loop and removed my extra object on the GameSession object.
Not sure where the extra log messages are as I didn't see any till I added a log line to the Character constructor.

@broxen mentioned maybe it is the `logDB` lines in the `toActualCharacter` call. But they don't seem to be spamming, at least to me.
I see one per character slot and an extra for the character selected.

Not sure what other adjustments to make.

Projects/CoX/Common/Messages/Game/GameEvents.h
```diff
@@ -12,6 +12,7 @@
 //#include "CRUDP_Protocol/CRUD_Link.h"
 #include "CRUDP_Protocol/CRUD_Events.h"
 #include "Servers/InternalEvents.h"
+#include "GameData/Character.h"
```
broxen: Do we need this header include?

Can we include in the source file instead?

Projects/CoX/Servers/GameServer/GameHandler.h
```diff
@@ -43,6 +45,8 @@ struct GameSession
     eTravelDirection m_direction;
     uint8_t m_access_level=0;
 
+    Character m_character;
```
broxen: I guess we don't really need this, or the include anymore

ZopharShinta: Nope. We don't need this or the include. I removed them both.

Projects/CoX/Servers/GameServer/GameHandler.h
```diff
@@ -7,11 +7,13 @@
 
 #pragma once
 #include "EventProcessor.h"
+#include "Common/GameData/Character.h"
```
broxen: I don't think we need this header include

Projects/CoX/Servers/GameServer/GameHandler.h
```diff
@@ -7,11 +7,13 @@
...
 #include "Common/Servers/ClientManager.h"
 #include "Messages/GameDatabase/GameDBSyncEvents.h"
 
 #include <unordered_set>
 class CharacterClient;
+class Character;
```
broxen: I dont think we need this class predeclaration

ZopharShinta: Done and done

PR #793, opened 2019-03-03
Scripting engine refactor

ZopharShinta:
## Summary
Refactored ScriptingEngine.
Corrected issue with ScriptingEngine.cpp size by moving RegisterTypes out into separate cpp file. 

Started exposing spawn points to scripting engine.


ZopharShinta:
Yeah. It builds fine locally (before and after the change). The Travis mac build doesn't like Sol it seems.

Projects/CoX/Servers/MapServer/ScriptingEngine/ScriptingEngine_Helpers.cpp
```diff
@@ -0,0 +1,250 @@
...
+    m_private->m_lua["heroName"] = qPrintable(client->m_name);
+    m_private->m_lua["Player"]["entityId"] = client->m_ent->m_db_id;
+    if(client->m_ent->m_player->m_tasks_entry_list.size() > 0)
+    {
+        m_private->m_lua["vTaskList"] = client->m_ent->m_player->m_tasks_entry_list[0].m_task_list;
+    }
+    return callFunc(name,arg1,loc);
+}
+
+std::string ScriptingEngine::callFunc(const char *name, int arg1)
```
broxen: Nem mentioned using a template for these in discord.

```
template <typename... Args>
std::string ScriptingEngine::callFunc(const char *name, Args&&... args)
{
    sol::protected_function funcwrap = m_private->m_lua[name];
    funcwrap.error_handler           = m_private->m_lua["ErrorHandler"];
    if (!funcwrap.valid())
    {
        qCritical() << "Failed to retrieve script func:" << name;
        return "";
    }
    sol::protected_function_result result = funcwrap.call(std::forward<Args>(args)...);
    if (!result.valid())
    {
        sol::error err = result;
        qCritical() << "Failed to run script func:" << name << err.what();
        return "";
    }
    return result.get<std::string>();
}
```

nemerle: This change will come with my physics PR, no need to duplicate it?

broxen: Awesome. Sounds good

Projects/CoX/Servers/MapServer/ScriptingEngine/ScriptingEnginePrivate.cpp
```diff
@@ -0,0 +1,57 @@
+/*
+ * SEGS - Super Entity Game Server
+ * http://www.segs.io/
+ * Copyright (c) 2006 - 2019 SEGS Team (see AUTHORS.md)
+ * This software is licensed under the terms of the 3-clause BSD License. See LICENSE.md for details.
+ */
+
+#pragma once
```
broxen: We never use `#pragma once` in cpp files

Projects/CoX/Servers/MapServer/ScriptingEngine/ScriptingEnginePrivate.h
```diff
@@ -0,0 +1,25 @@
+/*
+ * SEGS - Super Entity Game Server
+ * http://www.segs.io/
+ * Copyright (c) 2006 - 2019 SEGS Team (see AUTHORS.md)
+ * This software is licensed under the terms of the 3-clause BSD License. See LICENSE.md for details.
+ */
+
```
broxen: We should `#pragma once` in **every** header file

Projects/CoX/Servers/MapServer/ScriptingEngine/ScriptingEngine_RegisterTypes.cpp
```diff
@@ -0,0 +1,707 @@
...
+#include "Messages/Map/SendVisitLocation.h"
+#include "Common/GameData/Character.h"
+#include "Common/GameData/CharacterHelpers.h"
+#include "Common/GameData/Contact.h"
+#include "Common/GameData/Entity.h"
+#include "Common/GameData/Contact.h"
+#include "Common/GameData/Task.h"
+#include "TimeHelpers.h"
+
+#define SOL_CHECK_ARGUMENTS 1
```
broxen: We can define this once in a common header instead of in each source file

Projects/CoX/Servers/MapServer/ScriptingEngine/ScriptingEngine.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include "MapInstance.h"
+#include "GameData/playerdata_definitions.h"
+#include "DataHelpers.h"
+
+#include <QtCore/QFileInfo> // for include support
+#include <QtCore/QDir>
+#include <QtCore/QDebug>
+
+using namespace SEGSEvents;
+//static constexpr const int MAX_INCLUDED_FILE_SIZE=1024*1024; // 1MB of lua code should be enough for anyone :P
```
broxen: We're still using this in Helpers. Where is it defined now? If not here, we should delete this comment

ZopharShinta:
Ok. This should be most of the changes requested.
I really want to move the lambda's we can into classes and expose those through usertypes.
Since we have a lot more refactoring going on, I'll wait and play around with that later.
It might not even be an option with the MapInstance and MapClientSession calls we make.

Projects/CoX/Servers/MapServer/ScriptingEngine/ScriptingEngine_Null.cpp
```diff
@@ -17,6 +17,9 @@
 ScriptingEngine::ScriptingEngine() {}
 ScriptingEngine::~ScriptingEngine() {}
 void ScriptingEngine::registerTypes() {}
+void ScriptingEngine::register_GenericTypes() {}
```
broxen: Since we aren't toggling scripting_engine build flag anymore, this entire file can go away

Projects/CoX/Servers/MapServer/ScriptingEngine/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+
+SET (target_SOURCES
+${target_CPP}
+${target_INCLUDE}
+)
+
+add_library(ScriptingEngine ${target_SOURCES})
+target_link_libraries(ScriptingEngine PUBLIC Qt5::Core gameData)
+
+target_compile_definitions(ScriptingEngine PUBLIC -DSCRIPTING_ENABLED=1)
```
broxen: You don't need this line anymore

Projects/CoX/Servers/MapServer/ScriptingEngine/ScriptingEngine.h
```diff
@@ -6,20 +6,49 @@
  */
 
 #pragma once
+
+#include "ScriptingEnginePrivate.h"
+#include "DataHelpers.h"
```
broxen: Do we need all these headers? They would be better in the source files where their methods are used.

Here in the header the computer tries to include them every time this header is called. It speeds up compilation if we only use them where absolutely needed.

PR #798, opened 2019-03-11
Updating version for -dev build

broxen:
## Summary
Releases should have their own fixed version number, so we can tell from the logs if someone is using a release package or a dev build. Would be great if there were a way to append build number so we know what build someone is running from github. I can't think of a way to automate that easily.

**Comments on how to handle this moving forward are welcome**

### Issues closed
None filed.

- Iterate build number to v0.6.5

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,51 @@ PROJECT(SEGS)
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+# for version and revision info. Derived from https://github.com/dolphin-emu/dolphin/blob/master/CMakeLists.txt#L132
+find_package(Git)
+if(GIT_FOUND)
+  # defines SEGS_REVISION
+  execute_process(WORKING_DIRECTORY ${PROJECT_SOURCE_DIR} COMMAND ${GIT_EXECUTABLE} rev-parse HEAD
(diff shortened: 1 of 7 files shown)


ldilley:
We could formulate a solution based on Ryan Pavlik's answer here: https://stackoverflow.com/questions/1435953/how-can-i-pass-git-sha1-to-compiler-as-definition-using-cmake

@miles200 also brought up using the hash in Discord.

Utilizing Ryan's CMake modules, we could generate a file where `Version.h` resides containing various git data. The branch and hash (for example -- and/or other arbitrary git data such as tag, date, etc.) can be contained in the output that is displayed from `ACE_TMAIN()` on startup. This would allow people to easily identify the software being troubleshot once the end user provides a copy of their `output.log`.

ldilley:
Creating a custom CMake script based off an existing example works as well. Something capable of producing a dynamic string like the following should suffice:

```
SEGS v0.6.1-develop (Outbreak) [d34db33fc4f3]
```

broxen:
I like this approach

broxen:
Latest commit produces strings like:
```cpp
[02:38:51] SEGS v0.6.6c968fe6 (Outbreak) [version-dev: 6c968fe6531442fae5f528075f32747afc0367b6]
```

ldilley:
Nicely done, sir. I favor the convention you used in `Version.h`. I'll let you merge when ready in case there are any other tweaks you may have forgotten.

PR #801, opened 2019-03-13
Spawn system

YellowAdder:
## Summary
Allows for dynamic, customizable spawning via Lua using server provided Scenegraph data.

Spawning is automatic when the first player joins a map area. Control quantities of spawns by altering values found in each map's "entities.lua" script. (See the RandomSpawn() function below for examples.) 

Only Atlas, Outbreak and Perez Park currently support encounter spawning. Hazard and Trial zones can typically spawn civilians and vehicles, but are not currently configured to do so. Modifying this is pretty simple. (Look at Atlas's entities.lua for an example.) "Normal" zones include spawning of civilians and vehicles by default. 

Be warned that excessive spawning will cause noticeable slow downs. Zones without encounter spawns have moderately high civilian spawn values currently set.

Use /script RandomSpawn(int x) to manually generate encounter spawns
Use /script RandomScript(int x, "Cars") to manually generate vehicle spawns
Use /script RandomScript(int x, "Civilians") to manually generate NPC civilian spawns.

e.g.: RandomSpawn(25) or RandomSpawn(50, "Cars")


Projects/CoX/Data/scripts/City_04_02/entities.lua
```diff
@@ -0,0 +1,76 @@
...
+    end
+  
+--[[ NPC chat message test
+    MapClientSession.npcMessage(client, 'Hello Hero!', id)
+    MapClientSession.npcMessage(client, 'What are you doing here?', id)
+    ]]
+    return ""
+end
+
+dialog_button = function(id) -- Will be called if no callback is set
```
broxen: We repeat this code in every file, can we move it to a global include?

YellowAdder: Those files will (probably) become more unique as scriptors begin to customize each area. They just look alike because there is scant content. @ZopharShinta may be able to answer this better though as it's going to depend on how contacts work and how unique NPCs are interacted with for each map. For now, entities.lua is that future "hub" for controlling interaction. There just isn't much of that going on yet.

broxen: Ahh, I see. Is there a way to override functions in Lua? I'm not as familiar. But maybe we provide a default and if there's an override present in the file it supersedes?

YellowAdder: Technically, no. But it's possible to accomplish more or less the same thing: http://www.luafaq.org/#T1.22.

ZopharShinta: If I remember correctly, the dialog_button function is a backup if the script didn't set a dialog button callback function.
The server actually checks if a callback is set, if not, it uses the generic dialog_button function.
I think this could be moved into the Global.lua file as the scripters scripts should set the callback for the dialogs they create.

broxen:
I'd rather see TimeCop built into the server side of the spawn system.

Also, why the friction change in surfaceparams? What does that fix?

YellowAdder:
It made acceleration/speed faster while flying in giant maps like Independence Port.  Or any map, for that matter. Could've sworn I set that back before I committed, but apparently I overlooked it. Side note: 

>  { 1.50f, 3.00f, 0.00f, 3.00f, 3.00f },

That or 2/3/03/4 was really pleasant. I've been setting it to one or the other all the time.

Beyond that, I agree that TimeCop could/should be built into the server, and the scripts are redundant as the stand. (The only significant difference is the Time Cop's position in each map) This was the fastest solution I could muster while getting everything else debugged and set up for a better experience. Despite any of that, Time Cop is a pretty big QoL addition to the spawn system for both admins and users, but I didn't have more time, ironically, to refine it.

I'm also not familiar enough with the server-side timer system to alter it yet. @ZopharShinta and I have discussed making a more generalized timer system in recent days. (Time Cop is based on @ZopharShinta's hide and seek system, as you probably figured.) This is something that will be addressed in the not so distant future as it will be needed for at least a handful of important features.

YellowAdder:
Thanks to @HeraldOfOmega, we know that the problem for Linux is that "include_lua" isn't able to load files in a parallel directory, that being the spawndefs in our case. I presume this is some sort of permission issue. 

(e.g.: include_lua('../spawndefs/somefile.lua').)

Unless someone knows how to resolve that, the other option is for me to move each map's spawndefs into a /spawndef subdirectory. (I don't think this will run into the same permission issue. Someone correct me if I'm wrong. It should be easy to test, though.) 

This approach shouldn't result in duplicated files, but it isn't the best solution because we want spawndefs to be more generally accessible for possible re-use among maps in the future. It's a simple restructuring, however, and it will solve the issue for now. 

Your thoughts, @broxen? @ZopharShinta?

broxen:
This crashed once for me upon creating a new character with a new database, but then I was unable to replicate even after multiple attempts at restarting everything and overwriting the databases, sooo :man_shrugging: 

Aside from that, I have the same Linux issue that others are having. It's clearly filesystem related. I discussed with @YellowAdder and it sounds like we have a permanent fix worked out. Will wait for his commit before merge.

As for the Clang failure, this is fixed in Zophar's #793 PR

PR #805, opened 2019-03-14
[WIP] Pseudo physics

nemerle:
WIP

World class is taking over some responsibilities that MapInstance handled before.
Add WorldPhysics class, which will be responsible for:
* movement processing
* collision detection
* scene range queries ( retrieving entities in a configurable volume )


nemerle:
will be replaced by engine world&physics.

PR #806, opened 2019-03-17
Add XP Multiplier to SEGSAdmin

### LINKED ISSUE #497: Add Double XP/XP Multiplier to SEGSAdmin
Those of us that played a lot remember getting the email notices and in game notices about Double XP weekends and the the like. This would be a nice setting to have in SEGSAdmin that allows the server admin to do this easily. Here’s my idea: 

* A tick box to activate the feature. 
* When active, start and end date fields become active to set the length of time. 
* A multiplier box with a default of 2.

Thoughts?

Frozyn:
## Summary
Added a Modifiers page to SEGSAdmin with experience-related fields. 

The user is able to:
- Enable the experience modifier.
- Set the experience multiplier.
- Set the date range in which this multiplier is in effect.

Added a grantXP function to grant yourself an amount of experience. The grantXP function is effected by the multiplier set within the settings.

### Issues closed
Closes #497 

- Settings::setDefaultSettings set default values for the modifier fields.
- Add grantXP function to CharacterHelpers.cpp and CharacterHelpers.h.
- Add modifier settings to settings_template.cfg.
- Add xp_mod_checkbox_validator function to SettingsDialog.h and SettingsDialog.cpp. It will enable start date, end date, and multiplier if checked, and disable these fields if unchecked.
- Create a signal for the checkbox in SettingsDialog constructor. xp_mod_checkbox_validator called on checkbox click.
- Add xp_mod_checkbox_validator call to SettingsDialog::open_settings_dialog.
- SettingsDialog::read_config_file will read the modifier settings and set their respective UI fields.
- SettingsDialog::generate_default_config_file will write to the config file the modifier settings defined in settings_template.cfg.
- SettingsDialog::save_changes_config_file will write to the config file the values set in their respective UI fields.
- SettingsDialog::set_default_values will set the modifier UI fields with default values and call xp_mod_checkbox_validator.
- Replaced all_fields_validator in SettingsDialog::field_validator with validation_error_text. Added date range validator to this function to prevent saving an incorrect date range. 
- Add "Modifiers" page to SEGSAdmin with the "Experience" group. The group contains a checkbox, double_spin, and two datetime_edit widgets.
SEGSAdmin page

Here is a demonstration with the multiplier enabled and set to 2. Note that the grantXP slash command was not committed, it's just for testing purposes.

demonstration


Projects/CoX/Common/GameData/CharacterHelpers.cpp
```diff
@@ -101,6 +103,27 @@ void setXP(Character &c, uint32_t val)
     c.m_char_data.m_experience_points = val;
 }
 
+void grantXP(Character &c, uint32_t val) {
```
broxen: We should remove this and incorporate your modifiers into the existing `giveXP()` method

Also, please put brackets on their own line per our [coding style guide](https://github.com/Segs/Segs/blob/develop/docs/CONTRIBUTING.md#coding-styleguide)

Projects/CoX/Common/GameData/CharacterHelpers.cpp
```diff
@@ -101,6 +103,27 @@ void setXP(Character &c, uint32_t val)
...
+    QSettings settings(Settings::getSettingsPath(),QSettings::IniFormat);
```
broxen: We should move this to `GameDataStore::read_settings()` to minimize I/O calls. We can store the variables in GameDataStore.h like we do `m_time_to_afk`

Projects/CoX/Data/settings_template.cfg
```diff
@@ -126,3 +126,9 @@ log_scenegraph      = false
 log_stores          = false
 log_tasks           = false
 log_rpc             = false
+
+[Modifiers]
```
broxen: Nitpick, but can you please move this above `Logging` category. Logging is long and less relevant and some users will still edit this file manually.

Projects/CoX/Utilities/SEGSAdmin/SettingsDialog.cpp
```diff
@@ -79,6 +80,22 @@ void SettingsDialog::auto_logout_checkbox_validator()
     }
 }
 
+void SettingsDialog::xp_mod_checkbox_validator()
+{
+    if(ui->xp_mod_check->isChecked())
```
broxen: This if/else could be simplified:
```cpp
    bool is_checked = ui->xp_mod_check->isChecked();
    ui->xp_mod_multiplier_spin->setEnabled(is_checked);
    ui->xp_mod_startdate_edit->setEnabled(is_checked);
    ui->xp_mod_enddate_edit->setEnabled(is_checked);
```

Frozyn:
Thank you for your comments @miles200 @broxen! I've committed the refactored code for this issue.

One small note, the slash command is now giveXP rather than grantXP.

broxen:
Great First PR! Once this clears CI I'll merge it in

PR #817, opened 2019-03-26
[WIP] Initial AI combat response

cyanpill:
## Summary
Initial step of combat AI response, recording aggro, and responding with a text message in say range. Stalled until other systems progress.

- Checks aggro values to make sure highest is first
- aggro modifier from archetypes 
- Proximity aggro, which activates only the one critter
- Aggro only dropped when target dies, or goes out of range.

### Improvements To Do:
- Taunt takes priority(maybe add 1000 to values when flag is true?)
- EM activates entire spawn of critters
- Movement of critters to get into range, and turning to face target.
- Actually use powers, which requires critters storing powers, and lots of tweaks to how powers work. Might need to load a list of melee and a list of ranged powers, then have AI for: noattacks, melee only, prefer melee, prefer ranged


ldilley:
Attempting to build this PR with Visual Studio 2017 Pro to see why it chokes on the range-based loop syntax.

broxen:
This can probably merge if the Appveyor issue can be resolved

PR #818, opened 2019-03-26
Persistent revealed fog of war

### LINKED ISSUE #371: Minimap State needs to store array and serialize to database
Opening your map you'll see the "fog of war" or shadow over the map tiles. Moving around uncovers or reveals these map tiles. We need to store the array of tiles per map, and serialize them to the database in some form. Possibly under some kind of `PlayerProgress` struct that will include contact lists, clues, badges and souvenirs in the future.

`MapInstance::on_minimap_state(...)` in MapInstance is a place to start. That event has the following variables:
- `uint32_t tile_idx;` - the index of the tile being uncovered

Steps to complete this feature request:
- [ ] Get current map_idx for use in minimap_state
- [ ] Store tile idx in nested array grouped by map_idx
- [ ] Store array in PlayerData (maybe new struct called PlayerProgress? Is it necessary?)
- [ ] Serialize minimap_state_array to database for cross-session persistence
- [ ] Determine which opcode is used to receive map state upon map transfer or login (aka, how does the client know what parts of the map are revealed?)
- [ ] Serialize array from database back to array stored in PlayerData
- [ ] Send updated array to client upon log in, or map transfer based upon map_idx

Feel free to discuss or add additional features in the comments below. I'll add relevant info here.

Frozyn:
## Summary
Revealed map cells are saved to the database and are used to reveal the fog of war across sessions.

### Issues closed
Closes #371 

- Added cereal load and save functions to `serialization_types.h` for the type `std::map<uint32_t, std::array<bool, 1024> >`.
- Added 3 new files; `PlayerProgress.cpp`, `PlayerProgress.h`, and `VisitMapCells.h`. Added them to their respective `CMakeLists.txt` files.
- Added `PlayerProgress m_player_progress` member variable to the `PlayerData` struct.
- Added an archive for `m_player_progress` to `playerdata_serializers.cpp`.
- Created `PlayerProgress` struct for storing all map tiles that the player has revealed so far.
- Included `VisitMapCells.h` to `GameCommandList.h`.
- Uncommented the event declaration for `evVisitMapCells` map event type.
- Created `VisitMapCells` class. The class serializes to the BitStream data for setting the opacity of the fog of war, the number of map tiles (always 1024), and an array of uint8_t for determining which tiles are revealed.
- Added `sendVisitMapCells()` to `DataHelpers.cpp` and `DataHelpers.h`.
- `MapInstance::on_mini_map_state()` now also sets an element in `m_visible_map_cells` to true whose index corresponds to the revealed map tile from the triggered event.
- `MapInstance::on_client_resumed()` invokes `sendVisitMapCells()` using `m_visible_map_cells` data.
- Added `m_player_progress` to the instantiation of `PlayerData` in `MapInstance::send_character_update()` and `MapInstance::send_player_update()`.

### Demonstration
The following link demonstrates revealing the fog of war on log in and after map transfer (both using `on_client_resumed` event).
https://i.gyazo.com/f5fcb2c89fcc552902499992d5177f37.mp4

### Issues
A bug(?) exists where, if the map is left open while logging in or transferring to another map, then the fog of war won't be revealed as it should. The player won't be able to reveal any fog of war in tiles that were already supposed to be revealed unless they were to enter a tile they haven't yet revealed, in which case all of the tiles that should have been revealed will be revealed. If the map is closed before logging in or transferring maps, then the map is revealed correctly. Should we look into closing the map prior to these events?


Projects/CoX/Common/GameData/playerprogress.cpp
```diff
@@ -0,0 +1,28 @@
+/*
+ * SEGS - Super Entity Game Server
+ * http://www.segs.io/
+ * Copyright (c) 2006 - 2018 SEGS Team (see AUTHORS.md)
+ * This software is licensed under the terms of the 3-clause BSD License. See LICENSE.md for details.
+ */
+
+#include "PlayerProgress.h"
```
cyanpill: Minor problem- you named the files with lowercase, but reference with uppercase. Linux doesn't like that!

Frozyn: Doh! Good eye, I'll take care of that

nemerle: Just a note, most of our headers are UpperCase :)

cyanpill:
I had the same bug when testing, almost like it's not refreshing the map if it was already open? No idea if we even have a command to force the map to refresh, but that might do it.

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -377,6 +377,12 @@ void sendFloatingNumbers(MapClientSession &sess, uint32_t tgt_idx, int32_t amoun
     sess.addCommand<FloatingDamage>(sess.m_ent->m_idx, tgt_idx, amount);
 }
 
+void sendVisitMapCells(MapClientSession &sess, bool is_opaque, std::array<bool, 1024> visible_map_cells)
+{
+    // qCDebug();
```
broxen: Please remove this commented line.

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2151,6 +2155,8 @@ void MapInstance::on_client_resumed(ClientResumedRendering *ev)
         map_server->session_xfer_complete(session.link()->session_token());
     }
 
+    sendVisitMapCells(session, false, session.m_ent->m_player->m_player_progress.m_visible_map_cells[session.m_current_map->m_index]);
```
broxen: TODO: in a future PR we need to check the map type to verify if `is_opaque` should be true/false.

Projects/CoX/Common/GameData/playerdata_definitions.h
```diff
@@ -15,22 +15,23 @@
...
 #include "Common/GameData/Task.h"
-
+#include "Common/GameData/PlayerProgress.h"
 
 #include <stdint.h>
 
 // this part of the Entity is for Players only info
 struct PlayerData
 {
     enum : uint32_t {class_version   = 3};
```
broxen: Since you've changed this struct by adding `PlayerProgress` please iterate the class_version number to `4`

Projects/CoX/Common/GameData/PlayerProgress.h
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+#pragma once
+#include <QString>
+#include "cereal/cereal.hpp"
+#include "Logging.h"
+
+struct PlayerProgress
+{
+    enum : uint32_t {class_version = 2};
```
broxen: I think this is the first iteration of `PlayerProgress`, so class_version should be `1`

broxen:
Thank you for reviewing and testing @cyanpill 

I haven't pulled down to test, but for the map bug, we don't appear to have a way to minimize the map window from the server. This _could_ be a client bug.

PR #836, opened 2019-04-03
Update WhoAll

### LINKED ISSUE #831: Wrong level shows in /whoall
## Summary
Using `/whoall` shows the wrong level number for players. Looks to be `-1`.

This is likely because the client indexes levels at 0, and we're not adjusting for `/whoall` we should.

## Proposed fix
This one is dead simple. Check the code for `/whoall` in SlashCommand.cpp, and ensure that it's adjusting level + 1 when outputing to the console.

Marking `good-first-issue`

HeraldOfOmega:
## Summary
Show proper level in /whoall

### Issues closed
Closes #831

### DIFF
--- a/Projects/CoX/Servers/MapServer/SlashCommand.cpp
+++ b/Projects/CoX/Servers/MapServer/SlashCommand.cpp
@@ -1544,8 +1544,8 @@ void cmdHandler_WhoAll(const QString &/*cmd*/, MapClientSession &sess)
     {
         Character &c(*cl->m_ent->m_char);
         QString    name      = cl->m_ent->name();
-        QString    lvl       = QString::number(getLevel(c));
-        QString    clvl      = QString::number(getCombatLevel(c));
+        QString    lvl       = QString::number(getLevel(c)+1);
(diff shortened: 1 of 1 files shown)


Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -1544,8 +1544,8 @@ void cmdHandler_WhoAll(const QString &/*cmd*/, MapClientSession &sess)
     {
         Character &c(*cl->m_ent->m_char);
         QString    name      = cl->m_ent->name();
-        QString    lvl       = QString::number(getLevel(c));
-        QString    clvl      = QString::number(getCombatLevel(c));
+        QString    lvl       = QString::number(getLevel(c)+1);
```
broxen: Thanks Herald, You rock!

One nit pick, but since it may not be clear to future devs why we've added this `+1` to these values, can we add a comment to each line explaining that the server stores these values with a 0 index, and reference Issue #831?

Thank you!!!

PR #842, opened 2019-04-05
Alleviate some build warnings

dracc:
## Summary
Compilation outputs quite a few warnings for minor stuff such as unused variables and/or function arguments.
I decided to comment them out as they don't have any effect currently but could serve a good reminder if someone wants to use something similar in the future.

Look at my changes as deprecation before removal? :sweat_smile:


broxen:
Thank you for this!

Once CI clears I'll merge

PR #847, opened 2019-04-06
Fix for debug logs not showing with new settings.cfg

### LINKED ISSUE #846: SEGSAdmin doesn't set log_generic to appropriate values
### Summary
SEGSAdmin generates a new cfg if one does not exist, but it doesn't correctly set the value for `log_generic` from the template file.

We should set `log_generic         = *.debug=true\nqt.*.debug=false` so that debug output is properly shown in the console window. Building with Release should disable this output anyway.

It's useful to allow it to be configured, but setting to false by default seems like unexpected behavior and may make troubleshooting with a user unfamiliar with the code base more difficult.

broxen:
## Summary
SEGSAdmin generated settings.cfg does not currently pull values from the template file, and makes the assumption that `log_generic` takes a boolean value. This PR corrects this by ignoring `log_generic` for the logging tab (no checkbox will be generated) and by culling the default value from the settings_template.cfg

It also sets a default value for config file version, in case one isn't present.

Closes #846

### DIFF
--- a/Projects/CoX/Utilities/SEGSAdmin/SettingsDialog.cpp
+++ b/Projects/CoX/Utilities/SEGSAdmin/SettingsDialog.cpp
@@ -203,10 +203,11 @@ void SettingsDialog::read_config_file(QString filePath)
     logging_grid->setColumnStretch(2,1);
     for(int i = 0; i < config_file.childKeys().size(); ++i)
     {
+        if (config_file.childKeys().at(i) == "log_generic")
+            continue; // skip this one
+        
         if (config_file.childKeys().at(i) == "combine_logs")
             ui->combine_logs->setChecked(config_file.value("combine_logs", "").toBool());
-        else if(config_file.childKeys().at(i) == "log_generic")
-            continue; // just skip it
         else
         {
             QString temp_key = config_file.childKeys().at(i);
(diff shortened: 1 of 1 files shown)


Projects/CoX/Utilities/SEGSAdmin/SettingsDialog.cpp
```diff
@@ -205,6 +205,8 @@ void SettingsDialog::read_config_file(QString filePath)
     {
         if (config_file.childKeys().at(i) == "combine_logs")
             ui->combine_logs->setChecked(config_file.value("combine_logs", "").toBool());
+        else if(config_file.childKeys().at(i) == "log_generic")
+            continue; // just skip it
```
miles200: Could just be me but this doesn't seem in the right place? 
If ```if (config_file.childKeys().at(i) == "combine_logs")``` evaluates to `true` then your `else if` log_generic will never be executed?

broxen: Maybe I misunderstood the logic here. It seems like if combine_logs is true, then it doesn't show any checkboxes at all?

Projects/CoX/Utilities/SEGSAdmin/SettingsDialog.cpp
```diff
@@ -203,6 +203,9 @@ void SettingsDialog::read_config_file(QString filePath)
     logging_grid->setColumnStretch(2,1);
     for(int i = 0; i < config_file.childKeys().size(); ++i)
     {
+        if (config_file.childKeys().at(i) == "log_generic")
+            continue; // skip this one
+
```
miles200: Just one more question, it's probably my mis understanding. This will be executed first right and won't try and execute the 2nd if clause? (like async?)

broxen: It's looping through.

So if `at(i) == "log_generic"` it skips and then loops again

PR #839, opened 2019-04-04
Build Output Folder reorganization

broxen:
This PR aims to reorganized the build output folder, as it was getting unruly. It creates some helper methods in Settings.cpp to handle finding the root folder regardless of what application is calling settings.

Also I did some minor code cleanup here and there.

Please TEST TEST TEST @Segs/testers 

### Issues closed
Closes #556 

### Additions/modifications proposed in this pull request:
- Organize the output of the out folder, and standardize our path names
- Give the db files an extension, so it's clear to the end user where to look if they need to edit manually (aka, reduce support question "where is my database file?")

### Proposed file structure
- :open_file_folder: *compile output folder*
  - :open_file_folder: data
  - :open_file_folder: default_setup
    - :open_file_folder: mysql
    - :open_file_folder: pgsql
    - :open_file_folder: sqlite
    - settings_template.cfg
  - :open_file_folder: logs
    - output.log
  - :open_file_folder: maps
  - :open_file_folder: scripts
  - :open_file_folder: utilities
    - :open_file_folder: mapviewer_data
    - :open_file_folder: unit-tests
      - AuthDBEventRegistry
      - AuthEventRegistry
      - CommonEventRegistry
      - CRUDPEventRegistry
      - event_gen
      - FriendshipEventRegistry
      - FriendshipServiceFunctions
      - GameDBEventRegistry
      - GameEventRegistry
      - ManifestBuilder
      - MapEventRegistry
    - AuthExerciser
    - binConverter
    - dbtool
    - mapviewer
    - piggtool
    - SEGS.exe
    - slav
    - texConverter
  - segs_server
  - SEGSAdmin
  - segs.db
  - segs_game.db
  - settings.cfg
  - README.md

### Notes
This was broken out of my original DBTool PR #778, which will remain focused on dbtool-v5

PR #854, opened 2019-04-10
Simplify byte_sum adjustment in VisitMapCells.h

broxen:
## Summary
Small tweak to VisitMapCells.h to simplify the math on byte_sum to `byte_sum |= 1<<j;`


broxen:
@dracc to answer your question in #839:

In short, because a bitwise and of 0 and 1 is always zero, and powers of two give a 0 in the column related to value. It's explained a couple of useful ways in this stack discussion: https://stackoverflow.com/questions/1053582/how-does-this-bitwise-operation-check-for-a-power-of-2

PR #857, opened 2019-04-11
Proposed fix for settings file creation and other db drivers

broxen:
## Summary
Resolves the errors @Otlichno was having when using DB Drivers other than QSQLITE

### Issues closed
No issue filed

- Resolves issues with settings.cfg not being created properly
- Uses full file path for SQLite but a short name for other db drivers

### DIFF
--- a/Components/Settings.cpp
+++ b/Components/Settings.cpp
@@ -109,12 +109,12 @@ QString Settings::getTemplateDirPath()
 void Settings::createSettingsFile(const QString &new_file_path)
 {
     qCDebug(logSettings) << "Creating Settings file" << new_file_path;
-    QFile tpl(Settings::getSettingsTplPath());
-    QFile sfile(new_file_path);
+    QFile tpl_file(Settings::getSettingsTplPath());
+    QFile new_file(new_file_path);
 
-    if(!tpl.open(QIODevice::ReadOnly))
+    if(!tpl_file.open(QIODevice::ReadOnly))
     {
-        qWarning() << "Unable to read" << tpl.fileName() << "Check folder permissions.";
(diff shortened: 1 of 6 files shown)


Components/Settings.cpp
```diff
@@ -102,29 +106,30 @@ QString Settings::getTemplateDirPath()
...
-            qDebug() << "Unable to create" << Settings::getSettingsPath() << "Check folder permissions.";
-            return;
-        }
-
-        // QSettings setValue() methods delete all file comments, it's better to
-        // simply copy the template over to our destination directory.
-        QFile::copy(Settings::getSettingsTplPath(), Settings::getSettingsPath());
+    qCDebug(logSettings) << "Creating Settings file" << new_file_path;
+    QFile tpl(Settings::getSettingsTplPath());
+    QFile sfile(new_file_path);
```
broxen: I'll rename these variables to be more verbose

Components/Settings.cpp
```diff
@@ -136,13 +141,13 @@ void settingsDump()
 void settingsDump(QSettings *s)
 {
     QString output = "Settings File Dump\n";
-    foreach (const QString &group, s->childGroups()) {
+    foreach (const QString &group, s->childGroups())
```
nemerle: `for(const QString &group : s->childGroups())` plz

broxen:
Note, this still has some issues related to #840, specifically:
```
./dbtool adduser -l segsadmin -p segs15c00l!
WARNING! Unable to free statement: connection pointer is NULL
```
Database creation works great.

I'm going to go ahead and merge this, so we can get Blue/Red back up and running with MySQL, and then will focus on fixing PSQL with @ldilley in a new bug fix PR.

PR #865, opened 2019-04-13
Refinement to powers

cyanpill:
## Summary
More values in modifyAttrib(), so there shouldn't be any errors for missing values
Buff name/value sets now also contain duration, for powers that have multiple effects of different duration(ex cold and dark powers, rebirth inspiration)
refined datahelper's power functions, should be ready for delayed/repeat effects(next PR)
more power data loaded in powerserializer, damage and defense types saved as number for looking up on tables
types are applied to damage, but not resistances or defense, that will be a later PR


Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -1095,65 +1086,47 @@ void doAtrrib(Entity &ent, Entity *target_ent, StoredAttribMod const &mod, float
...
+            if (mod.StackType == AttribStackType::Replace)      //most are replace
             {
-                for (uint32_t i = 0; i<temp.m_buffs.size();i++)
+                for (buffset &bff :temp.m_buffs)
                 {
-                    if (temp.m_buffs[i].m_value_name == mod.name && mod.Attrib == temp.m_buffs[i].m_attrib)
+                    if (bff.m_value_name == srcbuff.m_value_name && srcbuff.m_attrib == bff.m_attrib)
                     {
-                        temp.m_duration = mod.Duration;      //if the same buff is there, refresh the duration and return
+                        bff.m_duration = srcbuff.m_duration;    //if the same buff is there, refresh the duration and return
```
ldilley: This seems a bit deeply nested, but can look at a potential restructure in a future PR.

ldilley:
Nice work as always, @cyanpill!

PR #869, opened 2019-04-13
Delayed/repeat power effects

cyanpill:
Added "delayedEffect", for power effects that start after a timer, and/or repeat
Added a vector of delayedEffects to each entity, which works with players and critters

I was testing logging out/zoning/dying while an effect was ticking down, and found that activation didn't properly check those as well. Oops! So I refactored checks for power range, to catch any null entity. Worked in every variation of player vs player vs critter I could think of.

Minor issue: the effects use the "character power" of the source to determine the strength of the effect, which isn't accessible without the entity. Ideally the strength would be calculated first. For now the effects end early, which was not as it was on live.

### DIFF
--- a/Projects/CoX/Common/GameData/Powers.h
+++ b/Projects/CoX/Common/GameData/Powers.h
@@ -338,11 +338,11 @@ static const int m_num_trays = 2; // was 3, displayed trays
 };
 struct DelayedEffect
 {
+    StoredAttribMod mod;
+    CharacterPower *power;
     int m_timer;
     int ticks;
     uint32_t src_ent;
-    StoredAttribMod mod;
-    CharacterPower *power;
 };
 
 /*
--- a/Projects/CoX/Common/GameData/Powers.h
+++ b/Projects/CoX/Common/GameData/Powers.h
@@ -336,7 +336,14 @@ static const int m_num_trays = 2; // was 3, displayed trays
     template<class Archive>
     void serialize(Archive &archive, uint32_t const version);
 };
-
+struct DelayedEffect
+{
+    int m_timer;
+    int ticks;
+    uint32_t src_ent;
+    StoredAttribMod mod;
+    CharacterPower *power;
+};
 
 /*
  * Powers Methods
--- a/Projects/CoX/Servers/MapServer/DataHelpers.cpp
+++ b/Projects/CoX/Servers/MapServer/DataHelpers.cpp
@@ -780,18 +780,7 @@ bool checkPowerRange(Entity &ent, Entity &target_ent, float range)
             return false;
     return true;
 }
-bool checkPowerRange(Entity &ent, uint32_t tgt_idx, uint32_t pset_idx, uint32_t pow_idx)
-{
-    CharacterPower * ppower =  getOwnedPowerByVecIdx(ent, pset_idx, pow_idx);
-    const Power_Data powtpl = ppower->getPowerTemplate();
-    if(powtpl.Range != 0.0f)
-    {
-        Entity *target_ent = getEntity(ent.m_client, tgt_idx);
-        if(glm::distance(ent.m_entity_data.m_pos,target_ent->m_entity_data.m_pos) > powtpl.Range)
-            return false;
-    }
-    return true;
-}
+
 void queuePower(Entity &ent, uint32_t pset_idx, uint32_t pow_idx, uint32_t tgt_idx)
 {
     CharacterPower * ppower =  getOwnedPowerByVecIdx(ent, pset_idx, pow_idx);
@@ -987,7 +976,14 @@ void findAttrib(Entity &ent, Entity *target_ent, CharacterPower * ppower)
             {
                 target_ent = &ent;
             }
-            doAtrrib(ent, target_ent, powtpl.pAttribMod[i], ppower);
+
+            if (powtpl.pAttribMod[i].Delay == 0 && powtpl.pAttribMod[i].Period == 0)
+                doAtrrib(ent, target_ent, powtpl.pAttribMod[i], ppower);
+            else
+            {           //queue the power up to fire later or multiple times
+                target_ent->m_delayed.push_back(
+                {powtpl.pAttribMod[i].Delay, powtpl.pAttribMod[i].Period,ent.m_idx, powtpl.pAttribMod[i], ppower});
+            }
         }
     }
 
--- a/Projects/CoX/Servers/MapServer/DataHelpers.cpp
+++ b/Projects/CoX/Servers/MapServer/DataHelpers.cpp
@@ -130,7 +130,23 @@ Entity * getEntity(MapInstance* mi, uint32_t idx)
     //sendInfoMessage(MessageChannel::USER_ERROR, errormsg, *src);
     return nullptr;
 }
+Entity * getEntity(Entity * srcEnt, MapInstance* mi, uint32_t idx)
+{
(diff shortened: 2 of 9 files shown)


Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -780,18 +780,7 @@ bool checkPowerRange(Entity &ent, Entity &target_ent, float range)
             return false;
     return true;
 }
-bool checkPowerRange(Entity &ent, uint32_t tgt_idx, uint32_t pset_idx, uint32_t pow_idx)
```
broxen: Was removing this intentional? Looks like we're still using it in `checkDelayedEffects()`

broxen: Nevermind, I see that this was overloaded.

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -88,10 +88,36 @@ void World::effectsStep(Entity *e,uint32_t msec)
...
+                dly_idx->ticks = -1;
+        }
 
+        if(dly_idx->ticks < 0)
+            e->m_delayed.erase(dly_idx);
+        else
+            ++dly_idx;
+    }
+}
 void World::checkPowerTimers(Entity *e, uint32_t msec)
```
broxen: This method has become a monster. In a future PR, can we refactor this into several smaller chunks?
```cpp
checkActivationTimers();
checkRechargeTimers();
checkAutoToggleTimers();
checkBuffTimers();
```
or something similar?

broxen: Maybe we break some of this out into another file as well to keep WorldSimulation.cpp lean and mean.

cyanpill: I agree, originally I had the delay stuff in powertimers, this is much cleaner.

broxen:
I pulled this down for some testing and ran into a segfault.

It appears to happen when you turn on auto-attack (right-click power) and target a Critter. While auto attacking continue to attack with other powers. With the delayed powers (charging) There is a small window where the Critter has died and is removed from the map, but the next auto attack fires anyway, resulting in:

```
[00:41:28] [log.slashcommand]Debug   : Sending 17 FloatingNumbers from 263 to 929
// charging text
[00:41:33] [log.slashcommand]Debug   : Sending 28 FloatingNumbers from 263 to 929
[00:41:33] Warning : "Entity 929 does not exist, or is not currently online."
```

cyanpill:
Ok, turns out auto-attack was not checking if the target was dead either. So I added a few checks to see if a target is dead, might be redundant in some cases, but it works.
I looked into fade timing(#872) and it turns out we are not actually setting that? So when critters die, I start the fade timer, which seems to take around 15 seconds. They are unable to be affected by most powers during that timer.
Also added a get entity that returns the source if nothing is found(#871). Changed over my checks to this new function.
And split up the checkpowertimers() as discussed.

broxen:
You're amazing Cyan 😍

Projects/CoX/Common/GameData/Powers.h
```diff
@@ -336,7 +336,14 @@ static const int m_num_trays = 2; // was 3, displayed trays
...
 };
-
+struct DelayedEffect
+{
+    int m_timer;
+    int ticks;
+    uint32_t src_ent;
+    StoredAttribMod mod;
+    CharacterPower *power;
+};
```
nemerle: I'd consider a slight reordering of the struct fields, 
```cpp
{
    StoredAttribMod mod;
    CharacterPower *power;
    int m_timer;
    int ticks;
    uint32_t src_ent;
};
```
Unless access patterns say otherwise :)

Projects/CoX/Servers/MapServer/DataHelpers.cpp
```diff
@@ -987,7 +976,14 @@ void findAttrib(Entity &ent, Entity *target_ent, CharacterPower * ppower)
             {
                 target_ent = &ent;
             }
-            doAtrrib(ent, target_ent, powtpl.pAttribMod[i], ppower);
+
+            if (powtpl.pAttribMod[i].Delay == 0 && powtpl.pAttribMod[i].Period == 0)
+                doAtrrib(ent, target_ent, powtpl.pAttribMod[i], ppower);
+            else
+            {           //queue the power up to fire later or multiple times
+                target_ent->m_delayed.push_back(
```
nemerle: `emplace_back` instead ?

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -2331,7 +2331,8 @@ void MapInstance::on_abort_queued_power(AbortQueuedPower * ev)
         return;
 
     // remove last queued power
-    session.m_ent->m_queued_powers.pop_back();
+    session.m_ent->m_queued_powers.end()->m_activation_state = false;
```
nemerle: You should use session.m_ent->m_queued_powers.back() here + check if there are any entries in `session.m_ent->m_queued_powers` at all

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -85,16 +85,41 @@ void World::effectsStep(Entity *e,uint32_t msec)
...
+                dly_idx->m_timer = int(dly_idx->mod.Duration*1000/dly_idx->mod.Period);
+                dly_idx->ticks--;
+            }
+            else
+                dly_idx->ticks = -1;
+        }
 
-void World::checkPowerTimers(Entity *e, uint32_t msec)
+        if(dly_idx->ticks < 0)
+            e->m_delayed.erase(dly_idx);
```
nemerle: depending on the underlying type of container, You'd want `dly_idx = e->m_delayed.erase(dly_idx);` here.

Projects/CoX/Servers/MapServer/WorldSimulation.cpp
```diff
@@ -270,7 +321,17 @@ void World::updateEntity(Entity *e, const ACE_Time_Value &dT)
 {
     physicsStep(e, uint32_t(dT.msec()));
     effectsStep(e, uint32_t(dT.msec()));
-    checkPowerTimers(e, uint32_t(dT.msec()));
+
+    // for now we only run these on players
+    if(e->m_type == EntType::PLAYER)
+    {
+        checkActivationTimers(e, uint32_t(dT.msec()));
```
nemerle: I'd refactor the dT.msec() calls to:
```cpp
const uint32_t dt_msec = uint32_t(dT.msec());
```
and then use dt_msec afterwards

cyanpill: how about using sim_frame_time? it's already set to delta.msec()/1000.0f, testing shows it works with the same timing as before. And it's public so we don't need to pass it... unless we are moving these functions out of worldsim.cpp

nemerle: One day I'd love to have ECS like architecture for things like this, so using parameter here might not make as much sense as sim_frame_time, but later on, it will be easier.

Quick note, take a look at the refactor that is coming https://github.com/Segs/Segs/blob/6002c76f78f6d93d72faf364005740a6835e4b16/Projects/CoX/Servers/MapServer/WorldSimulation.cpp#L743

Issue #756, opened 2019-02-06
Refactor SlashCommands for code hygiene


eos428:
Much like we talked about how MapInstance should be split, SlashCommands should also be refactored and split.

What to do:
- [X] Put the cmd.split regex thing into a single functions that will be used by most of these CMD functions
- [x] Split it up to different files. I forgot what's the convention again, but I'd like SlashCommands_Email, SlashCommands_Power and etc. Feel free to suggest better naming conventions though.
- [ ] Put functions aside from splitting up the ``cmd`` string and place them into DataHelpers and such, or the other way around - which is to remove the functions from DataHelpers and put everything inside the SlashCommands functions

broxen:
Thank you for filing this! 

Assigning to you and marking for v7. Feel free to un-assign yourself if you'd like to open it up to other contributors.

broxen:
Tagging #741 as they relate directly.

broxen:
As a part of this, it might be prudent to rewrite most of those commands to accept a target.

`setHP` `setEnd` `setTitles` etc would be useful for ServerOps to have for manipulating other entities (not just players) and so they really should accept a target. Commands without targets would assume `target = self`

eos428:
I forgot a bit about C++, but can you default a function param to null? If yes, then I wouldn't need to write overloads of the same functions.

Also, how do I get the targets of the commands? By parsing the cmd string or is the target stored in entity data?

broxen:
You can default function arguments.

For the target, I would take it from the command string, `/cmd $target args...` and `/cmd args...`

The entity stores it's target as an object, but for this case I think we'd want to take a name and look up target by name, or replace `$target` with the name of the players current target like we do for tells currently.

Thinking of it that way, I don't think you need to overload anything

cyanpill:
While we are changing slash commands, a feature mentioned on the wiki, you could add underscores at any place in commands. https://paragonwiki.com/wiki/Slash_Commands

broxen:
Interesting. Seems they just stripped them out.

broxen:
Regarding item three `remove the functions from DataHelpers and put everything inside the SlashCommands functions` originally the helper methods in DataHelpers were used in several places. This may have changed with recent restructuring.

Just something we should keep an eye out for, we may not be able to remove DataHelpers completely.

eos428:
[...]

Issue #741, opened 2019-01-23
Refactor string parsing out of each SlashCommand Handler


broxen:
## Feature Summary
We can improve readability and code clarity in SlashCommands.cpp by refactoring each handler to remove code duplication

One basic way in which this can be achieved is by handling command string parsing _before_ we pass the command to the command handler.

As an example, we often need each string argument from a slashcommand:
```cpp
void cmdHandler_SetCustomTitles(const QString &cmd, MapClientSession &sess)
{
    bool        prefix;
    QString     msg, generic, origin, special;
    QStringList args;
    args = cmd.split(QRegularExpression("\"?( |$)(?=(([^\"]*\"){2})*[^\"]*$)\"?")); // regex wizardry

    if(cmd.toLower() == "settitles")
    {
        setTitles(*sess.m_ent->m_char);
        msg = "Titles reset to nothing";
    }
    else
    {
        prefix  = !args.value(1).isEmpty();
        generic = args.value(2);
        origin  = args.value(3);
        special = args.value(4);
// so on...
```
This string parsing: `QStringList args = cmd.split(QRegularExpression("\"?( |$)(?=(([^\"]*\"){2})*[^\"]*$)\"?")); // regex wizardry` could be moved to `runCommand()` and the resulting QStringList passed to the SlashCommand Handlers to access directly.

Also this would standardize the way we parse strings for SlashCommands and would allow for complicated commands such as `/titles First Second "Long Third One"` to work across the board.

This change alone would reduce the file size by approximately 250 lines of code. I'm open to other ideas on how we can refactor SlashCommands.cpp

sfraint:
I can take this on

nemerle:
Nice, if You need any help, please drop by our discord :)

PR #882, opened 2019-04-25
Slash command refactor

sfraint:
## Summary
Refactor slash command arg parsing along with some other related cleanup.

### Issues closed
Closes https://github.com/Segs/Segs/issues/741
Related to/partly resolves https://github.com/Segs/Segs/issues/756

- Pass in list of params instead of raw command string
  - This causes a few changes to how we count/extract params passed into cmd handlers
  - Also allows us to use the same cmd parsing regex for all cmd handlers
  - Also no longer passing in the command itself e.g. `/dialog`
- Updated some of the log messages (though many are still not consistent across functions/duplicate info with `cmdlist` - seems like this could be done in a separate PR)
- Removed/replaced old helper functions
- Some misc clean up
  - Cleaned up some formatting
  - Fixed some type conversion warnings

### DIFF
--- a/Projects/CoX/Servers/MapServer/SlashCommand.cpp
+++ b/Projects/CoX/Servers/MapServer/SlashCommand.cpp
@@ -49,7 +49,7 @@ struct SlashCommand
 {
     QStringList m_valid_prefixes;
     QString m_help_text;
-    std::function<void(const QString &,MapClientSession &)> m_handler;
+    std::function<void(const QStringList &,MapClientSession &)> m_handler;
     uint32_t m_required_access_level;
 };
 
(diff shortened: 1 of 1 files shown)


Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -1087,69 +1045,67 @@ void cmdHandler_FaceEntity(const QString &cmd, MapClientSession &sess)
...
 
     bool ok = true;
-    uint32_t delay_status = args[0].toInt(&ok);
-    args.removeFirst(); // remove integer
-    QString msg = args.join(" ");
+    uint32_t delay_status = params.at(0).toUInt(&ok);
+
+    // Combine params after int and use those as message
+    QStringList params_copy;
+    for (QString s : params) {
```
broxen: Please put brackets on their own line per our coding style guide https://github.com/Segs/Segs/blob/develop/docs/CONTRIBUTING.md

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -954,19 +918,18 @@ void cmdHandler_SetClientState(const QString &cmd, MapClientSession &sess)
...
-    uint32_t v2 = args.value(2).toInt();
+    uint32_t v1 = params.value(0).toUInt();
+    uint32_t v2 = params.value(1).toUInt();
 
-    if(args.size() < 4)
-    {
-        qCDebug(logSlashCommand) << "Bad invocation:" << cmd;
-        sendInfoMessage(MessageChannel::USER_ERROR, "Bad invocation: " + cmd, sess);
+    if(params.size() < 2)
+    {;
```
broxen: Extraneous semi colon

Projects/CoX/Servers/MapServer/SlashCommand.cpp
broxen: Also we can simplify this `QString msg = params.removeFirst().join(" ");`

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -1087,69 +1045,67 @@ void cmdHandler_FaceEntity(const QString &cmd, MapClientSession &sess)
...
 
     sendDoorMessage(sess, DoorMessageStatus(delay_status), msg);
 }
 
-void cmdHandler_Browser(const QString &cmd, MapClientSession &sess)
+void cmdHandler_Browser(const QStringList &params, MapClientSession &sess)
 {
-    int space = cmd.indexOf(' ');
-    QString content = cmd.mid(space+1);
+    QString content = params.value(0);
```
broxen: We can probably simplify this by just passing params.value(0) directly to send browser()

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -1555,48 +1467,43 @@ void cmdHandler_WhoAll(const QString &/*cmd*/, MapClientSession &sess)
...
 }
 
-void cmdHandler_SetTitles(const QString &cmd, MapClientSession &sess)
+void cmdHandler_SetTitles(const QStringList &params, MapClientSession &sess)
 {
-    int space = cmd.indexOf(' ');
-    QString val = cmd.mid(space+1);
-
-    setTitle(sess, val);
+    QString title = params.value(0);
```
broxen: In cases like this where there is only one value passed, then perhaps we should just pass it directly to the helper method: `setTitle()` in this case.

But also, join any extra params because we can assume that any following strings in the list are a part of the same argument. For example:
`/afk "long message"`
and
`/afk long message`
would really be the same for slash commands like this.

Projects/CoX/Servers/MapServer/SlashCommand.cpp
sfraint: Fixed

Projects/CoX/Servers/MapServer/SlashCommand.cpp
sfraint: Removed

Projects/CoX/Servers/MapServer/SlashCommand.cpp
sfraint: Like we discussed, can't pass that directly into `sendBrowser` (and I don't think it is as simple as just changing `sendBrowser` to accept the `const` `QString`)

Projects/CoX/Servers/MapServer/SlashCommand.cpp
sfraint: Fixed.  I ended up doing this in all places that seemed possible (anywhere with only a single used param, that isn't immediately being converted to a number)

PR #885, opened 2019-05-01
Added powers.json to readme

cyanpill:
Moved to just /data/powers.json for simplicity
added to readme and motd to make sure people see it
Anywhere else?


broxen:
There's nowhere else I can think of. Invariably some people will miss it.

Perhaps we can convince @miles200 or someone to add a check or warning to segsadmin 😋

PR #891, opened 2019-05-06
No-coll Movement

spectre1989:
## Summary
No-collision movement on the server now matches the client.

The major change that made this work was how the server handles inputs from the client.
- Using the send_id in the input packet, the server is now able to figure out if any of the control state changes in the packet have been seen before.
- Inputs are then queued in the entity input state, and each tick they are processed. The server doesn't need to do any calculations about time, everything it needs is in the control state changes.

### Things I'm Unsure of:
- The way I've translated some of the RE movement code to server-side, it seems to work but I'm not sure how I'd fully test all code paths
- Compiler warnings, which can be ignored and which shouldn't
- When I tested with 2 clients with input and movement logging enabled, the sheer quantity of logs seemed to make my server fall over, if this is unacceptable then perhaps need a way to select which specific entity to log input/movement for?


broxen:
Regarding `When I tested with 2 clients with input and movement logging enabled, the sheer quantity of logs seemed to make my server fall over, if this is unacceptable then perhaps need a way to select which specific entity to log input/movement for?`

Perhaps a per-character variable `m_log_movement` to check before sending logs out. Not sure that's the cleanest way, but it would work.

spectre1989:
> Regarding `When I tested with 2 clients with input and movement logging enabled, the sheer quantity of logs seemed to make my server fall over, if this is unacceptable then perhaps need a way to select which specific entity to log input/movement for?`
> 
> Perhaps a per-character variable `m_log_movement` to check before sending logs out. Not sure that's the cleanest way, but it would work.

And enable it per-character with a /command?

broxen:
> And enable it per-character with a /command?

Precisely `/logmovement {optional: $target}` or something. Variable may be best in the Entity class, as different types of entities may have movement (Monorail, Blimp, Pedestrians, etc)

Projects/CoX/Common/GameData/StateStorage.h
```diff
@@ -40,114 +104,92 @@ class TimeState
...
+    // out if these control state changes are new or not.
+    uint16_t    m_first_control_state_change_id = 0xffff;
+
+    // whether this state change contains the keys currently held in m_keys,
+    // doesn't necessarily mean any of them have changed, but useful for 
+    // validating against current key state on server
+    bool        m_has_keys = false;
+    bool        m_keys[6] = {};
+
+    // whether this state change contains the client pitch/yaw which are sent
```
broxen, line 142: These may actually be camera pitch/yaw, and perhaps the angle deviates due to some kind of rounding error with mouse movement? Worth testing to see.

Projects/CoX/Common/GameData/StateStorage.h
```diff
@@ -40,114 +104,92 @@ class TimeState
...
-    {
-        if(m_inp_states.size() < 2)
-            return &m_inp_states.back();
+    // how long each key has been held
+    uint32_t                        m_key_press_duration_ms[6] = {};
 
-        return &m_inp_states.back()-1;
-    }
+    // sent by client, not sure what it's for
+    float                           m_velocity_scale = 1.0f;
```
broxen, line 191: I believe this should increase the speed of all movement if the client has changed it's value.

Projects/CoX/Common/GameData/Movement.cpp
```diff
@@ -49,195 +58,571 @@ SurfaceParams g_world_surf_params[2] = {
...
+        qCDebug(logMovement, "\nAdding: %s += \t%dms (%dms total)\n",
+                s_key_name[key],
+                milliseconds,
+                player->m_input_state.m_key_press_duration_ms[key]);
+    }
+}
 
-        if(vel.z < 0.0f)
-            vel_scale_copy *= motion->m_backup_spd;
+static void playerMotionSetInputVelocity(Entity* player, const TickState* tick_state) // based on pmotionSetVel()
```
broxen: ~~The var name `player` is misleading here. A player is always represented by an Entity, but an Entity may or may not be a Player.~~

Disregard my comment. This may not matter if we never use this method for other entity types (Critters, Blimps, Cars, etc). Input was the key word here and I missed it =P

spectre1989: Yeah I'm making an assumption that this code only runs for players, if that's not the case it'll need changing

spectre1989:
> > And enable it per-character with a /command?
> 
> Precisely `/logmovement {optional: $target}` or something. Variable may be best in the Entity class, as different types of entities may have movement (Monorail, Blimp, Pedestrians, etc)

ok, I've added it, but can't test it yet because of the db issues we're having at present

broxen:
Give this is another go pulling the latest from develop head.

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -257,6 +259,8 @@ static const SlashCommand g_defined_slash_commands[] = {
     {{"clientConsoleOutput"}, "Send message to ingame (~) console", cmdHandler_SendConsolePrint, 9},
     {{"clearTarget"}, "Clear current target", cmdHandler_ClearTarget, 9},
     {{"startTimer"}, "Create a small timer", cmdHandler_StartTimer, 9},
+    {{"toggleInputLog"}, "Enable input logging for a player", cmdHandler_ToggleInputLog, 9},
```
broxen: Smart idea.

Projects/CoX/Servers/MapServer/SlashCommand.cpp
```diff
@@ -1306,6 +1310,36 @@ void cmdHandler_StartTimer(const QStringList &params, MapClientSession &sess)
     sendMissionObjectiveTimer(sess, message, time);
 }
 
+void cmdHandler_ToggleInputLog(const QStringList &params, MapClientSession &sess)
+{
+    Entity* target = getTargetEntity(sess);
```
broxen: Following the format of `/teleport` may allow admins to type in a name of a player that they can't target directly (eventually, using UserRouter, another zone etc).

It makes several checks for name and target to avoid duplicate name issues where possible.

```cpp
    QString name = sess.m_ent->name();
    if(!params.isEmpty())
        name = params.join(" ");

    // getTargetByNameOrIdx will always return a valid Entity, or self
    Entity *tgt = getCmdTargetByNameOrIdx(sess, name);
    name = tgt->name(); // make sure we have the final name
```

spectre1989:
> Give this is another go pulling the latest from develop head.

Yep that worked, I'll finish up the logging stuff tomorrow

Projects/CoX/Servers/MapServer/SlashCommand.cpp
spectre1989: I've updated it to use that with the latest commit, and tested with 2 clients, switching logging between myself and another player, seems to work.

Projects/CoX/Common/GameData/StateStorage.h
```diff
@@ -26,6 +26,70 @@ enum BinaryControl
...
+struct ControlStateChange
+{
+    uint8_t control_id = 0xff;
+    uint32_t time_since_prev_ms = 0;
+
+    union
+    {
+      uint8_t key_state;            // control_id 0-5
+      float angle;                  // control_id 6-7
+      struct
```
broxen, line 38: It probably doesn't really matter, but `-wPedantic` is throwing roughly a dozen warnings about Anonymous structs here. Let's name this.

`warning: ISO C++ prohibits anonymous structs [-Wpedantic]`

Projects/CoX/Common/GameData/Movement.cpp
```diff
@@ -49,195 +58,601 @@ SurfaceParams g_world_surf_params[2] = {
     { 1.50f, 3.00f, 0.00f, 3.00f, 3.00f },      // air; test values
 };
 
-void roundVelocityToZero(glm::vec3 *vel)
+static bool floatEquals(float a, float b)
```
broxen: We don't actually use this anywhere. Is there a future use planned, or should we just remove it?

Projects/CoX/Common/GameData/Movement.cpp
```diff
@@ -49,195 +58,601 @@ SurfaceParams g_world_surf_params[2] = {
...
+                            e.m_motion_state.m_no_collision = csc.no_collision;
+                        break;
+
+                        case 8:
+                            if (logInput().isDebugEnabled() && e.m_input_state.m_debug)
+                            {
+                                qCDebug(logInput, "doing tick, controls_disabled=%d, time_diff_1=%u, time_diff_2=%u, velocity_scale=%hhu", csc.controls_disabled, csc.time_diff_1, csc.time_diff_2, csc.velocity_scale);
+                            }
+
+                            // do a tick!
```
broxen: Errant comment? or TODO?

Projects/CoX/Common/GameData/StateStorage.cpp
```diff
@@ -60,34 +60,28 @@ void TimeState::dump()
...
     {
-        if(!new_state.m_pyr_valid[i])
-            new_state.m_camera_pyr[i] = m_inp_states.back().m_camera_pyr[i];
-
-        if(new_state.m_orientation_pyr[i] != m_inp_states.back().m_orientation_pyr[i])
-            update_needed = true;
+        for (bool key : m_keys)
+        {
+            if (key)
+            {
```
broxen: For code consistency, let's omit brackets on single line code blocks, except for where nested. Referencing our coding guidelines:
```cpp
for (bool key : m_keys)
{
    if (key)
        return true;
}

**Review by broxen — CHANGES_REQUESTED** [review]
Thank you for working on this spectre!

I'm running into a few issues after building this PR
- server position doesn't actually appear to be updated, I'm not sure if I need to toggle server authority, but `/movement` didn't seem to have any effect.
- using powers sometimes disables movement altogether. In particular "Sprint"
- turning with the mouse or keyboard `q` or `e` has significant latency. It takes the client several seconds to update the current player's rotation.
- Other clients do see player rotation, but with the same delay as seen locally.
- Other clients don't see movement at all.

File: Projects/CoX/Common/GameData/Movement.cpp (hunk 8 above)
Comment (spectre1989) [reply Projects/CoX/Common/GameData/Movement.cpp]: It's a comment - control id 8 is "end of tick" which means a tick can be executed. Should I change this comment to something else?

File: Projects/CoX/Common/GameData/StateStorage.cpp (hunk 9 above)
Comment (spectre1989) [reply Projects/CoX/Common/GameData/StateStorage.cpp]: MOTHER OF GOD

File: Projects/CoX/Common/GameData/Movement.cpp (hunk 7 above)
Comment (spectre1989) [reply Projects/CoX/Common/GameData/Movement.cpp]: Oops, just left over from when I was using it, and then I wasn't...

File: Projects/CoX/Common/GameData/StateStorage.h (hunk 6 above)
Comment (spectre1989) [reply Projects/CoX/Common/GameData/StateStorage.h:38]: Ah yeah, I meant to do that, will do

**Comment by spectre1989** [comment]
> Thank you for working on this spectre!
> 
> I'm running into a few issues after building this PR
> 
> * server position doesn't actually appear to be updated, I'm not sure if I need to toggle server authority, but `/movement` didn't seem to have any effect.
> * using powers sometimes disables movement altogether. In particular "Sprint"
> * turning with the mouse or keyboard `q` or `e` has significant latency. It takes the client several seconds to update the current player's rotation.
> * Other clients do see player rotation, but with the same delay as seen locally.
> * Other clients don't see movement at all.

I can't repro those issues locally, are you running the server on your local machine or a remote one? I've only tried running the server on the same machine as the client.

Server authority is likely still broken, I couldn't get it to work before and I've not looked at fixing it yet. I was thinking perhaps (once we're happy with it) we should turn on server authority by default for no-coll movement only, but obviously as all other movement is unimplemented leave it off for that.

**Comment by spectre1989** [comment]
I'm still working on an update for this - hope it's alright with you @broxen, I'm integrating stuff from your movement branch as you've got the whole structure and flow of entMotion/entWalk/etc.

File: Projects/CoX/Common/GameData/Movement.cpp (hunk 8 above)
Comment (broxen) [reply Projects/CoX/Common/GameData/Movement.cpp]: I would make this more verbose so it's clear that the behavior is already there, and this isn't a TODO for future code or something.

**Comment by broxen** [comment]
Absolutely, feel free to use that code. Tread carefully though, I can't remember what I did there =P

**Comment by spectre1989** [comment]
Ok as discussed I've integrated a lot of the code from your movement branch, but I've left out entWorldCollide. For now I'm spoofing player movement so that it roughly matches player movement in the starting area that I was testing with. I've also renamed a lot of the functions to be consistent with yours, I figure that'll make it less confusing to work with going forward if I stop renaming things!

**Comment by broxen** [comment]
CI is failing due to the use of compare in the body parts changes Nem made recently. I'll try to push a fix for that tonight.

```cpp
/home/travis/build/Segs/Segs/Projects/CoX/Common/GameData/bodypart_definitions.cpp: In member function ‘BodyPart_Data* BodyPartsStorage::getBodyPartFromName(const QByteArray&)’:
/home/travis/build/Segs/Segs/Projects/CoX/Common/GameData/bodypart_definitions.cpp:120:27: error: ‘class QByteArray’ has no member named ‘compare’
         if ( 0==bp.m_Name.compare(name,Qt::CaseInsensitive) )
                           ^~~~~~~
```

nemerle:
That problem with CI is likely caused by the older Qt version on the server.

cajoebestrotheray:
> CI is failing due to the use of compare in the body parts changes Nem made recently. I'll try to push a fix for that tonight.
> 
> ```c++
> /home/travis/build/Segs/Segs/Projects/CoX/Common/GameData/bodypart_definitions.cpp: In member function ‘BodyPart_Data* BodyPartsStorage::getBodyPartFromName(const QByteArray&)’:
> /home/travis/build/Segs/Segs/Projects/CoX/Common/GameData/bodypart_definitions.cpp:120:27: error: ‘class QByteArray’ has no member named ‘compare’
>          if ( 0==bp.m_Name.compare(name,Qt::CaseInsensitive) )
>                            ^~~~~~~
> ```

Is there anything I need to do for this? Like merge upstream and push to my branch or something?

broxen:
We might be able to adjust CI to to use a newer QT version. Also Nem mentioned in discord a possible cmake change that would look for the latest version.

I've been pretty slammed with work and haven't had a chance to look at it, but those are two possible solutions

broxen:
This is an enormous step in the right direction. Fantastic work spectre!

PR #898, opened 2019-06-03
Find the latest installed qt version on windows machines.

nemerle:
Should solve issues with appveyor


broxen:
This fixed Appveyor, but did not fix the QByteArray issue with Travis, which I assume is just some kind of linkage order error. QByteArray is clearly defined, but compare isn't recognized as a member.

PR #912, opened 2020-03-13
Do not Merge!! Remove debug postfix from ACE build for MSVC

miles200:
## Summary
When building with MSVC, segs_server failed to start due to missing ACE.dll in the builds 'out' directory. `3rd_party/CMakeLists.txt` was not moving the dll due to the copy being inside the if clause `if(MINGW)` and the compiler used being MSVC. 

- Included MSVC in command to copy dll within `3rd_party/CMakeLists.txt`
- Removed debug postfix from ACE source CMakeLists for MSVC 
- Removed reference to debug location for MSVC within `ExternalProject_ACE.cmake`


miles200:
Replaced by #918

PR #919, opened 2020-03-16
Can we automate builds with GitHub Actions

broxen:
Seems like we can use these fancy new actions to automate a package build whenever we push something to Master branch.

## Summary
First stab at GitHub Actions to automate our package process

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -1,0 +1,33 @@
+name: C/C++ CI
+
+on: [push]
+
+jobs:
+  build-ubuntu:
+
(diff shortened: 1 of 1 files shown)


nemerle:
I wonder, what would we gain by switching over to github CI? I think both appveyor/travis support building/uploading packages as well, github CI would be slightly easier true. But we'd lose macos builds ?

broxen:
Ah I didn't realize that we could build out packages for release with Appveyor and Travis.

My thought was automating a sort of nightly release or master release.

broxen:
Closing as this is duplicitous

PR #924, opened 2020-03-23
SlashCommand refactoring

broxen:
## Summary
SlashCommand.cpp is large and unwieldy, creating conflicts with every merge. This refactors it into multiple files to minimize those issues.

### Issues closed
Closes #756 

### Additions/modifications proposed in this pull request:
- Breaks SlashCommands out into several manageable files
- Refactors some getEntity helper methods

Issue #690, opened 2018-11-08
Stagger Entity updates and expand StateBelief


broxen:
## Summary
Per discussion with Nem on Discord, we need to be more tactical with our entity updates per Client to conserve resources.
> nemerle: also, we should start considering 'staggering' entity updates, so we don't send all entities at once, but a minimal set of the closest entities to the player, and 'extend' the list to include 'entities-in-potential-view' right now we send every entity with every update even the incremental updates are not free :smiley:

> nemerle: Ya, that will require expanding  our 'StateBelief'  concepts a bit and funnily enough, finding entities in a neighborhood of a specific Entity

> broxen: Is scenegraph aware of neighborhoods now?

> nemerle: actually, writing queries about nearest entities should not require the scene-graph, unless we want to do fancy occlusions :smiley:
> nemerle: We could use some nice kd-tree, or even a simple grid-hash based  acceleration structure to keep those queries fast

broxen:
This is pretty important, going to flag as `release blocker` for v0.8

broxen:
Per discussion with psiLupan in Discord, the following information is relevant to this conversation

------

Pinging you since you've mostly been the one doing SG stuff but line 374 of EntityUpdateCodec is sent if "update" is true, otherwise it sends 0. 

So it doesn't necessarily mean that entity has a Supergroup.

Also, line 381 about the hash table key is the Emblem as suspected in the comment
It seems like if an entity doesn't have an SG/SG_ID if update is true, then it also sends a 2 bit packet of 0 in place of SG_ID and returns.
So rather than m_has_supergroup, it seems like just having m_has_update or something would work. Then if update is true, the rest of the code works as is
This would presumably fix a condition where if someone has an SG and leaves, they're probably not updating that player state for those around them that I'm guessing is happening.

In engine, they simply had the caller function set whether or not update was true, rather than it being tied to the entity.
So, it may be better to break it out into m_has_sgupdate or something of that nature.

Also, regarding https://github.com/Segs/Segs/blob/develop/Projects/CoX/Servers/MapServer/EntityUpdateCodec.cpp#L282

have_stats isn't a result of the player being dead. It's essentially another "has update" situation. 

The below is based on a code comment directly from the client receiving side and a slightly different one on the server side, which I've rewritten to clarify it.
```
The following sets have_stats to false.
The entity must have an associated player character and it cannot be a full_update
The given entity is the receiving player OR the entity's stats haven't changed since the last update
[...]

PR #925, opened 2020-03-24
Entity Update Flags

broxen:
## Summary
I took a stab at allowing Entity to Update the client only as needed. I did this by creating a integer bit set that can be toggled with flags for the items in EntityUpdateCodec's `serializeto()`.

This PR is built upon #924 SlashCommand Refactoring PR, which should merge first.

### Issues closed
None. But this is a step towards #690 

- EntityUpdateFlags for tracking when updates are necessary to send to the client. This differs from  ClientEntityStateBelief which is the client's understanding of our entity state.
- EntityUpdateFlag helper methods to mark, unmark, and reset the flags for an entity. 
- Lots of minor code cleanups. 
- Removed `GameCommandsList.h` per my note in Discord.


broxen:
This fails on OSX because... ???
```cpp
/Users/travis/build/Segs/Segs/Projects/CoX/Common/GameData/EntityHelpers.cpp:457:51: error: 'Entity::UpdateFlag::Full' is not a member of class 'Entity'
    e.m_entity_update_flags.setFlag(e.UpdateFlag::Full);
```
It seems it's not recognizing the enum, and that makes no sense to me =P

broxen:
Seems as though the issue is _very technical_ 
https://stackoverflow.com/questions/50027942/why-doesnt-clang-allow-accessing-a-nested-enum-class-through-an-instance

For now I've made it a classic enum and capitalized all the members to avoid name collisions.

Projects/CoX/Common/GameData/Movement.cpp
```diff
@@ -934,27 +936,27 @@ void forcePosition(Entity &e, glm::vec3 pos)
...
-    e.m_update_anims = true;
-    e.m_rare_update = true;
-
     TriggeredMove tmove;
     tmove.m_move_idx = move_idx;
     tmove.m_ticks_to_delay = delay;
     tmove.m_trigger_fx_idx = fx_idx;
 
     e.m_triggered_moves.push_back(tmove);
+    e.m_entity_update_flags.setFlag(e.UpdateFlag::ANIMATIONS);
```
nemerle, line 959: To be honest, this looks a bit strange..
usually when referring to a nested enum field you use the full path:
```cpp
ClassName::Enum::Field
```
My advice: 
extract the enum outside of the class, and name it `EntityUpdateFlag`

broxen: Sounds good.

PR #929, opened 2020-04-18
Separate Email-related functions in MapInstance to EmailService

eos428:
Reviving a year-old PR, but this time the changes are going to go bit by bit.

Some things I feel a bit off, and maybe others could chime to improve:
- Should make a variant of ServiceToClientData for MapLinkEvent that takes in entity (unless we can get the token from a MapLinkEvent)
- I'm not sure what to feel about these constructors in ServiceToClientData. Perhaps we should make a static helper class to help in constructing them?

A year ago, when it was just handling emails it did work, but now whether it works or not is still up in the air. But this should be far easier to troubleshoot rather than a single PR that does everything.


Projects/CoX/Common/Servers/InternalEvents.h
```diff
@@ -245,4 +249,33 @@ struct MapSwapCollisionData
...
+{
+    uint64_t m_token = 0;
+    GameCommandVector m_commands;
+    QString m_message;
+    MessageChannel m_message_channel;
+
+    ServiceToClientData(){};
+
+    // use this if you are sending GameCommands and find your session from token (eg EmailService)
+    ServiceToClientData(uint64_t token, GameCommandVector commands, QString msg = nullptr, MessageChannel messageChannel = MessageChannel::DEBUG_INFO)
```
nemerle: `QString msg=nullptr` does not make much sense, try either (`QString msg={}` or `QString *msg=nullptr` )
Could also consider passing commands as r-value reference ( `GameCommandVector &&commands`)
Bonus: Try laying out the structure in field size order ( can reduce the size of the structure )

Projects/CoX/Servers/GameServer/EmailService/EmailService.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+std::unique_ptr<ServiceToClientData> EmailService::on_email_header_response(Event* ev)
+{
+    EmailHeaderResponse* casted_ev = static_cast<EmailHeaderResponse *>(ev);
+
+    std::vector<EmailHeaders::EmailHeader> emailHeadersVector;
+    emailHeadersVector.reserve(casted_ev->m_data.m_email_headers.size());
+    for (const auto &email_header : casted_ev->m_data.m_email_headers)
+    {
+        emailHeadersVector.push_back(EmailHeaders::EmailHeader{email_header.m_email_id,
```
nemerle: you can use `emplace_back` here for a bit of performance gain.

Projects/CoX/Servers/GameServer/EmailService/EmailService.cpp
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// EmailHandler will send this event here
+std::unique_ptr<ServiceToClientData> EmailService::on_email_header_to_client(Event* ev)
+{
+    EmailHeaderToClientMessage* casted_ev = static_cast<EmailHeaderToClientMessage *>(ev);
+    GameCommandVector commands;
+    commands.push_back(std::make_unique<EmailHeaders>( casted_ev->m_data.m_email_id, casted_ev->m_data.m_sender_name,
+                                                       casted_ev->m_data.m_subject, casted_ev->m_data.m_timestamp));
+
```
nemerle, line 46: `emplace_back` can be used above.

Projects/CoX/Servers/GameServer/EmailService/EmailService.cpp
```diff
@@ -0,0 +1,95 @@
...
+    EmailHeadersToClientMessage* casted_ev = static_cast<EmailHeadersToClientMessage *>(ev);
+
+    std::vector<EmailHeaders::EmailHeader> emailHeadersVector;
+    emailHeadersVector.reserve(casted_ev->m_data.m_email_headers.size());
+    for (const auto &email_header : casted_ev->m_data.m_email_headers)
+    {
+        emailHeadersVector.push_back(EmailHeaders::EmailHeader{email_header.m_email_id,
+                                                          email_header.m_sender_name,
+                                                          email_header.m_subject,
+                                                          email_header.m_timestamp});
```
nemerle, line 61: `emplace_back` here as well.

Projects/CoX/Servers/GameServer/EmailService/EmailService.h
```diff
@@ -0,0 +1,26 @@
...
+private:
+    using GameCommandVector = std::vector<std::unique_ptr<SEGSEvents::GameCommandEvent>>;
+public:
+    std::unique_ptr<SEGSEvents::ServiceToClientData> on_email_header_response(SEGSEvents::Event* ev);
+    std::unique_ptr<SEGSEvents::ServiceToClientData> on_email_headers_to_client(SEGSEvents::Event* ev);
+    std::unique_ptr<SEGSEvents::ServiceToClientData> on_email_header_to_client(SEGSEvents::Event* ev);
+    std::unique_ptr<SEGSEvents::ServiceToClientData> on_email_read_response(SEGSEvents::Event* ev);
+    std::unique_ptr<SEGSEvents::ServiceToClientData> on_email_create_status(SEGSEvents::Event* ev);
+    std::unique_ptr<SEGSEvents::ServiceToClientData> on_email_read_by_recipient(SEGSEvents::Event* ev);
+
```
nemerle, line 24: Should we consider adding a type alias for `std::unique_ptr<SEGSEvents::ServiceToClientData>` in Projects/CoX/Common/Servers/InternalEvents.h ?
```cpp
using UPtrServiceToClientData = std::unique_ptr<SEGSEvents::ServiceToClientData>;
```
could increase readability a bit ?

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -116,7 +116,15 @@ MapInstance::MapInstance(const QString &mapdir_path, const ListenAndLocationAddr
     m_scripting_interface->setIncludeDir(mapdir_path);
     m_endpoint = new MapLinkEndpoint(m_addresses.m_listen_addr); //,this
     m_endpoint->set_downstream(this);
+
+    initServices();
+}
+
+void MapInstance::initServices()
+{
+    m_email_service = EmailService();
```
nemerle: a simpler initializer syntax can be used here:
```cpp
m_email_service = {};
```

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -504,22 +512,22 @@ void MapInstance::dispatch( Event *ev )
             on_reset_keybinds(static_cast<ResetKeybinds *>(ev));
             break;
         case evEmailHeaderResponse:
-            on_email_header_response(static_cast<EmailHeaderResponse *>(ev));
+            on_service_to_client_response(m_email_service.on_email_header_response(ev));
             break;
```
nemerle, line 516: This looks good, later we might be able to update all case branches here to this layout:
```
std::unique_ptr<SEGSEvents::ServiceToClientData> resp;
////////
case evXYZ:
  resp = std::move(m_service_name.method(ev));
  break;
////////
on_service_to_client_response(std::move(resp));
```

Projects/CoX/Servers/MapServer/MapInstance.cpp
```diff
@@ -3309,5 +3240,29 @@ void MapInstance::clearLuaTimer(uint32_t entity_idx)
...
+        MapClientSession& session = m_session_store.session_from_token(data->m_token);
+
+        for (auto &command : data->m_commands)
+            session.addCommandToSendNextUpdate(std::move(command));
+
+        // is not null and is not empty
+        if (!data->m_message.isEmpty() && !data->m_message.isNull())
+            sendInfoMessage(MessageChannel::DEBUG_INFO, data->m_message, session);
+    }
+    catch(std::exception &e)
```
nemerle: Hmm... are we throwing exceptions somewhere?

eos428: IIRC there were cases where you won't get the required session. And in the previous PR it used to find the map server too (for MapLinkEvents). This time, I do intend to separate the events based on SEGSEvents and MapLinkEvents.

nemerle: Ahh, are we trying to access non-existent data then? 
That's a bug, so we should log it instead and try to continue with processing ;)

I'm not a huge fan of exceptions :)

eos428: I see. Since the session is a reference, I'll assume null checks and such are done in `session_from_token`?

Projects/CoX/Common/Servers/InternalEvents.h
```diff
@@ -253,15 +253,15 @@ using GameCommandVector = std::vector<std::unique_ptr<GameCommandEvent>>;     //
 
 struct ServiceToClientData
 {
-    uint64_t m_token = 0;
+    uint64_t m_token : 8;
```
nemerle: Hmm, interesting, is m_token always <=255 ?
As for structure's layout, I'd propose :
```cpp
    GameCommandVector m_commands; // it's ~2*sizeof(void*) bytes in length
    QString m_message; // it's ~1*sizeof(void*) bytes in length
    uint64_t m_token;
    MessageChannel m_message_channel;
```
    
```

Comment (eos428) [reply Projects/CoX/Common/Servers/InternalEvents.h]: For token numbers I would say... no, just practicing the first search result of bit fields haha. I think since tokens are based on sessions, it depends on how populated the server is at one time.

Knowing that the order of variables determine optimization is new to me. Would you explaining this more in-depth? Thanks in advance :D

Comment (nemerle) [reply Projects/CoX/Common/Servers/InternalEvents.h]: Sure thing.
So, the compiler has a specific way of laying out structure fields in memory, which in many cases causes the actual structure size to be larger than the sum of the member sizes:
```cpp
struct A {
    bool a1;
    int b1;
    bool a2;
};
struct A2 {
    int b1;
    bool a1;
    bool a2;
};
```
In the above case, for 64 bit compile, we have `sizeof(A)==12` and `sizeof(A2)==8`
So proper structure layout will reduce the struct size. Thus increasing the number of structures that fit in a cache-line.
 Additional optimizations come into play when we take into account the frequency of field access. Members that are accessed often ( inner loops etc. ) should be located closer to the structure's start, this is especially important for structures larger than 64 bytes ( cache-line).

Both of those optimizations are most useful for structures that are held in continuous memory.

eos428: I see... Thanks for the insight nem. And after reading up on optimization based on member order, so I guess these are the factors to take into account when ordering them?
- Byte count
- How often they'd be used
- And group the same variable types together?

On another note, I think the code should be alright, pending testing?

nemerle: Yup, there are some deeper magics in play there as well ( using pragma(pack), volatile/atomic variables, forced alignment, etc. )

And yeah, I think we can merge unless You want to add anything?

eos428: Just tested the code at runtime. Looks fine and should be ready to merge

PR #930, opened 2020-05-01
feat: create ClientOptionService as part of map instance refactor

eos428:
New stuff I made:
- EventHelpers to help get the session token from an event. I don't want to use m_session_store in the services so the closest compromise is to get the tokens and let MapInstance find the sessions directly from the token instead
- ClientOptionService as the PR suggests
- ServiceToEntityData which is different from the already-used ServiceToClientData of course

Some points for discussion:
- In MapInstance's service response functions, the method to get the sessions are identical, maybe we could create a new function for it instead?
- What else we could/should put on ClientOptionService?

Also, I would like some help with troubleshooting. I tried to build my PR but I encountered this error. Was there something missing from the CMakeLists?
Annotation 2020-05-01 212556


nemerle:
It looks like you're adding a new message handler library ( `add_handler` )
I'm not seeing corresponding `add_messages_library` anywhere, though.
You can follow the way it's done in any of `CoX/Common/Messages/*Service` directories

PR #942, opened 2021-10-07
Mysql connection

dracc:
## Summary
Fixes some issues found while trying to use SEGS with a MySQL server.

### Issues closed
-

- dbtool: Fix MySQL connection check, old method required the database to be populated - not great when you want to create a new DB.
- dbtool: Let the MySQL server take care of timestamps, they're already set to be generated DB-side on entry creation.

### DIFF
--- a/Utilities/dbtool/DBConnection_AddUser.cpp
+++ b/Utilities/dbtool/DBConnection_AddUser.cpp
@@ -32,7 +32,7 @@ DBToolResult DBConnection::addAccount(const QString &username, const QString &pa
 {
     qInfo().noquote() << "Adding new account to" << getName() << "database";
 
-    if(!m_query->prepare("INSERT INTO accounts (username,passw,access_level,salt,creation_date) VALUES (?,?,?,?,?);"))
+    if(!m_query->prepare("INSERT INTO accounts (username,passw,access_level,salt) VALUES (?,?,?,?);"))
     {
         qDebug() << "SQL_ERROR:" << m_query->lastError();
         return DBToolResult::QUERY_PREP_FAILED;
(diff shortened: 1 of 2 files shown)


Utilities/dbtool/DBConnection_AddUser.cpp
```diff
@@ -32,7 +32,7 @@ DBToolResult DBConnection::addAccount(const QString &username, const QString &pa
 {
     qInfo().noquote() << "Adding new account to" << getName() << "database";
 
-    if(!m_query->prepare("INSERT INTO accounts (username,passw,access_level,salt,creation_date) VALUES (?,?,?,?,?);"))
+    if(!m_query->prepare("INSERT INTO accounts (username,passw,access_level,salt) VALUES (?,?,?,?);"))
```
nemerle, line 35: I wonder, will this break sqlite3 (  I don't remember if it supports `now` )

Seems it should `DATETIME DEFAULT CURRENT_TIMESTAMP`

Utilities/dbtool/DBConnection_Helpers.cpp
```diff
@@ -36,7 +36,10 @@ bool DBConnection::isConnected()
     }
     else if(m_config.isMysql())
     {
-        querytext = "SHOW TABLES";
+      if(m_db->isOpen())
+        return true;
+
+      return false;
```
nemerle: nitpick: simple `return m_db->isOpen();`  will work here 😄

dracc: Thanks, not sure what I was thinking. 😅

PR #964, opened 2022-09-06
SEGSAdmin Theming - Initial Test

miles200:
## Summary
Added Dark and Light custom themes to SEGSAdmin using Qt Style Sheets to try to eliminate theming issues on MacOS where system themes were being applied.

### DIFF
--- a/Utilities/SEGSAdmin/SEGSAdminTool.h
+++ b/Utilities/SEGSAdmin/SEGSAdminTool.h
@@ -9,13 +9,11 @@
 #define SEGSADMINTOOL_H
 
 #include "Version.h"
-#include "ThemeManager.h"
 
 #include <QMainWindow>
 #include <QDialog>
 #include <QProcess>
 #include <QFontDatabase>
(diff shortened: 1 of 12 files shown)


Utilities/SEGSAdmin/SEGSAdminTool.h
```diff
@@ -9,10 +9,13 @@
 #define SEGSADMINTOOL_H
 
 #include "Version.h"
+#include "ThemeManager.h"
+
 #include <QMainWindow>
 #include <QDialog>
 #include <QProcess>
 #include <QFontDatabase>
+#include <QSignalMapper>
```
nemerle: Small nitpick - I'd replace the `#include "ThemeManage.h"` with a forward declaration, and remove QSignalMapper include from here.

miles200: Thanks, signal mapper was left in error so removed. Have forward declared ThemeManager and switch include guards to use pragma once

PR #970, opened 2022-09-17
SEGSAdmin Updates

### LINKED ISSUE #969: DB names not appearing in SEGSAdmin as a setting
Add:

AccountDatabase\db_name
CharacterDatabase\db_name

- Change logic, if QSQLITE then default extension to .db, else leave extension

miles200:
## Summary
- Moved dark/ light theme toggle into the toolbar
- Added database name to settings for both char and account db's 
- If either char or account db's are not using QSQLITE, disabled the auto connectivity detection and database creation functions until a more permanent solution is in dbtool
- If QSQLITE is selected, SEGSAdmin will automatically add the .db extension as required 
- If QMYSQL or QPSQL is selected, SEGSAdmin will remove the .db extension when saving
- General layout bug fixes in Settings

### Issues closed
Closes #969

### DIFF
--- a/Utilities/SEGSAdmin/SEGSAdminTool.cpp
+++ b/Utilities/SEGSAdmin/SEGSAdminTool.cpp
@@ -72,7 +72,7 @@ SEGSAdminTool::SEGSAdminTool(QWidget *parent) :
     connect(this,&SEGSAdminTool::readyToRead,m_settings_dialog,&SettingsDialog::read_config_file);
     connect(this,&SEGSAdminTool::checkConfigVersion,this,&SEGSAdminTool::check_config_version);
     connect(this,&SEGSAdminTool::recreateConfig,m_generate_config_dialog,&GenerateConfigFileDialog::on_generate_config_file);
-    connect(ui->actionAbout,&QAction::triggered,m_about_dialog,&AboutDialog::show_ui);
+    connect(ui->action_about,&QAction::triggered,m_about_dialog,&AboutDialog::show_ui);
     connect(ui->update_detail,&QPushButton::clicked,m_update_dialog,&UpdateDetailDialog::show_update);
     connect(ui->createUser,&QPushButton::clicked,m_add_user_dialog,&AddNewUserDialog::on_add_user);
     connect(ui->runDBTool,&QPushButton::clicked,this,&SEGSAdminTool::check_db_exist);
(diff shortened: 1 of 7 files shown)


miles200:
@nemerle  @broxen  this can be reviewed and merged, give me a shout if you want me to just merge

broxen (approved):
This seems good for me. I think it's good to merge. I've started digging into dbtool to assist with the db checks.

Utilities/SEGSAdmin/SettingsDialog.cpp
```diff
@@ -113,20 +116,32 @@ void SettingsDialog::read_config_file(QString filePath)
     config_file.beginGroup("AdminServer");
     config_file.beginGroup("AccountDatabase");
     QString acc_db_driver = config_file.value("db_driver","").toString();
+    QString acc_db_name = config_file.value("db_name","").toString();
+    if (acc_db_driver == "QSQLITE")
+    {
+        acc_db_name.chop(3);
```
broxen, line 122: We might consider handling this a different way. Users may name sqlite databases with any extension (or with none at all). Common extensions are `.sql` `.db` `.sqlite` and this code would fail in those cases. 

It also trims the true name of a database without any extension. For example, in testing dbtool I will switch back and forth from sqlite, mysql, to pgsql and back. It's easiest for me to have no extension at all.

Perhaps a better approach is to simply search for illegal characters if database is MYSQL or PGSQL and throwing an error or filtering out?

https://dev.mysql.com/doc/refman/8.0/en/identifiers.html#:~:text=Internally,%20identifiers

Utilities/SEGSAdmin/SettingsDialog.cpp
```diff
@@ -325,15 +340,32 @@ void SettingsDialog::generate_default_config_file(QString ip)
...
 {
     QSettings config_file_write(Settings::getSettingsPath(), QSettings::IniFormat);
+    QString acc_db_driver = ui->acc_dbdriver->currentText();
+    QString char_db_driver = ui->char_dbdriver->currentText();
+    QString acc_db_name = ui->acc_dbname->text();
+    QString char_db_name = ui->char_dbname->text();
+
+    if (acc_db_driver == "QSQLITE")
+    {
+        acc_db_name.append(".db");
```
broxen, line 350: Same as before, I wouldn't assume the extension. Maybe we simply permit extensions if dbase driver is SQLITE, instead of forcing a specific one on the user.

Utilities/SEGSAdmin/SEGSAdminTool.cpp
```diff
@@ -222,31 +223,41 @@ void SEGSAdminTool::check_db_exist(bool on_startup)
...
     QSettings config(Settings::getSettingsPath(), QSettings::IniFormat, nullptr);
 
+    QString acc_db_driver = config.value("AdminServer/AccountDatabase/db_driver").toString();
+    QString char_db_driver = config.value("AdminServer/CharacterDatabase/db_driver").toString();
     QFileInfo file1(config.value(QStringLiteral("AdminServer/AccountDatabase/db_name"), "segs.db").toString());
     QFileInfo file2(config.value(QStringLiteral("AdminServer/CharacterDatabase/db_name"), "segs_game.db").toString());
     if(on_startup) // Runs this check on startup or for checking creation in other methods
     {
-        if(file1.exists() && file2.exists())
+        if (acc_db_driver == "QSQLITE" && char_db_driver == "QSQLITE")
```
broxen, line 232: In a (weird) scenario where account db uses SQLITE but char db uses another, this will fail to execute.

broxen:
Took another look from my desktop and saw a couple minor edge cases worth accounting for.

Utilities/SEGSAdmin/SettingsDialog.cpp
broxen: Along these lines a common regex filter for dbase name in php would be `[0-9a-zA-Z$_]+` and we could simply add a period to that for SQLITE

broxen: Or this, which also checks for length, from the interwebs: `^[^\\/?%*:|\"<>.]{1,64}$`

mobbyg:
This good to merge?

